PR #2, opened 2024-11-06
[DPE-5762] Add CI workflows

reneradoi:
# Summary
In order to build, test and publish the etcd operator automatically, we will run Github action workflows when creating or pushing to PR's or merging PR's to the default branch. These workflows will be based on `data-platform-workflows`.

# Note
We are currently still using:
- Ubuntu 22.04 base (due to https://github.com/canonical/data-platform-workflows/issues/246, `data-platform-workflows` don't support 24.04 yet)
- Juju 3.5.4 as default for integration testing (3.6 has not yet been released to stable, but is already added to scheduled test runs)

Furthermore the release job for publishing to Charmhub is not active yet because credentials still need to be added there.


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,87 @@
...
+on:
+  workflow_call:
+  pull_request:
+  schedule:
+    - cron: '53 0 * * *' # Daily at 00:53 UTC
+
+jobs:
+  lint:
+    name: Lint
+    uses: canonical/data-platform-workflows/.github/workflows/lint.yaml@v22.0.0
```
skourta, line 19: I think we should update to the latest data platform workflows

metadata.yaml
```diff
@@ -0,0 +1,10 @@
+# Copyright 2024 Canonical Ltd.
```
skourta, line 1: the `metadata.yaml` is no longer needed since charmcraft 2.5 I believe the only file needed now is `charmcraft.yaml`

skourta (approved):
I left minor comments

.github/workflows/ci.yaml
reneradoi: Agree.

metadata.yaml
reneradoi: Unfortunately we have to keep it due to https://github.com/canonical/data-platform-workflows/issues/169. I've started without, but CI job fail then.

PR #3, opened 2024-11-11
[DPE-5764] Add basic functionality

reneradoi:
This PR adds basic functionality to the etcd operator charm. It installs the snap, creates the configuration for the initial cluster consisting of the deployed units and starts the cluster.

Additions:
- basic source code structure in accordance with data-platform best practice
- basic config manager
- basic implementation of peer-relation app/unit data handling
- basic state implementation for etcd-unit/server and etcd-cluster
- basic event handling for charm- and peer-relation-events such as `on_install`, `on_start`, `on_peer_cluster_relation_created`
- basic unit test coverage

When deploying an etcd-cluster of n units, the charm sets up these units and their config, creates the initial etcd-cluster and brings the units into `active` status:
```
$ juju deploy ./charmed-etcd-operator_ubuntu-22.04-amd64.charm etcd -n 3
$ juju status
Model  Controller      Cloud/Region         Version  SLA          Timestamp
etcd   dev-controller  localhost/localhost  3.5.4    unsupported  15:24:28Z

App   Version  Status  Scale  Charm                  Channel  Rev  Exposed  Message
etcd           active      3  charmed-etcd-operator            20  no       

Unit      Workload  Agent  Machine  Public address  Ports  Message
etcd/27   active    idle   33       10.54.237.57           
etcd/28   active    idle   34       10.54.237.109          
etcd/29*  active    idle   35       10.54.237.30           

$ juju ssh etcd/27
# etcd.etcdctl endpoint health --endpoints=10.54.237.57:2379
10.54.237.57:2379 is healthy: successfully committed proposal: took = 2.017893ms

# etcd.etcdctl member list --endpoints=10.54.237.57:2379
7656a31b37f8b407, started, etcd29, http://10.54.237.30:2380, http://10.54.237.30:2379, false
b8a5a96e3126f130, started, etcd28, http://10.54.237.109:2380, http://10.54.237.109:2379, false
f7c51b64202cc9ca, started, etcd27, http://10.54.237.57:2380, http://10.54.237.57:2379, false
```

**Note:** The charm is currently still deploying the legacy-snap `etcd`.


src/core/cluster.py
```diff
@@ -0,0 +1,93 @@
...
+    def servers(self) -> Set[EtcdServer]:
+        """Get all servers/units in the current peer relation, including this unit itself.
+
+        Returns:
+            Set of EtcdServers with their unit data.
+        """
+        if not self.peer_relation:
+            return set()
+
+        servers = set()
```
skourta, line 81: Curiosity: Why is this a set?

src/managers/config.py
```diff
@@ -0,0 +1,116 @@
...
+
+from ops.model import ConfigData
+
+from core.cluster import ClusterState
+from core.workload import WorkloadBase
+from literals import CONFIG_PATH
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_PROPERTIES = """
```
skourta, line 17: I think we can make use of string templates with placeholders here. It would be more readable IMO. It would also avoid the string concatenation from lists.

src/workload.py
```diff
@@ -0,0 +1,58 @@
...
+    @override
+    def alive(self) -> bool:
+        try:
+            return bool(self.etcd.services[SNAP_SERVICE]["active"])
+        except KeyError:
+            return False
+
+    @override
+    def write(self, content: str, path: str) -> None:
+        os.makedirs(os.path.dirname(path), exist_ok=True)
```
skourta, line 55: nit: we can make use of the new Pathlib syntax for creating and managing paths, dirs, and files.

skourta (approved):
Great work @reneradoi. Simple and elegant. I left some minor comments and suggestions.

src/core/cluster.py
reneradoi: Three reasons:
- to comply with other charmed operators
- no duplicate elements possible
- very easy to check membership

src/core/models.py
```diff
@@ -0,0 +1,121 @@
...
+    def update(self, items: dict[str, str]) -> None:
+        """Write to relation data."""
+        if not self.relation:
+            logger.warning(
+                f"Fields {list(items.keys())} were attempted to be written on the relation before it exists."
+            )
+            return
+
+        delete_fields = [key for key in items if not items[key]]
+        update_content = {k: items[k] for k in items if k not in delete_fields}
```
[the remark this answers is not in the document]
reneradoi: But we only want to pass updates to the lib, not deletes.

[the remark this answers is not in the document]
skourta: Yes but in this case you will pass the same values for the deleted ones so no update will happen. Then you delete afterwards.

src/workload.py
reneradoi: I've replaced it.

src/workload.py
```diff
@@ -51,8 +51,9 @@ def alive(self) -> bool:
...
 
     @override
-    def write(self, content: str, path: str) -> None:
-        os.makedirs(os.path.dirname(path), exist_ok=True)
+    def write_file(self, content: str, file: str) -> None:
+        path = "/".join(file.split("/")[:-1])
+        pathlib.Path(path).mkdir(parents=True, exist_ok=True)
 
-        with open(path, "w") as f:
+        with open(file, "w") as f:
```
skourta, line 58: see [Path.write_text](https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_text)

src/workload.py
```diff
@@ -51,8 +51,9 @@ def alive(self) -> bool:
             return False
 
     @override
-    def write(self, content: str, path: str) -> None:
-        os.makedirs(os.path.dirname(path), exist_ok=True)
+    def write_file(self, content: str, file: str) -> None:
+        path = "/".join(file.split("/")[:-1])
```
skourta, line 55: It would be better if you create the Path object first thing then manage it though the pathlib interface.
See [Path.parents](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parents)

skourta (commented):
I left some comments about using Pathlib to create both the path and writing to the file. It should be cleaner if we use its interface.

src/workload.py
reneradoi: Done

src/workload.py
reneradoi: Thanks for the hint! I like that, it's even cleaner.

skourta (dismissed):

src/managers/config.py
```diff
@@ -0,0 +1,116 @@
...
+proxy-write-timeout: 5000
+proxy-read-timeout: 0
+force-new-cluster: false
+auto-compaction-mode: periodic
+auto-compaction-retention: "1"
+"""
+
+# these config properties are not used at the moment
+# they are only listed here for completeness
+TLS_PROPERTIES = """
```
Mehdi-Bendriss, line 40: Similar comment, regarding yaml parsing - we can have this defined as a python dictionary and dump it into the config file using the yaml lib

src/managers/config.py
```diff
@@ -0,0 +1,116 @@
...
+        """Assemble the config properties.
+
+        Returns:
+            List of properties to be written to the config file.
+        """
+        properties = [
+            f"name: {self.state.unit_server.member_name}",
+            f"initial-advertise-peer-urls: {self.state.unit_server.peer_url}",
+            f"initial-cluster-state: {self.state.cluster.initial_cluster_state}",
+            f"listen-peer-urls: {self.state.unit_server.peer_url}",
```
Mehdi-Bendriss, line 90: should we not have a list of peers here?

src/workload.py
```diff
@@ -0,0 +1,57 @@
...
+from literals import SNAP_NAME, SNAP_REVISION, SNAP_SERVICE
+
+logger = logging.getLogger(__name__)
+
+
+class EtcdWorkload(WorkloadBase):
+    """Implementation of WorkloadBase for running on VMs."""
+
+    def __init__(self):
+        self.etcd = snap.SnapCache()[SNAP_NAME]
```
Mehdi-Bendriss, line 23: we should have a retry mechanism here, this is often prone to failing the first time - similar to [here](https://github.com/canonical/opensearch-operator/blob/2/edge/src/opensearch.py#L52)

src/workload.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @override
+    def start(self) -> None:
+        try:
+            self.etcd.start(services=[SNAP_SERVICE])
+        except snap.SnapError as e:
+            logger.exception(str(e))
+
+    def install(self) -> bool:
```
Mehdi-Bendriss, line 32: Similarly here, we can have a retry mechanism with Tenacity

src/workload.py
```diff
@@ -0,0 +1,57 @@
...
+            self.etcd.hold()
+            return True
+        except snap.SnapError as e:
+            logger.error(str(e))
+            return False
+
+    @override
+    def alive(self) -> bool:
+        try:
+            return bool(self.etcd.services[SNAP_SERVICE]["active"])
```
Mehdi-Bendriss, line 49: what does `self.etcd.services[SNAP_SERVICE]["active"]` return? I'm worried it returns a non-empty string where `bool("my_str")` always evaluates to True

Mehdi-Bendriss (commented):
Thanks René! Great work, very clean! I left some comments and questions to clarify.

src/core/cluster.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class ClusterState(Object):
+    """Global state object for the etcd cluster."""
+
+    def __init__(self, charm: "EtcdOperatorCharm", substrate: SUBSTRATES):
+        super().__init__(parent=charm, key="charm_state")
+        self.substrate: SUBSTRATES = substrate
+        self.peer_app_interface = DataPeerData(self.model, relation_name=PEER_RELATION)
+        self.peer_unit_interface = DataPeerUnitData(self.model, relation_name=PEER_RELATION)
```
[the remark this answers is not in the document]
reneradoi: Just to keep the naming consistent with others:
- Zookeeper: https://github.com/canonical/zookeeper-operator/blob/main/src/core/cluster.py#L38-L39
- Kafka: https://github.com/canonical/kafka-operator/blob/main/src/core/cluster.py#L103-L104

src/managers/cluster.py
```diff
@@ -0,0 +1,33 @@
...
+import socket
+
+logger = logging.getLogger(__name__)
+
+
+class ClusterManager:
+    """Manage cluster members, quorum and authorization."""
+
+    def __init__(self):
+        pass
```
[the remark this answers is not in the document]
reneradoi: Ignored for now, will be extended in the next PR anyway.

src/workload.py
reneradoi: It returns a bool: https://charmhub.io/operator-libs-linux/libraries/snap#snapservice

metadata.yaml
```diff
@@ -4,11 +4,14 @@
...
 title: Charmed etcd VM operator
 summary: Operator for etcd databases in VM environments.
 description: |
   Etcd is a distributed, reliable key-value store for the most critical data of
   distributed systems, such as Kubernetes. This charmed operator deploys and operates
   etcd on virtual machines.
+
+peers:
+  etcd-cluster:
+    interface: etcd_cluster
```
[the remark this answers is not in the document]
reneradoi: Done.

src/core/models.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @property
+    def unit_name(self) -> str:
+        """The id of the unit from the unit name."""
+        return self.unit.name
+
+    @property
+    def member_name(self) -> str:
+        """The Human-readable name for this etcd cluster member."""
+        return f"etcd{self.unit_id}"
```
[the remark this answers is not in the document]
reneradoi: Agree, updated this to the application name.

src/core/models.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @property
+    def hostname(self) -> str:
+        """The hostname for the unit."""
+        return self.relation_data.get("hostname", "")
```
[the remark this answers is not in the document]
reneradoi: Agree, this is just a dummy for now and will be subject of change in the next PR.

src/core/models.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @property
+    def ip(self) -> str:
+        """The IP address for the unit."""
+        return self.relation_data.get("ip", "")
```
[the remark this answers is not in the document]
reneradoi: Agree, this is just a dummy for now and will be subject of change in the next PR.

src/events/etcd.py
```diff
@@ -0,0 +1,122 @@
...
+        self.framework.observe(
+            self.charm.on[PEER_RELATION].relation_departed, self._on_cluster_relation_departed
+        )
+        self.framework.observe(self.charm.on.leader_elected, self._on_leader_elected)
+        self.framework.observe(self.charm.on.update_status, self._on_update_status)
+
+    def _on_install(self, event: ops.InstallEvent) -> None:
+        """Handle install event."""
+        install = self.charm.workload.install()
+        if not install:
```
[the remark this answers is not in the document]
reneradoi: Done.

src/events/etcd.py
```diff
@@ -0,0 +1,122 @@
...
+            self.charm.set_status(Status.SERVICE_NOT_INSTALLED)
+            event.defer()
+            return
+
+    def _on_start(self, event: ops.StartEvent) -> None:
+        """Handle start event."""
+        # Make sure all planned units have joined the peer relation before starting the cluster
```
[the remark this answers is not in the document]
reneradoi: Required to be able to assemble all the ip's for the cluster-endpoints in the `initial-cluster` config. If those can't be set, the unit can't join the cluster and could only be started as individual instance.

At a later point, this may be subject of change, as the cluster initialization needs to be robust for HA and failover. For now, this is just to be able to form an etcd-cluster on startup.

src/managers/cluster.py
```diff
@@ -0,0 +1,33 @@
...
+        """Collect hostname mapping for current unit.
+
+        Returns:
+            Dict of string keys 'hostname', 'ip' and their values
+        """
+        hostname = socket.gethostname()
+
+        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        s.settimeout(0)
+        s.connect(("10.10.10.10", 1))
```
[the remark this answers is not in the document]
reneradoi: Has been replaced by `socket.gethostbyname()`.

src/managers/config.py
```diff
@@ -0,0 +1,116 @@
...
+
+from ops.model import ConfigData
+
+from core.cluster import ClusterState
+from core.workload import WorkloadBase
+from literals import CONFIG_FILE
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_PROPERTIES = """
```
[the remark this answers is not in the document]
reneradoi: I would like to avoid loading the config-file to avoid anyone overwriting config (as explained in the config "vision"/discussion some days ago). We could argue about moving this to a different place in the code or a different format (currently it's in line with other charms doing it the same way), but I think the charm should be able to fully control the config and not rely on anything written on disk.

src/managers/config.py
reneradoi: See above.

src/managers/config.py
reneradoi: No not in this case. The list of peers is in `initial-cluster`. The configs `initial-advertise-peer-urls`, `listen-peer-urls`, `listen-client-urls` and `advertise-client-urls` are only for configuring the local host on which addresses to listen. Multiple urls here could be relevant for e.g. listening on traffic from localhost or listening on multiple ports. Details see [clustering guide](https://etcd.io/docs/v3.5/op-guide/clustering/#static)(scroll down to the three commands starting the three machines).

src/workload.py
reneradoi: Done.

src/workload.py
reneradoi: Done.

src/events/etcd.py
```diff
@@ -0,0 +1,122 @@
...
+        )
+        self.framework.observe(self.charm.on.leader_elected, self._on_leader_elected)
+        self.framework.observe(self.charm.on.update_status, self._on_update_status)
+
+    def _on_install(self, event: ops.InstallEvent) -> None:
+        """Handle install event."""
+        install = self.charm.workload.install()
+        if not install:
+            self.charm.set_status(Status.SERVICE_NOT_INSTALLED)
+            event.defer()
```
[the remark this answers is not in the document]
reneradoi: Done.

src/workload.py
Mehdi-Bendriss: we shouldn't need the `bool()` casting then, right?

src/events/etcd.py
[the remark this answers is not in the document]
Mehdi-Bendriss: This can be the case for the time being - while we haven't tackled scale-up, however we will need to abandon it and move to a rolling-(re)start approach when we start HA, where we should not be constrained with all planned units joining to start a cluster or** to add new units (think of a hardware failure in a new unit for ex). 

In which case this condition will no longer be relevant, as we should be able to bootstrap a cluster at first with only the first node and have additional nodes join the existing cluster using `--initial-cluster-state=existing` (as opposed to `=new`).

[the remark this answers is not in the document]
reneradoi: Yes, that's exactly the plan.

src/workload.py
reneradoi: Yes, it's only there for better readability and type hinting.

src/managers/config.py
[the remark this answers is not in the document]
reneradoi: Thanks for the advice! I'd like to have it stored as part of the code, meaning I'll move the plain text to a yaml config file and load it from there. I will make this a separate PR if that's fine?

Mehdi-Bendriss (approved):

pyproject.toml
```diff
@@ -7,6 +7,7 @@ package-mode = false
 [tool.poetry.dependencies]
 python = "^3.10"
 ops = "^2.17.0"
+tenacity = "*"
```
skourta, line 10: Why is tenacity not pinned at least to a major version?

skourta (commented):
Just one comment about tenacity version.

skourta (approved):

pyproject.toml
reneradoi: I didn't pin it because I didn't consider it important enough. If there is a known breaking change (like, for example with `websockets`) or something is very important and it's better to be cautious (like, e.g. `ops`), it makes sense to pin a version. If not, I prefer to fail fast :)

PR #5, opened 2024-11-25
[DPE-5946] Add cluster manager

reneradoi:
This PR adds two integral components to the etcd operator:
- a ClusterManager class that can run tasks against the Raft leader of the etcd cluster
- an EtcdClient class that can run `etcdctl` commands against the etcd daemon

With these, it is possible to implement further functionality such as adding and promoting members or any other quorum-related task.

The PR also adds test coverage for both newly added components by asserting the `get_leader` method was successful. This method is executed test-wise in the `on_peer_relation_joined` event.


src/managers/cluster.py
```diff
@@ -26,3 +37,93 @@ def get_host_mapping(self) -> dict[str, str]:
...
+                    break
+            except KeyError:
+                # for now, we don't raise an error if there is no leader
+                # this may change when we have actual relevant tasks performed against the leader
+                logger.warning("No raft leader found in cluster.")
+
+        return leader
+
+
+class EtcdClient:
```
skourta, line 64: Is there a best practise on where we can store these `helper` classes. I would prefer to have them in independent files if possible unless it's a DP best standard.

src/managers/cluster.py
```diff
@@ -26,3 +37,93 @@ def get_host_mapping(self) -> dict[str, str]:
...
+        for endpoint in self.cluster_endpoints:
+            client = EtcdClient(client_url=endpoint)
+            try:
+                endpoint_status = client.get_endpoint_status()
+                member_id = endpoint_status["Status"]["header"]["member_id"]
+                leader_id = endpoint_status["Status"]["leader"]
+                if member_id == leader_id:
+                    leader = endpoint
+                    break
+            except KeyError:
```
skourta, line 56: Is there a case where no leader is found?

skourta (commented):
Thank you, René. I left some comments.

src/managers/cluster.py
skourta: For example I will use the class for TLS as well and I think it will be used in multiple managers.

src/managers/cluster.py
```diff
@@ -26,3 +37,93 @@ def get_host_mapping(self) -> dict[str, str]:
...
+            return an empty string. It will not raise an error in order to leave error handling
+            up to the caller. Depending on what command is executed, the ways of handling errors
+            might differ.
+        """
+        try:
+            result = subprocess.run(
+                args=[
+                    "etcdctl",
+                    command,
+                    subcommand,
```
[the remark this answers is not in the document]
skourta: There is also the [`check_output`](https://docs.python.org/3/library/subprocess.html#subprocess.check_output) which basically does this for you.

src/managers/cluster.py
```diff
@@ -26,3 +37,93 @@ def get_host_mapping(self) -> dict[str, str]:
...
+                    f"--endpoints={endpoints}",
+                    f"-w={output_format}",
```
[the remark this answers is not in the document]
reneradoi: Thank you, I missed that one. Added a default now.

src/managers/cluster.py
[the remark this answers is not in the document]
reneradoi: I've started the implementation with `check_output`, but as the recommended approach to invoking subprocesses is to use the `run()` function, I switched to that.

I also tried to avoid "passing a list of arguments". I know this is possible and very convenient, but I've also seen a lot of issues with this. A recent one was masking credentials from being written to logfiles, which is much harder if you have just a list instead of dedicated parameters for e.g. password. Therefore I opted for a "controlled" approach of explicitly creating and naming those parameters that we need. This list will have to be extended eventually, that's true. Let's try to follow the approach of adding complexity only then when we need it. That means adding e.g. the arguments for the subcommands when we encounter the first need for it.

src/managers/cluster.py
```diff
@@ -26,3 +37,93 @@ def get_host_mapping(self) -> dict[str, str]:
         ip = socket.gethostbyname(hostname)
 
         return {"hostname": hostname, "ip": ip}
+
+    def get_leader(self) -> str:
+        """Query the etcd cluster for the raft leader and return the client_url as string."""
+        leader = ""
```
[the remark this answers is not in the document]
reneradoi: Agree.

src/managers/cluster.py
```diff
@@ -26,3 +37,93 @@ def get_host_mapping(self) -> dict[str, str]:
...
+        # if they match, return this host's endpoint
+        for endpoint in self.cluster_endpoints:
+            client = EtcdClient(client_url=endpoint)
+            try:
+                endpoint_status = client.get_endpoint_status()
+                member_id = endpoint_status["Status"]["header"]["member_id"]
+                leader_id = endpoint_status["Status"]["leader"]
+                if member_id == leader_id:
+                    leader = endpoint
+                    break
```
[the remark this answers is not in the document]
reneradoi: Will do.

src/managers/cluster.py
reneradoi: Yes, this could happen during leader election or in error cases.

src/managers/cluster.py
```diff
@@ -4,17 +4,28 @@
...
+
+
 class ClusterManager:
     """Manage cluster members, quorum and authorization."""
 
-    def __init__(self):
-        pass
+    def __init__(self, state: ClusterState):
+        self.state = state
+        self.cluster_endpoints = [server.client_url for server in self.state.servers]
```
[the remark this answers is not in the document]
reneradoi: I've been thinking about it and went for the simplicity of an attribute. What is your reasoning or the benefit of having this as a property?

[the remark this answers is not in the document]
skourta: Resolved in a meeting.

src/managers/cluster.py
reneradoi: I've now moved `EtcdClient` to `src/common/client.py`. This way we can easily use it in multiple managers and don't have to handle it as a lib. The path `src/common/` follows the example from several Big Data repos (e.g. Spark, Kyuubi).

skourta (approved):
Thank you for making the changes René.

src/common/client.py
```diff
@@ -0,0 +1,79 @@
...
+            endpoint_status = json.loads(result)[0]
+
+        return endpoint_status
+
+    def _run_etcdctl(
+        self,
+        command: str,
+        subcommand: str | None,
+        endpoints: str,
+        output_format: str | None = "simple",
```
Mehdi-Bendriss, line 41: 1. what does `output_format = None` represent? 
2. should the default not be always `json` to simplify the usage of output, so that `json.loads` happens only once in this method?

src/common/client.py
```diff
@@ -0,0 +1,79 @@
...
+                    "etcdctl",
+                    command,
+                    subcommand,
+                    f"--endpoints={endpoints}",
+                    f"-w={output_format}",
+                ],
+                check=True,
+                capture_output=True,
+                text=True,
+            ).stdout
```
Mehdi-Bendriss, line 74:
```suggestion
            ).stdout.strip()
```

src/common/client.py
```diff
@@ -0,0 +1,79 @@
...
+        except subprocess.CalledProcessError as e:
+            logger.warning(e)
+            return None
```
Mehdi-Bendriss, line 77: We should perhaps propagate the error here instead of returning `None`

Mehdi-Bendriss (commented):
Thanks René! I left a couple of comments

src/common/client.py
reneradoi: This is not required here, as we use `.stdout`. It would cause an error instead:

```
subprocess.run(["charmed-etcd.etcdctl", "member", "list", "--endpoints=10.54.237.224:2379", "-w=simple"]).stdout.strip()
3c64c92fd7480501, started, etcd50, http://10.54.237.252:2380, http://10.54.237.252:2379, false
75a3fc47cc32bceb, started, etcd48, http://10.54.237.224:2380, http://10.54.237.224:2379, false
75b9c521876ed7e5, started, etcd49, http://10.54.237.117:2380, , false
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'strip'
```

src/common/client.py
reneradoi: 1. This means you don't have to set `output_format` when calling `_run_etcdctl()` if you only need `simple` output. Only if you need `json`, you need to set it.
2. I assume the majority of our calls will only require `simple` output format, especially for the imperative commands like `member add`, `user add`, `role add` and so on. Only the `list` or `status` commands will require json-output. That's why I opted for the default to be `simple`.

src/common/client.py
reneradoi: I'm open for adjusting this, just want to give my reasoning here and would like to get your feedback on it: 

This function `_run_etcdctl` is very generic, as it will execute any command passed to it. The only error we will ever get back is the `CalledProcessError`, which would then have to be raised to the caller. We would then have to handle the same `CalledProcessError` in all places where we call `_run_etcdctl`. If we would only have to check the return, this would be much simpler: `if result := self._run_etcdctl(...).

Please let me know if you agree or not.

src/common/client.py
Mehdi-Bendriss: @reneradoi the above command fails as you're not capturing the output of the command - it is simply printing it.
However for the setup you have with `capture_output=True, text=True` will succeed. 
```
>>> subprocess.run(["charmed-etcd.etcdctl", "member", "list", "--endpoints=localhost:2379", "-w=simple"], check=True, capture_output=True, text=True).stdout
'8e9e05c52164694d, started, default, http://localhost:2380, http://localhost:2379, false\n'
>>> subprocess.run(["charmed-etcd.etcdctl", "member", "list", "--endpoints=localhost:2379", "-w=simple"], check=True, capture_output=True, text=True).stdout.strip()
'8e9e05c52164694d, started, default, http://localhost:2380, http://localhost:2379, false'
```

src/common/client.py
Mehdi-Bendriss: It's not an issue - however note that you will have to handle the result of your commands a bit more gracefully in the case of `result == ""` vs `result == None`

Meaning that you will not be able to perform checks like:
```
if not self._run_etcdctl(..):
   # this can mean Exception or success with no text result 
```

src/common/client.py
reneradoi: Sorry, missed that. Thanks for the hint!

src/common/client.py
reneradoi: Yes I'm aware of this. That's why we have `capture_output=True` in the `subprocess` command, to always get something back.

Mehdi-Bendriss (approved):

skourta (approved):

skourta (approved):

PR #6, opened 2024-11-26
Add basic integration testing

reneradoi:
This PR adds a basic integration test to the etcd operator which:
- deploys a cluster of three etcd nodes
- checks the cluster's availability
- checks if all deployed nodes where added as cluster members
- writes test data to etcd
- reads the test data

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python3
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import json
+import subprocess
+from pathlib import Path
+
+import yaml
+from pytest_operator.plugin import OpsTest
+
+from literals import CLIENT_PORT
+
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+APP_NAME = METADATA["name"]
+
+
+def put_key(endpoints: str, key: str, value: str) -> str:
+    """Write data to etcd using `etcdctl`."""
+    return subprocess.run(
+        args=[
+            "etcdctl",
+            "put",
+            key,
+            value,
+            f"--endpoints={endpoints}",
+        ],
+        check=True,
+        capture_output=True,
+        text=True,
+    ).stdout.split("\n")[0]
+
+
+def get_key(endpoints: str, key: str) -> str:
+    """Read data from etcd using `etcdctl`."""
+    return subprocess.run(
+        args=[
+            "etcdctl",
+            "get",
+            key,
+            f"--endpoints={endpoints}",
+        ],
+        check=True,
+        capture_output=True,
+        text=True,
+    ).stdout.split("\n")[1]
+
+
+def get_cluster_members(endpoints: str) -> list[dict]:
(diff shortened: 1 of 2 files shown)


tests/integration/helpers.py
```diff
@@ -0,0 +1,61 @@
...
+    """Resolve the etcd endpoints for a given juju application."""
+    return ",".join(
+        [
+            f"http://{unit.public_address}:{CLIENT_PORT}"
+            for unit in ops_test.model.applications[app_name].units
+        ]
+    )
+
+
+async def get_juju_leader_unit_name(ops_test: OpsTest, app_name: str = APP_NAME) -> str:
```
skourta, line 57: The return should be `str | None`. I know there is always a leader but str is not returned in all code paths.
You can also add a raise an error after the for loop.

tests/integration/helpers.py
```diff
@@ -0,0 +1,61 @@
...
+
+    return json.loads(result)["members"]
+
+
+def get_cluster_endpoints(ops_test: OpsTest, app_name: str = APP_NAME) -> str:
+    """Resolve the etcd endpoints for a given juju application."""
+    return ",".join(
+        [
+            f"http://{unit.public_address}:{CLIENT_PORT}"
+            for unit in ops_test.model.applications[app_name].units
```
skourta, line 52: I would add some asserts or tests on the model and the app to ensure they exist.

tests/integration/test_charm.py
```diff
@@ -3,29 +3,49 @@
...
+    leader_unit = await get_juju_leader_unit_name(ops_test, APP_NAME)
+
+    cluster_members = get_cluster_members(model, leader_unit, endpoints)
+    assert len(cluster_members) == NUM_UNITS
+
+    # make sure data can be written to the cluster
+    test_key = "test_key"
+    test_value = "42"
+    assert put_key(model, leader_unit, endpoints, key=test_key, value=test_value) == "OK"
+    assert get_key(model, leader_unit, endpoints, key=test_key) == test_value
```
skourta, line 51: I would check all 3 units separately to ensure it was replicated.

skourta (approved):
Thank you, René. This should get us started with the integration tests. I left enhancement comments but it already LGTM.

tests/integration/test_charm.py
Mehdi-Bendriss: May be premature, and will be addressed more in-depth during HA - but that's a valid point. 
For HA, we also need to have continuous writes routines implemented as part of the tests. Which should have replication checks happen as part of the continuous writes interface.

Mehdi-Bendriss (approved):

tests/integration/test_charm.py
reneradoi: Agree and will be done when working on HA.

reneradoi:
Thank you for your feedback. We can address the comments when we further enhance the tests. Of course we need to add more tests and more functionality to the testing suite. This PR here is just considered a basic coverage to get fast feedback when making changes to the code, and to give a starting point on how to create the tests.

PR #9, opened 2024-12-02
[DPE-5976] Add admin user

reneradoi:
This PR adds the following functionality to the etcd-operator:
- automatically create an internal admin user (`root` as per [etcd standard](https://etcd.io/docs/v3.5/op-guide/authentication/rbac/#special-users-and-roles))
- always enable authentication in the etcd cluster on startup
- methods `add_user()`, `update_password()` and `enable_auth()` in the `EtcdClient` class
- a configuration option `system-users` with which users can configure a secret_id. The corresponding secret has to be granted to the etcd application and has to include the password as key `root=...`.
- unit and integration test coverage for the above mentioned


tests/unit/test_actions.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # make sure admin credentials are created initially
+    state_in = testing.State(relations={relation}, leader=True)
+    state_out = ctx.run(ctx.on.leader_elected(), state_in)
+
+    ctx.run(ctx.on.action("get-password", params={"username": f"{INTERNAL_USER}"}), state_out)
+    assert ctx.action_results.get("username") == INTERNAL_USER
+    assert ctx.action_results.get("password")
+    assert ctx.action_results.get("ca-chain")
+
```
skourta, line 33: Maybe add a test case with a user different from root?

tests/unit/test_actions.py
```diff
@@ -0,0 +1,71 @@
...
+def test_get_password():
+    ctx = testing.Context(EtcdOperatorCharm)
+    relation = testing.PeerRelation(id=1, endpoint=PEER_RELATION)
+
+    # make sure admin credentials are created initially
+    state_in = testing.State(relations={relation}, leader=True)
+    state_out = ctx.run(ctx.on.leader_elected(), state_in)
+
+    ctx.run(ctx.on.action("get-password", params={"username": f"{INTERNAL_USER}"}), state_out)
+    assert ctx.action_results.get("username") == INTERNAL_USER
```
skourta, line 30: nit: `assert ctx.action_results`

tests/unit/test_actions.py
```diff
@@ -0,0 +1,71 @@
...
+    with patch(
+        "subprocess.run", return_value=CompletedProcess(returncode=0, args=[], stdout="OK")
+    ):
+        ctx.run(
+            ctx.on.action(
+                "set-password", params={"username": INTERNAL_USER, "password": password}
+            ),
+            state_in,
+        )
+        assert ctx.action_results.get(f"{INTERNAL_USER}-password") == password
```
skourta, line 71: nit: `assert ctx.action_results`

tests/unit/test_charm.py
```diff
@@ -20,6 +26,16 @@ def test_install_failure_blocked_status():
...
 
 
+def test_internal_user_creation():
+    ctx = testing.Context(EtcdOperatorCharm)
+    relation = testing.PeerRelation(id=1, endpoint=PEER_RELATION)
+
+    state_in = testing.State(relations={relation}, leader=True)
+    state_out = ctx.run(ctx.on.leader_elected(), state_in)
+    secret_out = state_out.get_secret(label=f"{PEER_RELATION}.{APP_NAME}.app")
+    assert secret_out.latest_content.get(f"{INTERNAL_USER}-password")
```
skourta, line 36: nit: `assert secret_out.latest_content`

tests/integration/test_charm.py
```diff
@@ -45,7 +48,62 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
     assert len(cluster_members) == NUM_UNITS
```
skourta, line 48: nit: 
```    
assert ops_test.model
assert model
```

Before the deploy statement

tests/integration/test_charm.py
```diff
@@ -45,7 +48,62 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
...
+    leader_unit = await get_juju_leader_unit_name(ops_test, APP_NAME)
+    test_key = "test_key"
+    test_value = "42"
+    new_password = "my_new_pwd"
+
+    # check that reading/writing data without credentials fails
+    assert get_key(model, leader_unit, endpoints, key=test_key) != test_value
+    assert put_key(model, leader_unit, endpoints, key=test_key, value=test_value) != "OK"
+
+    # run set-password action
```
skourta, line 92: We should add a test case to get data with the auto generated password before we change it.

skourta (commented):
Great work @reneradoi. I left some comments

src/core/models.py
```diff
@@ -119,3 +119,16 @@ def __init__(
...
+        """Retrieve the credentials for the internal admin user."""
+        if password := self.relation_data.get(f"{INTERNAL_USER}-password"):
+            return {INTERNAL_USER: password}
+
+        return {}
+
+    @property
+    def auth_enabled(self) -> bool:
+        """Flag to check if authentication is already enabled in the Cluster."""
+        return self.relation_data.get("authentication", "") == "enabled"
```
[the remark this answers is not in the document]
reneradoi: What do you mean with that, could you please elaborate?

src/events/actions.py
```diff
@@ -0,0 +1,74 @@
...
+
+        if not self.charm.state.cluster.internal_user_credentials:
+            event.fail("User credentials not created yet.")
+            return
+
+        event.set_results(
+            {
+                "username": username,
+                "password": self.charm.state.cluster.internal_user_credentials[INTERNAL_USER],
+                "ca-chain": "...",
```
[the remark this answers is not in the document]
reneradoi: Done

src/core/models.py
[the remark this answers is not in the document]
skourta: Sorry, I meant the `self.relation_data` -> `self.data` as in [here](https://github.com/canonical/charmed-etcd-operator/blob/4ea0531b174c446623a078b6a2d6baa7a2859f73/src/core/models.py#L35).

tests/unit/test_actions.py
reneradoi: This will not work as we get back the whole dict.

tests/unit/test_actions.py
reneradoi: This will not work as we get back the whole dict.

tests/unit/test_charm.py
reneradoi: This will not work as we get back the whole dict.

tests/integration/test_charm.py
reneradoi: What is the purpose of the assertions?

tests/integration/test_charm.py
reneradoi: Yes, there is, in the previous test function. We could move it here if you want to.

tests/unit/test_actions.py
skourta: This is nitpicking to silence pylance. I didn't suggest to make that the whole test just to add the test to make sure the 'action_results' is not None

tests/integration/test_charm.py
skourta: These are also mostly to assert the objects are not None and silence pylance.

src/common/client.py
```diff
@@ -18,9 +20,13 @@ class EtcdClient:
 
     def __init__(
         self,
+        username: str,
```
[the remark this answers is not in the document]
reneradoi: There are cases where we could omit it (e.g. `endpoints status`). Still I'd like to always instantiate the client the same way, just to be consistent. Otherwise we might have to adjust the instantiation of the client  alot in the future when we mix auth-requiring with non-auth commands in the same instance.

src/core/models.py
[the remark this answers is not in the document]
reneradoi: Done.

src/common/client.py
```diff
@@ -88,16 +87,16 @@ def _run_etcdctl(
         self,
         command: str,
         endpoints: str,
-        subcommand: Optional[str] = None,
+        subcommand: str = None,
```
skourta, line 90: The syntax should be:
```
subcommand: str | None = None,
```

reneradoi: Thanks for the hint, have updated it.

src/common/client.py
```diff
@@ -87,16 +87,16 @@ def _run_etcdctl(
...
-        auth_password: str = None,
-        user: str = None,
-        user_password: str = None,
-        output_format: str = "simple",
-        use_input: str = None,
+        auth_username: str | None = None,
+        auth_password: str | None = None,
+        user: str | None = None,
+        user_password: str | None = None,
+        output_format: str | None = "simple",
```
skourta, line 98: There's no need for `None` here

tests/integration/test_charm.py
skourta: True.

skourta (dismissed):
Thank you!

actions.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+set-password:
+  description: Change the admin user's password, which is used by charm. It is for internal charm users and SHOULD NOT be used by applications.
+  params:
+    username:
+      type: string
+      description: The username, the default value 'root'. Possible values - 'root'.
```
delgod, line 9: please excuse me, we should use user secrets for new charms instead of actions

src/literals.py
```diff
@@ -21,6 +21,9 @@
 CLIENT_PORT = 2379
 PEER_PORT = 2380
 
+INTERNAL_USER = "root"
```
delgod, line 24: is it possible to create `operator` user with `root` role instead of `root` user?

delgod (changes requested):

src/literals.py
reneradoi: Not instead, only additionally. The `root` user always has to be created in etcd (see [here](https://etcd.io/docs/v3.5/op-guide/authentication/rbac/#special-users-and-roles)). We could only have the `operator` user in addition to `root`, which doesn't really make sense from my point of view.

pyproject.toml
```diff
@@ -89,7 +89,7 @@ lint.extend-ignore = [
     "D409",
     "D413",
 ]
-lint.ignore = ["E501", "D107"]
+lint.ignore = ["E501", "D107", "C901"]
```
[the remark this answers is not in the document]
reneradoi: Done.

actions.yaml
reneradoi: This has been reworked: The user actions `get-password` and `set-password` are no longer supported. Instead users can add a secret `admin-password=...`, grant it to the charm and add the secret-id as a configuration option `admin-password=secret_id`.

src/events/etcd.py
```diff
@@ -102,16 +123,56 @@ def _on_cluster_relation_joined(self, event: RelationJoinedEvent) -> None:
...
 
     def _on_leader_elected(self, event: LeaderElectedEvent) -> None:
         """Handle all events in the 'cluster' peer relation."""
         if not self.charm.state.peer_relation:
             self.charm.set_status(Status.NO_PEER_RELATION)
             return
 
+        if self.charm.unit.is_leader() and not self.charm.state.cluster.internal_user_credentials:
+            self.charm.state.cluster.update(
+                {f"{INTERNAL_USER}-password": self.charm.workload.generate_password()}
```
skourta, line 140: Just a notice:
If the user does not specify the password and we generate it then the user will not have a way to get it since we removed the `get-password` action.

src/literals.py
delgod: lets than use `root` word instead of `admin` everywhere

skourta (approved):
Minor comments. Well done @reneradoi

src/events/etcd.py
```diff
@@ -102,16 +123,56 @@ def _on_cluster_relation_joined(self, event: RelationJoinedEvent) -> None:
...
+                # only update admin credentials if the password has changed
+                if new_password != self.charm.state.cluster.internal_user_credentials.get(
+                    INTERNAL_USER
+                ):
+                    logger.debug(f"{INTERNAL_USER_PASSWORD_CONFIG} has changed.")
+                    try:
+                        self.charm.cluster_manager.update_credentials(
+                            username=INTERNAL_USER, password=new_password
+                        )
+                        self.charm.state.cluster.update(
```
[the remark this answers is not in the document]
reneradoi: I believe we need to, for two reasons:
- otherwise we cannot compare if the password has actually changed when we get a `secret_changed` event
- if the secret and config is removed, we would not know the password anymore

src/events/etcd.py
reneradoi: Exactly, but I assume that is how it is designed.

The workaround is to query the application-secret where we store the password ourselves.

src/events/etcd.py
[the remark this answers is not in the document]
skourta: Yes, I believe this is a design decision. The alternative can be:

- Store the URI on the data bag.
- On `secret_changed` if the the id is the same we update the password always.
- If the secret is removed then we regenerate the password.

I do not mind both approaches but I think the one implemented is better in the case of the `tls-private-key` because always assuming the content changed can add an overhead of re-issueing certificates.

NB: I'm assuming we want to handle all juju user secrets the same way.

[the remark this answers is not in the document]
reneradoi: I also think it is a question of resilience: If the secret is - for whatever reason - not granted to our application anymore, we would not be able to retrieve the password. This would break our operator immediately, because not having the password means not being able to generate a new one.

src/literals.py
reneradoi: Yes we do, with this literal `INTERNAL_USER` it is transparent in the code, no hardcoded usernames or so.

config.yaml
```diff
@@ -0,0 +1,11 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+options:
+  admin-password:
```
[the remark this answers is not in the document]
reneradoi: Done, handling of the config option and secret have been reworked to support this secret structure:
```
root: <password>
other-user: <password>
```

src/events/etcd.py
delgod: yes, it is intentional

delgod (approved):

skourta (approved):

PR #12, opened 2024-12-10
[DPE-6160] block charm if password update fails

reneradoi:
This PR adds functionality to the charmed etcd operator to give feedback to users in case a password update fails. It will set the app and unit status to blocked status if:
- the password cannot be updated in etcd
- the secret has not been granted to the charmed etcd application
- the content of the secret is invalid (username other than `root`)


reneradoi:
This will be addressed in https://github.com/canonical/charmed-etcd-operator/pull/50

PR #14, opened 2025-01-06
[DPE-6265] - Add poetry plugin export to dependencies

skourta:
This pull request includes a change to the `pyproject.toml` file. The change adds a new plugin requirement for `poetry-plugin-export`. The plugin is no longer installed by default starting in poetry 2.

Changes to dependencies:

* [`pyproject.toml`](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R24-R26): Added the `poetry-plugin-export` plugin with a version requirement of ">=1.8".

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,9 @@ pydantic = "^1.10, <2"
 cryptography = ">=42.0.5"
 jsonschema = "*"
 
+[tool.poetry.requires-plugins]
+poetry-plugin-export = ">=1.8"
+
 [tool.poetry.group.format]
 optional = true
 


reneradoi (dismissed):

reneradoi (commented):
It took me some time to figure out why, but `charmcraft pack` didn't work on my local machine(s). I had to add this line to the `charmcraft.yaml` file, just after installing poetry:

`/root/.local/bin/poetry self add poetry-plugin-export`

skourta:
> It took me some time to figure out why, but `charmcraft pack` didn't work on my local machine(s). I had to add this line to the `charmcraft.yaml` file, just after installing poetry:
> 
> `/root/.local/bin/poetry self add poetry-plugin-export`

That is interesting. I did not need to install the plugin explicitly. I just followed [the docs](https://python-poetry.org/docs/cli/#export) and used `>=` instead of `>` because 1.18 is the latest version.

reneradoi:
Let's see how it goes on CI once the build is usable again. I assume we explicitly need to add the line mentioned because it does no longer come by default and `poetry export` should fail if the plugin is not installed on the host.

charmcraft.yaml
```diff
@@ -10,11 +10,11 @@ type: charm
 # see: https://github.com/canonical/data-platform-workflows/issues/246
 bases:
   - build-on:
-    - name: ubuntu
-      channel: "22.04"
+      - name: ubuntu
```
Mehdi-Bendriss, line 13: the previous should not have been an issue yaml-formatting wise of arrays

Mehdi-Bendriss (approved):

reneradoi (approved):

PR #15, opened 2025-01-06
[DPE-5947] Scale up: add new units as etcd cluster members

reneradoi:
The etcd operator needs to integrate new juju units as members to the etcd cluster. Currently it waits on startup until all juju units have joined the peer relation and form an initial cluster. With this PR, this behaviour is changed to:
- starting the etcd cluster with just one member on the juju leader unit
- add subsequent juju units that joined the peer relation as cluster members (“learning-only”)
- promote “learner”-members to full membership once replication from the raft leader has completed

Please also see: https://github.com/canonical/charmed-etcd-operator/wiki/Scaling-up:-Workflow

This PR also adds integration testing for scaling up, including a continuous-writes-functionality (based on the one used in mongodb operator) to make sure the cluster is available for writing data during scaling operations.


src/events/etcd.py
```diff
@@ -68,29 +68,31 @@ def _on_install(self, event: ops.InstallEvent) -> None:
...
+                    self.charm.cluster_manager.enable_authentication()
+                    self.charm.state.cluster.update({"authentication": "enabled"})
+                except (EtcdAuthNotEnabledError, EtcdUserManagementError) as e:
+                    logger.error(e)
+                    self.charm.set_status(Status.AUTHENTICATION_NOT_ENABLED)
+                    return
+        elif (
+            self.charm.state.unit_server.member_endpoint
+            in self.charm.state.cluster.cluster_members
         ):
```
reneradoi, line 87: This will be changed to check cluster membership against etcd itself, not against the endpoint-list in the app databag. It will then use the `member list` command of `etcdctl`, but I didn't want to mess with the implementation of this that @skourta is preparing in the TLS PR. The change will happen after the TLS PR was merged.

tests/integration/ha/continuous_writes.py
```diff
@@ -0,0 +1,50 @@
...
+            count += 1
+            time.sleep(1)
+        except (subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
+            logger.warning(e)
+            time.sleep(1)
+            continue
+
+
+def main():
+    model = sys.argv[1]
```
skourta, line 40: I would use [argparse](https://docs.python.org/3/library/argparse.html) instead.

tests/integration/ha/helpers.py
```diff
@@ -0,0 +1,73 @@
...
+            unit,
+            endpoints,
+            user,
+            password,
+        ]
+    )
+
+
+def stop_continuous_writes() -> None:
+    proc = subprocess.Popen(["pkill", "-9", "-f", "continuous_writes.py"])
```
skourta, line 51: Why use `Popen` here instead of `run`?

tests/integration/ha/test_scaling.py
```diff
@@ -0,0 +1,94 @@
...
+    secret = await get_secret_by_label(ops_test, label=f"{PEER_RELATION}.{app}.app")
+    password = secret.get(f"{INTERNAL_USER}-password")
+
+    # start writing data to the cluster
+    start_continuous_writes(
+        ops_test, app_name=app, endpoints=init_endpoints, user=INTERNAL_USER, password=password
+    )
+
+    # after some time, get the current count
+    time.sleep(30)
```
skourta, line 66: we should aim to reduce the test times. Can we make do with 10s?

skourta (dismissed):
Thank you, René.
I left some questions.

src/core/models.py
```diff
@@ -110,9 +121,9 @@ def __init__(
         self.app = component
 
     @property
-    def initial_cluster_state(self) -> str:
-        """The initial cluster state ('new' or 'existing') of the etcd cluster."""
-        return self.relation_data.get("initial_cluster_state", "")
+    def cluster_state(self) -> str:
+        """The cluster state ('new' or 'existing') of the etcd cluster."""
+        return self.relation_data.get("cluster_state", "")
```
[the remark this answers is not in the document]
reneradoi: I thought we were having snake case everywhere, but you're right, the `admin-password` is not. I will address that.

tests/integration/ha/continuous_writes.py
reneradoi: Was taken over from MongoDB for consistency, but will be reworked anyway when making the continuous writes a fixture.

tests/integration/ha/helpers.py
reneradoi: Same as above: taken over from MongoDB, will be reworked with continuous writes.

tests/integration/ha/test_scaling.py
reneradoi: Will give it a try.

tests/integration/ha/continuous_writes.py
skourta: True

src/core/models.py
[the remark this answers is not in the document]
skourta: Please let me know what the convention you want us to use is and I will convert all my keys to it. I've been using the kebab case because I found some kebab case ones in the code. It also works nicely with YAML files from etcd.

src/managers/cluster.py
```diff
@@ -85,3 +88,67 @@ def update_credentials(self, username: str, password: str) -> None:
             client.update_password(username=username, new_password=password)
         except EtcdUserManagementError:
             raise
+
+    def add_member(self, unit_name: str) -> None:
+        """Add a new member to the etcd cluster."""
+        # retrieve the member information for the newly joined unit from the set of EtcdServers
+        member_name = ""
```
Mehdi-Bendriss, line 95: nit, I believe we can be a bit more concise here and get rid of these 3 variables and quite a bit of indentation:
```
server = next(iter([s for s in self.state.servers if s.unit_name == unit_name]), None)
if not server:
    raise KeyError(f"Peer relation data for unit {unit_name} not found.")

....
```

src/managers/cluster.py
```diff
@@ -85,3 +88,67 @@ def update_credentials(self, username: str, password: str) -> None:
...
+                    client_url=self.state.unit_server.client_url,
+                )
+                cluster_members, member_id = client.add_member_as_learner(member_name, peer_url)
+                self.state.cluster.update(
+                    {"cluster_members": cluster_members, "learning_member": member_id}
+                )
+                logger.info(f"Added unit {unit_name} as new cluster member {member_id}.")
+            except EtcdClusterManagementError:
+                raise
+        else:
```
Mehdi-Bendriss, line 121: under which conditions this may happen?

src/managers/cluster.py
```diff
@@ -85,3 +88,67 @@ def update_credentials(self, username: str, password: str) -> None:
...
+        """Start a cluster member and update its status."""
+        self.workload.start()
+        # this triggers a relation_changed event which the leader will use to promote
+        # a learner-member to fully-voting member
+        self.state.unit_server.update({"state": "started"})
+        if not self.state.cluster.cluster_state:
+            # mark the cluster as initialized
+            self.state.cluster.update(
+                {
+                    "cluster_state": "existing",
```
Mehdi-Bendriss, line 134: Could you extract these state constants in a string enum? future typos can lead to hard to debug issues.

tests/integration/ha/continuous_writes.py
```diff
@@ -0,0 +1,50 @@
...
+import subprocess
+import sys
+import time
+
+from literals import SNAP_NAME
+
+logger = logging.getLogger(__name__)
+
+
+def continuous_writes(model: str, unit: str, endpoints: str, user: str, password: str):
```
Mehdi-Bendriss, line 15: side note for a next PR, as discussed, please make use of the OpenSearch continuous writes structure - it's more robust and flexible.

tests/integration/ha/test_scaling.py
```diff
@@ -0,0 +1,94 @@
...
+
+    cluster_members = get_cluster_members(model, leader_unit, endpoints)
+    assert len(cluster_members) == init_units_count + 2
+
+    # check if data was continuously written to the cluster
+    stop_continuous_writes()
+    final_writes = count_writes(
+        ops_test, app_name=app, endpoints=endpoints, user=INTERNAL_USER, password=password
+    )
+    assert final_writes > init_writes
```
Mehdi-Bendriss, line 94: In a next PR - we should for continuous writes test on 3 things:
1. The writes increased
2. The same count of writes are replicated across all members
3. The last written key ID is as expected and is the same in every unit 

(refer to the continuous writes in opensearch)

Mehdi-Bendriss (changes requested):
Thanks René, great work! I left some comments.

src/core/models.py
```diff
@@ -95,6 +96,16 @@ def client_url(self) -> str:
         """The client connection endpoint for the etcd server."""
         return f"http://{self.ip}:{CLIENT_PORT}"
 
+    @property
+    def member_endpoint(self) -> str:
+        """Concatenate member_name and peer_url."""
+        return f"{self.member_name}={self.peer_url}"
+
+    @property
+    def started(self) -> bool:
```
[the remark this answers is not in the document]
reneradoi: Agree. This was a premature preparation for restarting the service (disaster recovery) that had previously been started. But also in this case, we cannot just start the daemon, we still have to check if it is a cluster member. So the use case is not there and I will remove it.

src/events/etcd.py
```diff
@@ -107,27 +109,31 @@ def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
     def _on_cluster_relation_created(self, event: RelationCreatedEvent) -> None:
         """Handle event received by a new unit when joining the cluster relation."""
         self.charm.state.unit_server.update(self.charm.cluster_manager.get_host_mapping())
```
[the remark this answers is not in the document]
reneradoi: Agree, this will be addressed when implementing failover.

src/managers/cluster.py
reneradoi: This is an edge case that can happen when the leader processes the `relation_joined` event before the new unit has populated its unit-databag. I've seen it when my test machine was taking very long to provision the newly added unit.

src/common/client.py
```diff
@@ -95,6 +148,9 @@ def _run_etcdctl(  # noqa: C901
         auth_password: str | None = None,
         user: str | None = None,
         user_password: str | None = None,
+        member: str | None = None,
+        peer_url: str | None = None,
+        learner: bool = False,
```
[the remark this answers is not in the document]
reneradoi: Handling the `etcdctl` command syntax is very error prone, even more through the subprocess wrapper. For example the arguments and commands/subcommands have to follow a certain order. To avoid handling errors, I'd like to make it as explicit as possible, accepting the additional effort that comes for maintaining the `_run_etcdctl` parameter list.

src/common/client.py
```diff
@@ -112,6 +168,9 @@ def _run_etcdctl(  # noqa: C901
             auth_password: password used for authentication
             user: username to be added or updated in etcd
             user_password: password to be set for the user that is added to etcd
+            member: member name or id, required for commands `member add/update/promote/remove`
+            peer_url: url of a member to be used for cluster-internal communication
+            learner: flag for adding a new cluster member as not-voting member
```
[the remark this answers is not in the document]
reneradoi: In addition to above's comment reg. handling errors, I want to avoid a situation where e.g. credentials are passed to `etcdctl` as part of the additional args-list (e.g. with `username:password`) and the args-list is added to the log. By not having a generic args-list, we can avoid implementing workarounds like [this one in Opensearch](https://github.com/canonical/opensearch-operator/blob/2/edge/lib/charms/opensearch/v0/helper_charm.py#L226).

src/common/client.py
```diff
@@ -83,6 +88,54 @@ def enable_auth(self) -> None:
...
+            #
+            # ETCD_NAME="etcd8"
+            # ETCD_INITIAL_CLUSTER="etcd8=http://10.86.196.119:2380,etcd7=http://10.86.196.232:2380"
+            # ETCD_INITIAL_ADVERTISE_PEER_URLS="http://10.86.196.119:2380"
+            # ETCD_INITIAL_CLUSTER_STATE="existing"
+            #
+            # we need to parse this for the `ETCD_INITIAL_CLUSTER` configuration
+            result = result.split("\n")
+            logger.debug(f"Updated cluster members: {result[0]}")
+            return result[3].split("ETCD_INITIAL_CLUSTER=")[1].strip('"'), result[0].split()[1]
```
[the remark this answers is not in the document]
reneradoi: This would of course be the preferred way. But unfortunately `etcdctl` does not return the required information when `json` output is requested:

```
etcdctl member add etcd17 --peer-urls=http://10.105.253.155:2380 --endpoints=10.105.253.23:2379 --user=root --password=xxx -w=json
{"header":{"cluster_id":12514281256660926174,"member_id":14323290831912880307,"raft_term":2},"member":{"ID":4733236940464715678,"peerURLs":["http://10.105.253.155:2380"]},"members":[{"ID":4733236940464715678,"peerURLs":["http://10.105.253.155:2380"]},{"ID":10353878936597754775,"name":"etcd16","peerURLs":["http://10.105.253.208:2380"],"clientURLs":["http://10.105.253.208:2379"]},{"ID":14323290831912880307,"name":"etcd15","peerURLs":["http://10.105.253.23:2380"],"clientURLs":["http://10.105.253.23:2379"]}]}
```

We only get the `ETCD_INITIAL_CLUSTER` and `MEMBER_ID` in return when we use the `simple` output. I've tried a lot of ways of getting these information otherwise, but this is the only possible way. That's why I parsed it in this case, and put an excessive comment into the code.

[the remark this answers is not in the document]
reneradoi: I've replaced this now with json data, even though it requires some data wrangling because we need to assemble the required information ourselves.

Mehdi-Bendriss (approved):

skourta (approved):

PR #19, opened 2025-01-10
[DPE-6075] Make continuous writes independent of specific units

reneradoi:
The continuous writes function in the integration test suite needs to be able to run independently. Currently it is executed via the `etcdctl` command being run in one of the deployed juju units. This is problematic as we might want to remove this specific unit as part of the scaling/HA tests.

This PR adjusts integration test setup and installs the deb-package `etcd-client` on the host machine. The continuous writes function is then run with this client's `etcdctl` and thus not depending on execution in specific units anymore.

The function is still kept very simple:
- It needs to be started with `start_continuous_writes(endpoints=...)` and stopped with `stop_continuous_writes()`. 
- It can be started with multiple endpoints though and persists through changes of Raft or Juju leadership.
- Use the `assert_continuous_writes_increasing` and `assert_continuous_writes_consistent` helpers to ensure that the writes continued throughout any cluster/HA related operation.

**Note:**
There is no currently maintained Python client lib for etcd (see [issue](https://github.com/etcd-io/etcd/issues/13285)), that's why we keep using `etcdctl` to connect to etcd on each write. Testing with several of the available, unsupported clients failed because they either don't support authentication or cannot be installed to due incompatibilities with protobuf.


Mehdi-Bendriss (changes requested):
Thanks René! Could you adapt the [continuous writes in opensearch](https://github.com/canonical/opensearch-operator/blob/2/edge/tests/integration/ha/continuous_writes.py) to the ETCD context? we should have the same methodology and logic

tests/integration/ha/helpers.py
```diff
@@ -52,22 +45,40 @@ def stop_continuous_writes() -> None:
...
+        with attempt:
+            with open(WRITES_LAST_WRITTEN_VAL_PATH, "r") as f:
+                last_written_value = int(f.read().rstrip())
+
+    for endpoint in endpoints.split(","):
+        last_etcd_value = count_writes(endpoint, user, password)
+        # when stopping the writes, it may happen that data was written to etcd but not to file yet
+        assert (
+            last_written_value <= last_etcd_value
+        ), f"endpoint: {endpoint}, expected value: {last_written_value}, last value in endpoint: {last_etcd_value}."
```
Mehdi-Bendriss, line 84: This "approximately" checks the consistency, we need to be precise here and use equality, as follows:
1. Ensure that the `count of inserted records == last_written_value` (reference to the previous comment on incrementing the keys) 
2. Ensure that `last_written_value == last_etcd_value`

tests/integration/ha/continuous_writes.py
```diff
@@ -3,47 +3,48 @@
...
 
-def continuous_writes(model: str, unit: str, endpoints: str, user: str, password: str):
+def continuous_writes(endpoints: str, user: str, password: str):
     key = "cw_key"
     count = 0
 
     while True:
-        etcd_command = f"""{SNAP_NAME}.etcdctl \
+        etcd_command = f"""etcdctl \
                         put {key} {count} \
```
Mehdi-Bendriss, line 22: Could you increment the key? we need to test on the count - as well as the max key/val

tests/integration/ha/continuous_writes.py
```diff
@@ -3,47 +3,48 @@
...
-            logger.warning(e)
-            time.sleep(1)
-            continue
+        except (subprocess.CalledProcessError, subprocess.TimeoutExpired):
+            pass
+
+        # write last expected written value on disk
+        with open(WRITES_LAST_WRITTEN_VAL_PATH, "w") as f:
+            f.write(str(count))
+            os.fsync(f)
```
Mehdi-Bendriss, line 37: I believe this can be put outside of the loop for better performances

Mehdi-Bendriss (commented):
Thanks René! I left some comments mainly on the consistency part of the writes

tests/integration/ha/helpers.py
reneradoi: Has been adjusted; we now ensure equality of the count, the value stored on disk and the key value.

tests/integration/ha/continuous_writes.py
reneradoi: No need to do this in `etcd`, as updated keys are always stored as a new revision. I updated the `count_writes` function to include the revision number in the return values.

tests/integration/ha/continuous_writes.py
reneradoi: True.

tests/integration/ha/continuous_writes.py
Mehdi-Bendriss: Very nice, thank you!

Mehdi-Bendriss (approved):
Thank you René! Great work!

tests/integration/helpers.py
```diff
@@ -136,3 +124,14 @@ def get_certificate_from_unit(model: str, unit: str, cert_type: TLSType) -> str
         return output
 
     return None
+
+
+async def download_client_certificate_from_unit(
```
skourta, line 129: We might have a somewhat of a similar function that I added with TLS `get_certificate_from_unit`. They're not the same but close

tox.ini
```diff
@@ -63,7 +63,23 @@ pass_env =
...
+    tar
+    mv
+    wget
+    etcdctl
 commands_pre =
     poetry install --only integration
+    # workaround to install etcdctl from release archive
+    # in 22.04 only etcdctl v3.3.25 is included, which is incompatible to v3.4.x
+    # this workaround will be removed when we use 24.04 base
+    sudo apt-get install wget
```
skourta, line 80: `curl` usually comes preinstalled on ubuntu

tests/integration/test_tls.py
```diff
@@ -115,9 +110,6 @@ async def test_tls_enabled(ops_test: OpsTest) -> None:
 @pytest.mark.abort_on_fail
 async def test_disable_tls(ops_test: OpsTest) -> None:
     """Disable TLS on a running cluster and check if it is still accessible."""
-    model = ops_test.model_full_name
-
-    # disable TLS and check if the cluster is still accessible
     etcd_app: Application = ops_test.model.applications[APP_NAME]  # type: ignore
```
skourta, line 113: We should add `start_continuous_writes` in all tests now no?

skourta (approved):
Thank you @reneradoi. Solid work. I just left minor comments and a question.

tests/integration/helpers.py
reneradoi: Yes I saw and considered using that. It has a different purpose though, that's why I chose to keep them separate from another. Now both are very simple, easy to understand and serve one purpose each, not complicating things.

tox.ini
reneradoi: Thanks, didn't know that. It's only a temporary workaround, so I will keep it for now.

tests/integration/test_tls.py
reneradoi: I'd say only mandatory for HA tests. It will get a little complicated with the TLS tests, as the endpoints are constantly changing between the schemes.

PR #26, opened 2025-01-22
[DPE-5966] Implement CA rotation

skourta:
This pull request added CA rotation to the charm.

### TLS Certificate Management Enhancements:

* **New TLS CA Rotation States:**
  - Added `TLSCARotationState` enum to represent different stages of CA rotation (`src/literals.py`).

* **CA Rotation and Cleanup Methods:**
  - Introduced methods in `TLSManager` for setting CA rotation state, checking CA status across servers, and cleaning up old CAs (`src/managers/tls.py`). [[1]](diffhunk://#diff-a2c036f44ae9bbab595aad84b23bc3106a342c5ab12f7b0cf3d26268b51cdc83R136-R204) [[2]](diffhunk://#diff-a2c036f44ae9bbab595aad84b23bc3106a342c5ab12f7b0cf3d26268b51cdc83L57-R102)

* **Event Handling Updates:**
  - Added `CleanCAEvent` to handle CA cleanup events and integrated it into the `TLSEvents` class (`src/events/tls.py`). [[1]](diffhunk://#diff-3dfb4a104b407f7d4b2e07795a63745a09cde228a8dbe86602e7b72ea600fbe4L15-R57) [[2]](diffhunk://#diff-3dfb4a104b407f7d4b2e07795a63745a09cde228a8dbe86602e7b72ea600fbe4R88-R89) [[3]](diffhunk://#diff-3dfb4a104b407f7d4b2e07795a63745a09cde228a8dbe86602e7b72ea600fbe4R229-R243)

### Codebase Improvements:

* **Logging Enhancements:**
  - Improved logging messages to provide more context during rolling restarts and CA rotations (`src/charm.py`). [[1]](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cL68-R78) [[2]](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR178-R202)


src/events/tls.py
```diff
@@ -99,36 +130,65 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
...
+        ]:
+            if not self.charm.tls_manager.is_new_ca_saved_on_all_servers(cert_type):
+                logger.debug("Waiting for all servers to update CA")
+                event.defer()
+                return
+
+        # write certificates to disk
+        self.charm.tls_manager.write_certificate(cert, private_key)  # type: ignore
+
+        # TLS is enabled, New CA added to all servers, and cert updated -> no rolling restart needed
```
reneradoi, line 171: This comment is a little misleading, as we will have rolling restart on the `clean_ca_event`.

src/events/tls.py
```diff
@@ -99,36 +130,65 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
...
+        if tls_state == TLSState.TLS and tls_ca_rotation_state == TLSCARotationState.NEW_CA_ADDED:
+            logger.debug(f"Updating {cert_type.value} certificates with new CA")
+            self.charm.tls_manager.set_ca_rotation_state(
+                cert_type, TLSCARotationState.CERT_UPDATED
+            )
+            self.clean_ca_event.emit(cert_type=cert_type)
```
reneradoi, line 177: I really like the idea of decoupling this clean-up from the actual `CertificateAvailableEvent`!

src/managers/tls.py
```diff
@@ -121,3 +133,72 @@ def delete_certificates(self, cert_type: TLSType) -> None:
...
+        """
+        cas = self._load_trusted_ca(tls_type)
+        # clear cas file
+        cas_path = (
+            self.workload.paths.tls.client_ca
+            if tls_type == TLSType.CLIENT
+            else self.workload.paths.tls.peer_ca
+        )
+        self.workload.remove_file(cas_path)
+        self.add_trusted_ca(cas[-1], tls_type)
```
reneradoi, line 151: I understand that this solution does the job and it would be more complex to do a more qualified removal of the old ca. Please add a comment to explain it for future engineers, as it relies on implicit assumptions.

tests/unit/test_tls.py
```diff
@@ -683,3 +805,234 @@ def test_certificate_expiration():
                 charm.tls_events._on_certificate_available(event)
                 assert charm.state.unit_server.tls_client_state == TLSState.TLS
                 restart_mock.assert_not_called()
+
+
+def test_ca_peer_rotation(certificate_available_context):
+    manager = certificate_available_context.manager
+    peer_relation = certificate_available_context.peer_relation
+    # provider_ca_certificate = certificate_available_context.provider_ca_certificate
```
reneradoi, line 813: Can be removed I guess?

tests/unit/test_tls.py
```diff
@@ -683,3 +805,234 @@ def test_certificate_expiration():
...
+    with (
+        manager,
+        patch("managers.tls.TLSManager._load_trusted_ca", return_value=[]),
+        patch("managers.tls.TLSManager.add_trusted_ca"),
+        patch(
+            "charms.tls_certificates_interface.v4.tls_certificates.TLSCertificatesRequiresV4.get_assigned_certificates",
+            return_value=([client_provider_certificate], requirer_private_key),
+        ),
+        patch("charm.EtcdOperatorCharm._restart"),
+        # patch("ops.framework.BoundEvent.emit"),
```
reneradoi, line 960: Can also be removed I guess?

reneradoi (commented):
Thank you Smail! The PR is implemented in a very clear way, I like that. Also the workflow of the CA rotation is executed nice and clean, without unnecessary complexity.

I only left minor comments regarding commenting the code, which will help us to understand the workflow in the future. Please also update and add the sequence diagram you put into the [TLS spec](https://docs.google.com/document/d/1N9kiats5ozVEeJS9mkM8wLnivy_aEC9xH-ql_vC9h3U/edit?tab=t.0) to the [wiki](https://github.com/canonical/charmed-etcd-operator/wiki).

src/charm.py
```diff
@@ -165,6 +175,31 @@ def _restart_disable_peer_tls(self, _) -> None:
             self.set_status(Status.TLS_PEER_TRANSITION_FAILED)
             raise HealthCheckFailedError("Failed to check health of the member after restart")
 
+    def _restart_ca_rotation(self, _):
+        """Restart callback for CA rotation."""
+        logger.debug("ca rotation restart")
+        self._restart(None)
+        if self.state.unit_server.tls_peer_ca_rotation_state == TLSCARotationState.NEW_CA_DETECTED:
+            self.tls_manager.set_ca_rotation_state(TLSType.PEER, TLSCARotationState.NEW_CA_ADDED)
+        if (
```
Mehdi-Bendriss, line 184: nit:
```suggestion

        if (
```

src/charm.py
```diff
@@ -165,6 +175,31 @@ def _restart_disable_peer_tls(self, _) -> None:
...
+            self.state.unit_server.tls_client_ca_rotation_state
+            == TLSCARotationState.NEW_CA_DETECTED
+        ):
+            self.tls_manager.set_ca_rotation_state(TLSType.CLIENT, TLSCARotationState.NEW_CA_ADDED)
+
+    def _restart_clean_cas(self, _):
```
Mehdi-Bendriss, line 190:
```suggestion
    def _restart_clean_cas(self, _) -> None:
```

src/events/tls.py
```diff
@@ -12,20 +12,49 @@
...
+
+    def __init__(
+        self,
+        handle: Handle,
+        cert_type: TLSType,
+    ):
+        super().__init__(handle)
+        self.cert_type = cert_type
+
+    def snapshot(self) -> dict:
```
Mehdi-Bendriss, line 44: nit: missing type on `dict`

src/events/tls.py
```diff
@@ -12,20 +12,49 @@
...
+        """Snapshot of lock event."""
+        return {"cert_type": self.cert_type.value}
+
+    def restore(self, snapshot: dict):
```
Mehdi-Bendriss, line 48: nit: 
```suggestion
    def restore(self, snapshot: dict[str, str]) -> None:
```

src/managers/tls.py
```diff
@@ -54,40 +54,52 @@ def write_certificate(self, certificate: ProviderCertificate, private_key: Priva
...
             private_key_path = self.workload.paths.tls.peer_key
-            self.add_trusted_ca(ca_cert.raw, client=False)
 
+        self.add_trusted_ca(ca_cert.raw, cert_type)
         self.workload.write_file(private_key.raw, private_key_path)
         self.workload.write_file(certificate.certificate.raw, certificate_path)
         self.set_cert_state(cert_type, is_ready=True)
 
-    def add_trusted_ca(self, ca_cert: str, client: bool = False) -> None:
+    def is_new_ca(self, ca_cert: str, tls_type: TLSType) -> bool:
```
Mehdi-Bendriss, line 66: this is nice

src/events/tls.py
```diff
@@ -99,36 +130,65 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
...
             return
 
-        if (
-            cert_type == TLSType.CLIENT
-            and self.charm.state.unit_server.tls_client_state == TLSState.TLS
-        ):
-            logger.debug("Rotating client certificates")
+        # writing certificate after CA rotation
+        if tls_ca_rotation_state in [
+            TLSCARotationState.NEW_CA_DETECTED,
```
Mehdi-Bendriss, line 160: when would `== NEW_CA_DETECTED` be true? given the `_restart_ca_rotation` wrote `NEW_CA_ADDED`

src/managers/tls.py
Mehdi-Bendriss: Also worth specifying that this solution is based on the assumption that only 2 CAs at most are in the trust store

src/managers/tls.py
```diff
@@ -121,3 +133,73 @@ def delete_certificates(self, cert_type: TLSType) -> None:
...
+                if cert_type == TLSType.PEER
+                else server.tls_client_ca_rotation_state
+            )
+            if server_ca_rotation_state in [
+                TLSCARotationState.NO_ROTATION,
+                TLSCARotationState.NEW_CA_DETECTED,
+            ]:
+                is_ca_stored_on_all_servers = False
+                break
+        return is_ca_stored_on_all_servers
```
Mehdi-Bendriss, line 183: nit - can be simplified and remove the variable:
```suggestion
                return False
                
        return True
```

src/managers/tls.py
```diff
@@ -121,3 +133,73 @@ def delete_certificates(self, cert_type: TLSType) -> None:
...
+            server_ca_state = (
+                server.tls_peer_ca_rotation_state
+                if cert_type == TLSType.PEER
+                else server.tls_client_ca_rotation_state
+            )
+            if server_ca_state in [
+                TLSCARotationState.NEW_CA_DETECTED,
+                TLSCARotationState.NEW_CA_ADDED,
+            ]:
+                is_cert_updated_on_all_servers = False
```
Mehdi-Bendriss, line 202: similar comment

Mehdi-Bendriss (dismissed):
Thanks Smail, great work! Looks good! only few minor nits from my side

src/managers/tls.py
reneradoi: This is the assumption I was referring to :+1:

src/events/tls.py
skourta: This event will be deferred so it will be executed in the next hook as well. In case this event is handled before we acquire the lock the value would be `NEW_CA_DETECTED`.

src/managers/tls.py
skourta: It is this implementation that makes that way. It is not a necessity but in our case since these certs are only used by the server I do not see why we would have more than 2 CAs.

tests/integration/tls/test_ca_rotation.py
```diff
@@ -59,7 +59,7 @@ async def test_build_and_deploy_with_tls(ops_test: OpsTest) -> None:
     endpoints = get_cluster_endpoints(ops_test, APP_NAME, tls_enabled=True)
     leader_unit = await get_juju_leader_unit_name(ops_test, APP_NAME)
 
-    cluster_members = get_cluster_members(model, leader_unit, endpoints, tls_enabled=True)
+    cluster_members = get_cluster_members(endpoints, tls_enabled=True)
```
reneradoi, line 62: Before performing `etcdctl` commands with `get_cluster_members` we need to download the client certificate files via `download_client_certificate_from_unit(ops_test, APP_NAME)` as in [this](https://github.com/canonical/charmed-etcd-operator/blob/main/tests/integration/test_tls.py#L177-L180) example.

skourta: I'm currently fixing it. And adding some asserts on the certificates. I will push the

reneradoi (approved):

PR #29, opened 2025-01-24
[DPE-5965] - Set private key with juju secrets

skourta:
This pull request introduces the ability to set TLS private keys through a config option to pass a juju secret id.
The TLS v4 lib used is forked [here](https://github.com/skourta/tls-certificates-interface/tree/private-key-secret-label).
### Config and literals:
* [`config.yaml`](diffhunk://#diff-d8d0422389f03d783e32e627250fe29834bd09c6361640d1ff00661dd6820034R12-R17): Added a new configuration option `tls-private-key` for specifying the private key for TLS certificates.
* [`src/literals.py`](diffhunk://#diff-959ba637043ab4ac88c6f9f9109022f78a7df1b9a2ab3a0407bff4f124c2db64R36): Introduced `TLS_PRIVATE_KEY_CONFIG` constant to reference the new configuration option.

### Event Handling Updates:
* [`src/events/etcd.py`](diffhunk://#diff-4d4b45129d72bf79a08e70009a59f4ebe0a016b4727117155249c864ae4a1b49R125-R132): Updated event handlers to manage the TLS private key, including fetching and updating the private key from secrets. [[1]](diffhunk://#diff-4d4b45129d72bf79a08e70009a59f4ebe0a016b4727117155249c864ae4a1b49R125-R132) [[2]](diffhunk://#diff-4d4b45129d72bf79a08e70009a59f4ebe0a016b4727117155249c864ae4a1b49R181-R190) [[3]](diffhunk://#diff-4d4b45129d72bf79a08e70009a59f4ebe0a016b4727117155249c864ae4a1b49R214-R245)

### Testing Improvements:
* [`tests/integration/helpers.py`](diffhunk://#diff-e9349e06331f9876f984bcf05671235cfbe546046bccab6ad1abb0b183982b30R129-R149): Added a helper function `add_secret` to facilitate adding secrets during integration tests.
* [`tests/integration/tls/test_private_key_rotation.py`](diffhunk://#diff-858b4ba3b588d303d7396271ac3575bbe9585625ee8b278feb1411524d533a97R1-R232): Added new integration tests to verify the private key rotation functionality and ensure the cluster remains accessible after key changes.
* [`tests/unit/test_tls.py`](diffhunk://#diff-a32fba3e75331a440e32977437daaa8d9c93cb87e72dc5f5e3a543794b0df51bR690-R794): Expanded unit tests to cover the new TLS private key configuration and handling.


src/events/etcd.py
```diff
@@ -195,3 +211,35 @@ def update_admin_password(self, admin_secret_id: str) -> None:
...
+                        secret = self.charm.model.get_secret(
+                            label=self.charm.tls_events.peer_certificate._get_private_key_secret_label()
+                        )
+                        secret.set_content({"private-key": private_key})
+                        secret.get_content(refresh=True)
+                        self.charm.tls_events.peer_certificate._cleanup_certificate_requests()
+                        self.charm.tls_events.peer_certificate._send_certificate_requests()
+                        self.charm.tls_events.client_certificate._cleanup_certificate_requests()
+                        self.charm.tls_events.client_certificate._send_certificate_requests()
+
```
reneradoi, line 243: This whole block of logic should not be our concern. The workflow here should be:
- get secret content
- check for valid format
- call `regenerate_private_key` of the TLS lib

I understand that this is not possible right now as the TLS lib does not support providing an own private key. Unless we wait until [their feature PR](https://github.com/canonical/tls-certificates-interface/pull/300) is available for us, we should at least move this logic out of here and have it in the TLS manager.

Please also add a `todo` comment so that we don't forget to refactor this.

tests/integration/tls/test_private_key_rotation.py
```diff
@@ -0,0 +1,232 @@
...
+    put_key,
+)
+
+logger = logging.getLogger(__name__)
+
+TLS_NAME = "self-signed-certificates"
+NUM_UNITS = 3
+TEST_KEY = "test_key"
+TEST_VALUE = "42"
+CERTIFICATE_EXPIRY_TIME = 90
```
reneradoi, line 33: I believe this can be removed as we don't use it in this test.

tests/integration/tls/test_private_key_rotation.py
```diff
@@ -0,0 +1,232 @@
...
+    logger.info("Integrating peer-certificates and client-certificates relations")
+    await ops_test.model.integrate(f"{APP_NAME}:peer-certificates", TLS_NAME)
+    await ops_test.model.integrate(f"{APP_NAME}:client-certificates", TLS_NAME)
+    await ops_test.model.wait_for_idle(apps=[APP_NAME], status="active", timeout=1000)
+
+
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "large"])
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_tls_enabled(ops_test: OpsTest) -> None:
```
reneradoi, line 64: This test function seems to a duplicate of the `test_tls_enabled` in `test_tls.py`. Do we need it here for any reason? If not, we can remove it because it does not really matter for testing the rotation of the private key(s).

tests/integration/tls/test_private_key_rotation.py
```diff
@@ -0,0 +1,232 @@
...
+
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "large"])
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_set_private_key(ops_test: OpsTest) -> None:
+    """Set a new private key and check if the cluster is still accessible."""
+    model = ops_test.model_full_name
+    logger.info("Getting current private keys")
+    current_private_keys: list[str] = [
+        (await get_secret_by_label(ops_test, label=f"{LIBID}-private-key-{i}"))["private-key"]  # type: ignore
```
reneradoi, line 120: Please add a comment here referencing where the label is generated in the TLS lib's code. If they decide to change that format, our tests might fail and we will go searching.

tests/unit/test_charm.py
```diff
@@ -110,6 +111,27 @@ def test_start():
         state_out = ctx.run(ctx.on.start(), state_in)
         assert state_out.unit_status == ops.BlockedStatus("etcd service not running")
 
+    # non leader waiting promoted
```
reneradoi, line 114: Thank you for adding unit test coverage here and for the peer relation events!

reneradoi (changes requested):
Thank you Smail, I really appreciate the work you did and all the back-and-forth with the TLS team.

Unfortunately the TLS v4 lib has another couple of flaws here. Especially:
- not being able to have separate private keys for multiple use cases
- not being able to provide a user-supplied private key when regenerating it

are major downsides from my point of view. We should only consider the solution implemented here as a workaround, as it introduced potential liabilities to our code when we need to access "protected" methods of the lib to fulfill our use cases.

We as a team should make a decision here if we want to go with the workaround for now or request these features to be delivered by the TLS team and postponing the TLS private key topic until then.

src/events/etcd.py
```diff
@@ -195,3 +211,35 @@ def update_admin_password(self, admin_secret_id: str) -> None:
...
         except (ModelError, SecretNotFoundError) as e:
             logger.error(e)
+
+    def update_private_key(self, private_key_id: str) -> None:
+        """Update the private key in etcd."""
+        logger.debug("Updating TLS private key.")
+        try:
+            if secret_content := get_secret_from_id(self.charm.model, private_key_id).get(
+                "private-key"
+            ):
```
Mehdi-Bendriss, line 221: imo, it would be easier to negate on the condition and return to reduce the indentation and nestedness levels here. 
The exceptions are only raised by `get_secret_from_id` i think?

src/events/etcd.py
```diff
@@ -195,3 +211,35 @@ def update_admin_password(self, admin_secret_id: str) -> None:
...
+                private_key = (
+                    secret_content
+                    if re.match(r"(-+(BEGIN|END) [A-Z ]+-+)", secret_content)
+                    else base64.b64decode(secret_content).decode("utf-8").strip()
+                )
+
+                if not re.match(r"(-+(BEGIN|END) [A-Z ]+-+)", private_key):
```
Mehdi-Bendriss, line 228: I believe we should also add this validation on `_config_changed` and put the charm in a blocked state if the validation fails.

tests/integration/tls/test_private_key_rotation.py
```diff
@@ -0,0 +1,232 @@
...
+    assert (
+        get_key(
+            endpoints,
+            user=INTERNAL_USER,
+            password=password,
+            key=TEST_KEY,
+            tls_enabled=True,
+        )
+        == TEST_VALUE
+    )
```
Mehdi-Bendriss, line 177: nit: missing failure text on assert

Mehdi-Bendriss (commented):
Thanks Smail! great work! 
My main concern is on the lack of 2 separate private keys, as per the spec.
In addition to this, we should avoid the `# type: ignore` hint, it pollutes the code.

config.yaml
```diff
@@ -8,4 +8,10 @@ options:
       Configure the internal system user and it's password. The password will
       be auto-generated if this option is not set. It is for internal use only
       and SHOULD NOT be used by applications. This needs to be a Juju Secret URI pointing
-      to a secret that contains the following content: `root: <password>`.
\ No newline at end of file
+      to a secret that contains the following content: `root: <password>`.
+  
+  tls-private-key:
```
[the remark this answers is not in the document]
skourta: Yes, that is how it was initially implemented ([see](https://github.com/canonical/charmed-etcd-operator/pull/29/commits/402f5ae72dc1b341d3526185d28fe0db2367bfae)) then I changed it to one due to the library limits. The TLS team is working on this. 
How about I switch it back to use 2 config options but the underlying secret will be the same until the TLS team PR is merged and the library is updated?

src/events/etcd.py
skourta: You are right. I followed the same coding style used for the `update_admin_password`. I will switch it.

src/events/etcd.py
skourta: `_on_config_changed` calls this function to update the private key so we get validation in both config changed and secret changed.

src/events/etcd.py
```diff
@@ -113,11 +122,14 @@ def _on_start(self, event: ops.StartEvent) -> None:
 
     def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
         """Handle config_changed event."""
+        if tls_private_key_id := self.charm.config.get(TLS_PRIVATE_KEY_CONFIG):
+            self.update_private_key(tls_private_key_id)  # type: ignore
```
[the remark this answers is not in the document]
skourta: Because the return type of `get` is `(bool | int | float | str | None)`.

tests/integration/tls/test_private_key_rotation.py
skourta: My thinking was that we need to validate that TLS is enabled and used before we do anything related to TLS.

src/events/etcd.py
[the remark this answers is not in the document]
skourta: I removed all of them

tests/integration/tls/test_private_key_rotation.py
reneradoi: ok

src/events/etcd.py
```diff
@@ -195,3 +220,60 @@ def update_admin_password(self, admin_secret_id: str) -> None:
...
+            if tls_type == TLSType.PEER
+            else self.charm.tls_events.client_certificate
+        )
+        try:
+            secret_content = get_secret_from_id(self.charm.model, private_key_id).get(
+                "private-key"
+            )
+        except (ModelError, SecretNotFoundError) as e:
+            logger.error(e)
+            return
```
Mehdi-Bendriss, line 245: This looks risky. This effectively fails silently. I believe we should let the exception propagate and put the charm in a blocked state when this happens -- mainly `on_config_changed`

src/events/etcd.py
```diff
@@ -195,3 +220,60 @@ def update_admin_password(self, admin_secret_id: str) -> None:
...
+
+        if secret_content is None:
```
Mehdi-Bendriss, line 247: Under which conditions would `secret_content is None` be true, given we are catching a `SecretNotFoundError` previously?

Mehdi-Bendriss (commented):
Thanks Smail, I left a couple of comments.

src/events/etcd.py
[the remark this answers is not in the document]
Mehdi-Bendriss: In general, if a method returns a value as an indication of a side effect, we should  rely on the return value of the said method

src/events/etcd.py
skourta: @reneradoi this would be relevant for updating the admin password as well.

reneradoi: I agree with Mehdi, in case of failure there needs to be some feedback to the user. This is more difficult for secrets than for an action. Most sane option (without blocking the workload) is to keep everything running and set the unit status to blocked.

For the admin password, I had prepared a PR for this (see [#12](https://github.com/canonical/charmed-etcd-operator/pull/12)). Mykola asked for holding this back until we have the status implementation figured out. His reasoning was, that the blocking status might get overwritten by something else and the user feedback is gone then.

We could keep the logic here as it is for now and update #12 for the private keys, imo.

config.yaml
```diff
@@ -8,4 +8,14 @@ options:
...
+  
+  tls-peer-private-key:
+    type: secret
+    description: |
+      A Juju secret URI of a secret containing the private key for peer-to-peer TLS certificates.
+  
+  tls-client-private-key:
+    type: secret
+    description: |
+      A Juju secret URI of a secret containing the private key for server-to-client TLS certificates.
```
Mehdi-Bendriss:
```suggestion
      A Juju secret URI of a secret containing the private key for client-to-server TLS certificates.
```

config.yaml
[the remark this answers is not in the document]
Mehdi-Bendriss: Note: We are going with the fork while the TLS team merges the PR addressing this

Mehdi-Bendriss (dismissed):
Thanks Smail! Great work! (just a small typo to fix)

reneradoi (approved):
Thank you Smail!

Mehdi-Bendriss (approved):

PR #32, opened 2025-01-30
[DPE-6341] Upgrade to 24.04 base and etcd v3.5.18

reneradoi:
This PR upgrades the etcd operator to run on Ubuntu 24.04 base image and deploy etcd in version 3.5.18.

Besides those changes, it includes the following adjustments / fixes:

- **ClusterManager:** cluster membership operation cannot be executed against any cluster member anymore in etcd 3.5.18, use all cluster endpoints instead
- **Integration tests:** `ops_test.build_charm` is deprecated for 24.04 in data-platform-workflows (see deprecation notice [here](https://github.com/canonical/data-platform-workflows/blob/main/python/pytest_plugins/pytest_operator_cache/deprecation_notice.md?plain=1)). Therefore we now build the charm outside of the tests when running the tests locally (see [here](https://github.com/canonical/charmed-etcd-operator/blob/upgrade-to-2404-base/tox.ini#L87)), and assume the charm file exists in the tests.
- **Integration test for scaling:** Avoid instability in CI run by setting `raise_on_error=False` in `wait_for_idle` when scaling down. The `member remove` operation might result in errors on the `storage_detaching` hook if the member cannot be removed at the specific moment (desired behaviour). The operation might succeed on the next run, but the test would fail then.

### DIFF
--- a/tests/integration/ha/test_scaling.py
+++ b/tests/integration/ha/test_scaling.py
@@ -38,7 +38,7 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
     etcd_charm = await ops_test.build_charm(".")
 
     # Deploy the charm and wait for active/idle status
-    await ops_test.model.deploy(etcd_charm, num_units=1)
+    await ops_test.model.deploy(etcd_charm, base="ubuntu@24.04", num_units=1)
     await ops_test.model.wait_for_idle(apps=[APP_NAME], status="active", timeout=1000)
 
     assert len(ops_test.model.applications[APP_NAME].units) == 1
--- a/tests/integration/ha/test_scaling.py
+++ b/tests/integration/ha/test_scaling.py
@@ -3,6 +3,7 @@
 # See LICENSE file for licensing details.
 
 import logging
+import time
 
 import pytest
 from pytest_operator.plugin import OpsTest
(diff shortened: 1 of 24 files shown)


tests/integration/ha/test_scaling.py
```diff
@@ -114,6 +113,8 @@ async def test_scale_down(ops_test: OpsTest) -> None:
         status="active",
         wait_for_exact_units=init_units_count - 1,
         wait_for_active=True,
+        # if the cluster member cannot be removed immediately, the `storage_detaching` hook might fail temporarily
+        raise_on_error=False,
```
Mehdi-Bendriss, line 117: I'm a bit worried that we would mask unrelated hooks' errors, what about using the `idle_period` to a minute or so?

Mehdi-Bendriss (commented):
Thanks René, I have a question / concern on `raise_on_error`

tests/integration/ha/test_scaling.py
reneradoi: That does not help in this case, as `wait_for_idle` immediately raises if a unit goes to error (which it can in `storage_detaching`).

Side note: Once PR #29 gets merged, we will move away from `wait_for_idle` and no longer have the issue. Until then it is just a temporary workaround to avoid flaky tests.

Mehdi-Bendriss (approved):

skourta (approved):
LGTM. Thank you @reneradoi

PR #36, opened 2025-02-04
[DPE-6342] trigger leader transfer on restart

### LINKED ISSUE #37: Cluster becomes unavailable when enabling peer TLS
When testing continuous writes for triggering leadership transfer before restarting, I have experienced cluster unavailability when enabling peer TLS. This results in an inconsistency on the continuous writes, where some writes were lost along the way:

`expected value: 358, current value: 358, revision: 356`

This can be seen in [this](https://github.com/canonical/charmed-etcd-operator/actions/runs/13132511313/job/36643489841?pr=36#step:29:956) integration test, where I added continuous writes to the test enabling peer TLS. There are also errors in the log [here](https://github.com/canonical/charmed-etcd-operator/actions/runs/13132511313/job/36643489841?pr=36#step:32:1433) and [here](https://github.com/canonical/charmed-etcd-operator/actions/runs/13132511313/job/36643489841?pr=36#step:32:1546), saying `unhealthy cluster` when performing the health check.

## Steps to reproduce
Either run the integration test `ha/test_ha_on_rolling_restart.py` from [this](https://github.com/canonical/charmed-etcd-operator/pull/36) draft PR, or with manual setup:
- `juju deploy self-signed-certificates --channel=edge`
- `juju deploy ./charmed-etcd_ubuntu@22.04-amd64.charm etcd -n 2`
- wait for active/idle
- `juju integrate etcd:peer-certificates self-signed-certificates:certificates`

## Log output
The following errors can be seen in the logfile of the etcd servers in `/var/snap/charmed-etcd/common/var/log/etcd/etcd.log`.

On the instance first restarted with peer TLS: `tls: failed to verify certificate: x509: certificate signed by unknown authority`

On the instance which gets applied second: `remote error: tls: bad certificate`

## Additional information
My assumption is: We do not wait for the TLS files (esp. CA certificate) to be known in each unit before enabling peer TLS (by running `broadcast_peer_url`). This leads to communication via TLS with a unit that cannot verify the cert.

For testing purposes, I have adjusted the workflow to something like this:
- update config with the `peer-transport-security` properties, but no `https` scheme in the urls
- restart the etcd service
- wait for other units to have restarted
- update the member via `broadcast_peer_url`

This seems to work, but needs more investigation.

reneradoi:
When restarting a member of an etcd cluster, the cluster could experience a short unavailability if the member is the current raft leader of the cluster. Until a new leader was elected (or rather: for the notice period of the other cluster members), no write operation can be performed. To avoid this, we should conduct a leader-transfer on the etcd cluster before initiating a restart on the current raft leader. This PR provides implementation and test coverage for this.

The following changes are introduced:

**ClusterManager:**
- add a new method `move_leader_if_required()` that checks if the member is currently the leader and, if so, transfers leadership to another cluster member
- run the `move_leader_if_required()` method as part of the `restart_member()`

**Restart callback functions in charm.py:**
- when enabling or disabling peer TLS: execute `move_leader_if_required()` before broadcasting the member's updated peer URL (fix for https://github.com/canonical/charmed-etcd-operator/issues/37)
- all other cases: execute `move_leader_if_required()` directly before restarting the workload (as part of `cluster_manager.restart_member()`)

**Integration tests:**
- add a new test `test_ha_on_rolling_restart.py` to ensure continuous writes on restarts when enabling or disabling peer TLS
- add a logfile output to continuous writes for easier debugging missed writes
- move away from the `wait_for_cluster_formation()` helper introduced in https://github.com/canonical/charmed-etcd-operator/pull/28 (no longer necessary because we can use `wait_until`)
- adjust certificate validity from 1m to 3m in `test_certificate_expiration` because in some cases the rolling restart did not complete before the certificate expired (test results flickering)


skourta (dismissed):
LGTM. Thank you @reneradoi

skourta:
@reneradoi Please add the `log_file` to the `.gitignore`

skourta (dismissed):
Thanks again

skourta (dismissed):

src/charm.py
```diff
@@ -105,6 +105,9 @@ def _restart_enable_peer_tls(self, _) -> None:
         """Enable peer TLS."""
         logger.debug("Peer TLS custom callback")
 
+        # in case of peer TLS we need to move the leader before broadcasting membership updates
```
Mehdi-Bendriss, line 108: Can you explain why? Also I believe by having it as part of the `restart_member` it effectively has the same behavior of broadcasting the peer data _after_ the current hook completes

src/charm.py
```diff
@@ -120,7 +123,7 @@ def _restart_enable_peer_tls(self, _) -> None:
 
         # write config and restart workload
         self.config_manager.set_config_properties()
-        if not self.cluster_manager.restart_member():
+        if not self.cluster_manager.restart_member(move_leader=False):
```
Mehdi-Bendriss, line 126: Why not always move the leader on restart?

src/managers/cluster.py
```diff
@@ -271,3 +275,23 @@ def select_new_leader(self) -> str:
             raise ValueError("member list command failed")
         member_list.pop(self.state.unit_server.member_name, None)
         return next(iter(member_list.values())).id
+
+    def move_leader(self) -> None:
```
Mehdi-Bendriss: Could you rename this method to imply that we are only moving the leader if the current unit is elected as such? (I understood it as: move the current elected leader regardless of it being the current unit or not).

tests/integration/ha/test_scaling.py
```diff
@@ -139,9 +138,6 @@ async def test_remove_raft_leader(ops_test: OpsTest) -> None:
     await ops_test.model.applications[app].add_unit(count=1)
     await wait_until(ops_test, apps=[app], wait_for_exact_units=3, idle_period=60)
 
-    # we need to wait for all members to be promoted to full-voting member before scaling down
-    await wait_for_cluster_formation(ops_test, app)
```
Mehdi-Bendriss, line 143: Instead of blocking until it forms (as you rightfully deleted it) - can we have a check that checks on the current membership that it was correctly updated with the new unit?

Mehdi-Bendriss (commented):
Thanks René! Left a couple of questions

src/charm.py
skourta: This was a bug fix. We need to move the leader before we initiate the logic to enable TLS. I initially implemented it as you described but the issue persisted. The fix only worked when the leader was moved before the peer URL was broadcasted.

src/charm.py
skourta: We can but it would be unnecessary here as the leader was already moved on line 109.

src/charm.py
reneradoi: Thank you for the explanation, Smail!

For reference: https://github.com/canonical/charmed-etcd-operator/issues/37

As Smail described, this is due to TLS workflow. It was a finding of the integration tests added in this PR.

src/managers/cluster.py
reneradoi: will do

tests/integration/ha/test_scaling.py
reneradoi: sure, will do

src/charm.py
reneradoi: Moving the leader here would fail because the membership information for this cluster member has already been updated in `broadcast_peer_url()`. This is also described in the [issue](https://github.com/canonical/charmed-etcd-operator/issues/37) referenced above.

Mehdi-Bendriss: Yes - i'm asking why we should not move leadership *always* in restarts - in general

reneradoi: You are right, we should always move leadership in restarts. But in case of enabling/disabling peer TLS, we need to move leadership before broadcasting the new peer url. If we would broadcast first, the member might not be able to communicate with the rest of the cluster anymore (and thus, not able to move leadership), until restarting. That's why in this case, `move_leader()` is some lines above before `restart_member()`.

Mehdi-Bendriss (approved):

src/charm.py
Mehdi-Bendriss: Thanks for clarifying!

skourta (approved):

PR #38, opened 2025-02-07
[DPE-6539] Enable storage reuse

reneradoi:
Re-using previously used storage volumes in `charmed-etcd` is possible under the following conditions:
- An existing database in `/var/snap/charmed-etcd/common/var/lib/etcd/member` can be used with a new `charmed-etcd` application or an existing application that is scaled to zero units.
- There is no way of using an existing database with previous membership information, but new IP address. Therefore, the cluster membership data has to be purged/renewed on startup of the first cluster member. This is because membership information is also stored in the database itself; when loading the existing database, membership information is also loaded/overwritten.
- It is not possible to use an existing database for unit `n+1` and only catch up on the diff from the leader. Additional units/cluster members can only be added by transferring the complete database from the leader.
- When deploying a new `charmed-etcd` application with an existing database/existing storage, the previously used root password has to be supplied by the user, before deploying etcd. This is required because the password is stored in the database itself, and if not provided, the `charmed-etcd` operator can not perform its operations on the database. The password has to be provided as a secret and configured to the operator via the `system_users` config option.

This PR includes the necessary changes to enable storage reuse in etcd.

**EtcdEvents**
- add event handler for `storage_attached`: fix the permissions of the data directory to be able to re-install the snap on existing storage
- `on_start`: if it is a new cluster with an existing database, start the first member and update its membership information (peer-url)
- `on_start`: if the cluster is already existing and there is an existing database in the storage volume, remove the database and start the new unit/member after full data transfer from the leader
- `on_leader_elected`: only generate a new root password if there is no user-provided root password configured

**ConfigManager:**
- introduce new config parameter `force-new-cluster` for re-using an existing database/storage

**Workload:**
- add method `exists()` to check if there already is a database on the storage volume
- add method `remove_directory()` to purge an existing database if required
- add method `exec()` to perform directory permission/ownership changes on the filesystem

**Integration tests:**
new test suite `test_storage_reuse.py` with the following tests:
- `test_attach_storage_after_scale_down`
- `test_attach_storage_after_scale_to_zero`
- `test_attach_storage_after_removing_application`


pyproject.toml
```diff
@@ -54,7 +54,7 @@ pytest-asyncio = "*"
 pytest-operator = "^0.28.0"
 pytest-operator-cache = {git = "https://github.com/canonical/data-platform-workflows", tag = "v29.0.0", subdirectory = "python/pytest_plugins/pytest_operator_cache"}
 pytest-operator-groups = {git = "https://github.com/canonical/data-platform-workflows", tag = "v29.0.0", subdirectory = "python/pytest_plugins/pytest_operator_groups"}
-juju = "3.5.2"
+juju = "==3.6.0"
```
Mehdi-Bendriss, line 57: In a new PR - please keep an eye on juju `3.6.3` stable release (currently in `candidate`) also on CI. We'll need to keep following the `3.6.x` releases to regularly see how the charm behaves in this juju LTS release

src/core/workload.py
```diff
@@ -113,3 +114,28 @@ def remove_file(self, file: str) -> None:
...
         """
         pass
+
+    @abstractmethod
+    def remove_directory(self, directory: str) -> None:
+        """Remove a directory.
+
+        Args:
+            directory (str): Path to the directory.
+        """
```
Mehdi-Bendriss, line 124: nit:
```suggestion
        """
        pass
```

src/events/etcd.py
```diff
@@ -94,7 +107,13 @@ def _on_start(self, event: ops.StartEvent) -> None:
 
         self.charm.config_manager.set_config_properties()
 
-        if self.charm.unit.is_leader() and not self.charm.state.cluster.cluster_state:
+        if not self.charm.state.cluster.cluster_state and self.charm.workload.exists(DATABASE_DIR):
```
Mehdi-Bendriss, line 110: This is fine, perhaps a bit too broad on checking the "dir" existence which could be empty - we should perhaps rather check on the dir not empty `len(os.listdir(DATABASE_DIR)) > 0`

tests/integration/ha/test_storage_reuse.py
```diff
@@ -0,0 +1,266 @@
...
+            key=TEST_KEY,
+            value=TEST_VALUE,
+        )
+        == "OK"
+    )
+    logger.info(f"{new_unit.name} is available again.")
+
+    # check cluster formation after unit with existing storage was added
+    updated_endpoints = get_cluster_endpoints(ops_test, app)
+    cluster_members = get_cluster_members(updated_endpoints)
```
Mehdi-Bendriss, line 106: Could you add an explicit check on the new unit being part of the members?
(applies to all subsequent tests)

tests/integration/ha/test_storage_reuse.py
```diff
@@ -0,0 +1,266 @@
...
+    assert len(cluster_members) == NUM_UNITS, (
```
Mehdi-Bendriss: NUM_UNITS may not be accurate, if running the tests against an existing app

tests/integration/ha/test_storage_reuse.py
```diff
@@ -0,0 +1,266 @@
...
+
+    # add unit with previous storage attached
+    add_unit_cmd = (
+        f"add-unit {app} --model={ops_test.model.info.name} --attach-storage={storage_id}"
+    )
+    return_code, _, _ = await ops_test.juju(*add_unit_cmd.split())
+    assert return_code == 0, f"Failed to add unit with storage {storage_id}"
+
+    new_unit = ops_test.model.applications[app].units[-1]
+    await wait_until(ops_test, apps=[app], wait_for_exact_units=NUM_UNITS, idle_period=60)
```
Mehdi-Bendriss: `NUM_UNITS` may not be accurate, if running the tests against an existing app 
(applies to all subsequent checks on `NUM_UNITS`)

tests/integration/ha/test_storage_reuse.py
```diff
@@ -0,0 +1,266 @@
...
+        endpoints=updated_endpoints, user=INTERNAL_USER, password=password
+    )
+
+
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "large"])
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_attach_storage_after_scale_to_zero(ops_test: OpsTest) -> None:
+    """Make sure storage can be re-attached after removing all units."""
+    app = (await existing_app(ops_test)) or APP_NAME
```
Mehdi-Bendriss: I believe this test is destructive to be ran against an existing application. I would rather have this test ran against "APP_NAME" instead

tests/integration/ha/test_storage_reuse.py
```diff
@@ -0,0 +1,266 @@
...
+    secret = await get_secret_by_label(ops_test, label=f"{PEER_RELATION}.{app}.app")
+    password = secret.get(f"{INTERNAL_USER}-password")
+
+    # remove all units except one - we need to know which storage to attach when scaling up again
```
Mehdi-Bendriss: I believe we don't need to do this, we could already keep the storage ids before destroying all units

tests/integration/ha/test_storage_reuse.py
```diff
@@ -0,0 +1,266 @@
...
+
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "large"])
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_attach_storage_after_scale_to_zero(ops_test: OpsTest) -> None:
+    """Make sure storage can be re-attached after removing all units."""
+    app = (await existing_app(ops_test)) or APP_NAME
+    secret = await get_secret_by_label(ops_test, label=f"{PEER_RELATION}.{app}.app")
+    password = secret.get(f"{INTERNAL_USER}-password")
+
```
Mehdi-Bendriss: Could you add a variable here for storing the current `cluster id` to ompare it after scale-up with the new cluster id which should match

tests/integration/ha/test_storage_reuse.py
```diff
@@ -0,0 +1,266 @@
...
+    await ops_test.model.wait_for_idle(
+        apps=[app],
+        wait_for_exact_units=0,
+        # if the cluster member cannot be removed immediately, the `storage_detaching` hook might fail temporarily
+        raise_on_error=False,
+        timeout=1000,
+    )
+
+    # scale up again re-attaching the storage
+    add_unit_cmd = (
```
Mehdi-Bendriss: This is fine but also an ideal scenario, realistically the user may just add all 3 units with their old storage volumes at once. 
I think it's better if we add all 3 units with their old storage volumes and see how the charm behaves (the assumption being that 2 volumes will be wiped)

tests/integration/ha/test_storage_reuse.py
```diff
@@ -0,0 +1,266 @@
...
+
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "large"])
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_attach_storage_after_removing_application(ops_test: OpsTest) -> None:
+    """Make sure storage can be re-attached to a completely new etcd application."""
+    app = (await existing_app(ops_test)) or APP_NAME
+    secret = await get_secret_by_label(ops_test, label=f"{PEER_RELATION}.{app}.app")
+    password = secret.get(f"{INTERNAL_USER}-password")
+
```
Mehdi-Bendriss, line 205: Similar comment on the cluster_id and checking that the new one is different than the old one. 
But also checking before starting the writes that they have the same data.

Mehdi-Bendriss (commented):
Thanks René! Looks good, I left a few comments

src/core/workload.py
reneradoi: Thank you, not sure why I forgot that here.

tests/integration/ha/test_storage_reuse.py
reneradoi: updated

tests/integration/ha/test_storage_reuse.py
reneradoi: updated

src/events/etcd.py
reneradoi: updated

tests/integration/ha/test_storage_reuse.py
reneradoi: updated

tests/integration/ha/test_storage_reuse.py
reneradoi: If it's too destructive in 2 of 3 cases, then I've removed the option to run this integration test against an existing app completely.

tests/integration/ha/test_storage_reuse.py
reneradoi: updated

tests/integration/ha/test_storage_reuse.py
reneradoi: updated

tests/integration/ha/test_storage_reuse.py
reneradoi: updated

tests/integration/ha/test_storage_reuse.py
reneradoi: sure

Mehdi-Bendriss (dismissed):
Thanks René!

src/workload.py
```diff
@@ -72,3 +76,29 @@ def restart(self) -> None:
...
         path.unlink(missing_ok=True)
+
+    @override
+    def remove_directory(self, directory: str) -> None:
+        rmtree(directory)
+
+    @override
+    def exists(self, path: str) -> bool:
+        if Path(path).exists():
+            return len(os.listdir(path)) > 0
```
skourta: I don't think this should be here. The `exists` function works on files as well. I would also use the `pathlib` version instead.

tests/integration/ha/test_storage_reuse.py
```diff
@@ -0,0 +1,284 @@
...
+    )
+    return_code, _, _ = await ops_test.juju(*add_unit_cmd.split())
+    assert return_code == 0, f"Failed to add unit with storage {storage_id}"
+
+    new_unit = ops_test.model.applications[app].units[-1]
+    await wait_until(ops_test, apps=[app], wait_for_exact_units=init_units_count, idle_period=60)
+
+    # ensure data can be written on the new unit
+    unit_endpoint = get_unit_endpoint(ops_test, unit_name=new_unit.name, app_name=app)
+    assert (
```
skourta, line 92: Writing to the unit will be redirected to the leader of the cluster. We should instead write to another unit and read from it to make sure the change was committed to the new unit.

skourta (commented):
Thank you @reneradoi I left some comments.

tests/integration/ha/test_storage_reuse.py
reneradoi: I'd say that doesn't really matter here: If the write is forwarded, it will only be committed by the cluster leader if the unit is a functional member of the cluster again. This is all we want to ensure here, so this is fine.

skourta (approved):

Mehdi-Bendriss (approved):

PR #42, opened 2025-02-13
[DPE-6444] Add self healing and tests for failover

reneradoi:
This PR adds integration test coverage for failover situations as well as a mechanism to restart non-running workloads.

It includes the following changes:

**EtcdEvents:**
- if the etcd service is not running, it is restarted when the operator is executing the `update_status` hook

**New integration test suite `test_failover.py`:**
- `test_kill_db_process_on_raft_leader`
- `test_freeze_db_process_on_raft_leader`
- `test_restart_db_process_on_raft_leader`
- `test_full_cluster_restart`
- `test_full_cluster_crash`

**Integration test helpers:**
- add method `get_unit_endpoint()` to resolve the etcd endpoint of a specific unit and test against this unit's endpoint
- add method `send_process_control_signal()` to stop/start/freeze/etc the etcd process on a specific unit
- add a parameter `ignore_revision` to `assert_continuous_writes_consistent()` to account for expected missed writes when the cluster is down by not checking if the revision of a key is in sync with the write-counter


tests/integration/ha/test_failover.py
```diff
@@ -0,0 +1,380 @@
...
+        put_key(
+            unit_endpoint,
+            user=INTERNAL_USER,
+            password=password,
+            key=TEST_KEY,
+            value=TEST_VALUE,
+        )
+        == "OK"
+    )
+    logger.info(f"{leader_unit} is available again.")
```
skourta, line 119: nit: I prefer to check if the member is listed in the member list results and add the write as an additional test.

tests/integration/ha/test_failover.py
```diff
@@ -0,0 +1,380 @@
...
+    cluster_members = get_cluster_members(endpoints)
+    assert len(cluster_members) == NUM_UNITS, (
+        f"expected {NUM_UNITS} cluster members, got {len(cluster_members)}"
+    )
+    logger.info(f"Cluster fully formed again with {len(cluster_members)} members.")
+
+    stop_continuous_writes()
+    # By default, etcd uses a 1s election timeout before attempting to replace a lost leader
+    # that's why we will miss writes here, and therefore ignore the revision of the key
+    assert_continuous_writes_consistent(
```
skourta, line 214: We can also test how many revisions are lost. We can calculate how many we expect to lose in 1s and use it.

tests/integration/ha/test_failover.py
```diff
@@ -0,0 +1,380 @@
...
+    cluster_members = get_cluster_members(endpoints)
+    assert len(cluster_members) == NUM_UNITS, (
+        f"expected {NUM_UNITS} cluster members, got {len(cluster_members)}"
+    )
+    logger.info(f"Cluster has come back, all {len(cluster_members)} members joined the cluster.")
+
+    # ensure data is written in the cluster
+    assert_continuous_writes_increasing(endpoints=endpoints, user=INTERNAL_USER, password=password)
+    stop_continuous_writes()
+    assert_continuous_writes_consistent(
```
skourta, line 353: Just curious, did you check how many writes we lost in this scenario?

tests/integration/ha/test_failover.py
```diff
@@ -0,0 +1,380 @@
...
+    password = secret.get(f"{INTERNAL_USER}-password")
+
+    # start writing data to the cluster
+    start_continuous_writes(endpoints=endpoints, user=INTERNAL_USER, password=password)
+    time.sleep(10)
+
+    # axe away the etcd process on all units
+    for unit in ops_test.model.applications[app].units:
+        send_process_control_signal(
+            unit_name=unit.name, model_full_name=ops_test.model_full_name, signal="SIGKILL"
```
skourta, line 396: Is there a significant difference between gracefully and forcefully shutting down the etcd processes?

skourta (dismissed):
Thank you @reneradoi. The tests look good and comprehensive. I left a couple of minor comments but everything looks great already.

tests/integration/ha/test_failover.py
reneradoi: I agree, membership also needs to be tested. I opted for asserting the length of the cluster member list, [just some lines below](https://github.com/canonical/charmed-etcd-operator/blob/add-self-healing/tests/integration/ha/test_failover.py#L106-L108).

tests/integration/ha/test_failover.py
reneradoi: I've considered that, and agreed with Mehdi that this test might be very flakey. It might also depend on the configuration of `election-timeout` how many writes are lost, so we decided to only check for consistency the way it is implemented now.

tests/integration/ha/test_failover.py
reneradoi: It highly depends on the `RESTART_DELAY` for the etcd process. According to the `log_file` output for continuous writes on my local tests, it was around 21 to 23 seconds until restarted. Once restarted, the cluster is available again almost immediately.

tests/integration/ha/test_failover.py
reneradoi: If terminated, all writing operation will be completed. If killed, it's just gone, which could lead to writes lost if not committed by the leader (as described [here](https://etcd.io/docs/v3.6/op-guide/failures/#leader-failure)).

tests/integration/ha/test_failover.py
skourta: Yes, I saw that. It is an implicit membership test.

tests/integration/ha/test_failover.py
skourta: But in terms of the test we are running. It is the same as the `SIGTERM` one in both cases we will lose writes. This still feels redundant to me.

src/events/etcd.py
```diff
@@ -193,7 +193,11 @@ def _on_leader_elected(self, event: LeaderElectedEvent) -> None:
     def _on_update_status(self, event: ops.UpdateStatusEvent) -> None:
         """Handle update_status event."""
         if not self.charm.workload.alive():
-            self.charm.set_status(Status.SERVICE_NOT_RUNNING)
+            if not self.charm.cluster_manager.restart_member():
```
Mehdi-Bendriss, line 193: How does this play out with other hooks potentially stopping the member for valid reasons? e.g: TLS enable

src/events/etcd.py
```diff
@@ -193,7 +193,11 @@ def _on_leader_elected(self, event: LeaderElectedEvent) -> None:
...
+                self.charm.set_status(Status.SERVICE_NOT_RUNNING)
+                return
+
+        self.charm.set_status(Status.ACTIVE)
```
Mehdi-Bendriss, line 197: in another PR: we should avoid setting active statuses in hooks, we should rather *clear* select statuses (until eventually getting the advanced status handling components)

tests/integration/ha/helpers.py
```diff
@@ -98,7 +104,31 @@ def assert_continuous_writes_consistent(endpoints: str, user: str, password: str
...
+                f"endpoint: {endpoint}, expected value: {last_written_value}, current value: {last_etcd_value}"
+            )
+        else:
+            assert last_written_value == last_etcd_value == last_etcd_revision, (
+                f"endpoint: {endpoint}, expected value: {last_written_value}, current value: {last_etcd_value}, revision: {last_etcd_revision}."
+            )
+        logger.info(f"Continuous writes are consistent on {endpoint}.")
+
+
+def send_process_control_signal(unit_name: str, model_full_name: str, signal: str) -> None:
```
Mehdi-Bendriss, line 119: nit: `model_name`?

tests/integration/ha/helpers.py
```diff
@@ -98,7 +104,31 @@ def assert_continuous_writes_consistent(endpoints: str, user: str, password: str
...
+    """Send control signal to an etcd-process running on a Juju unit.
+
+    Args:
+        unit_name: the Juju unit running the process
+        model_full_name: the Juju model for the unit
+        signal: the signal to issue, e.g `SIGKILL`
+    """
+    juju_cmd = f"JUJU_MODEL={model_full_name} juju ssh {unit_name} sudo -i 'pkill --signal {signal} -f {ETCD_PROCESS}'"
```
Mehdi-Bendriss, line 127: JFYI: [juju exec](https://canonical-juju.readthedocs-hosted.com/en/latest/user/reference/juju-cli/list-of-juju-cli-commands/exec/)

tests/integration/ha/test_failover.py
```diff
@@ -0,0 +1,380 @@
...
+
+    # the deployment is only HA if at least 3 units
+    await wait_until(ops_test, apps=[app], wait_for_exact_units=NUM_UNITS)
+
+    endpoints = get_cluster_endpoints(ops_test, app)
+    secret = await get_secret_by_label(ops_test, label=f"{PEER_RELATION}.{app}.app")
+    password = secret.get(f"{INTERNAL_USER}-password")
+
+    # start writing data to the cluster
+    start_continuous_writes(endpoints=endpoints, user=INTERNAL_USER, password=password)
```
Mehdi-Bendriss, line 82: Could you move this to the beginning of each test?

tests/integration/ha/test_failover.py
```diff
@@ -0,0 +1,380 @@
...
+
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "large"])
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_kill_db_process_on_raft_leader(ops_test: OpsTest) -> None:
+    """Make sure the cluster can self-heal when the leader goes down."""
+    app = (await existing_app(ops_test)) or APP_NAME
+
+    # the deployment is only HA if at least 3 units
+    await wait_until(ops_test, apps=[app], wait_for_exact_units=NUM_UNITS)
```
Mehdi-Bendriss: This seems to be redundant, it is already done in `test_build_and_deploy`. 
What we should however do, is check if the current cluster has less than 2 nodes and scale up until 2. (In the case where we are running the tests against an existing cluster, not one from `test_build_and_deploy`)

please refer to [this](https://github.com/canonical/opensearch-operator/blob/2/edge/tests/integration/ha/test_ha.py#L142).

tests/integration/ha/test_failover.py
```diff
@@ -0,0 +1,380 @@
...
+    logger.info(f"initial raft leader: {initial_raft_leader}")
+    leader_unit = initial_raft_leader.replace(app, f"{app}/")
+
+    # axe away the etcd process of the cluster/raft leader
+    send_process_control_signal(
+        unit_name=leader_unit, model_full_name=ops_test.model_full_name, signal="SIGKILL"
+    )
+
+    # now check the availability and formation of the cluster
+    time.sleep(RESTART_DELAY)
```
Mehdi-Bendriss: We should not wait for the previous unit to come up. We should rather:
1. assert that the cluster has now -1 nodes
2. immediately check that a new raft leader has been elected 
3. assert that continuous writes are increasing 
5. then after the 20 sec restart delay, check that the the previous raft leader has rejoined the cluster

Refer to the [opensearch HA](https://github.com/canonical/opensearch-operator/blob/2/edge/tests/integration/ha/test_ha.py#L128) tests logic

tests/integration/ha/test_failover.py
```diff
@@ -0,0 +1,380 @@
...
+
+
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "large"])
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_kill_db_process_on_raft_leader(ops_test: OpsTest) -> None:
+    """Make sure the cluster can self-heal when the leader goes down."""
+    app = (await existing_app(ops_test)) or APP_NAME
+
+    # the deployment is only HA if at least 3 units
```
Mehdi-Bendriss: That is true, but we should also ensure that existing clusters with 2 units can have the integ. tests ran again them

tests/integration/ha/test_failover.py
Mehdi-Bendriss: I believe we should be explicit here and check against the unit id/name, as the check on the `NUM_UNITS` is only relevant in the case of clusters bootstrapped by the integ. tests (`test_build_and_deploy`) it does not take into account the case where we run against an existing 4 nodes cluster for ex.

tests/integration/ha/test_failover.py
```diff
@@ -0,0 +1,380 @@
...
+            key=TEST_KEY,
+            value=TEST_VALUE,
+        )
+        == "OK"
+    )
+    logger.info(f"{leader_unit} is stopped.")
+
+    # ensure data is still written in the cluster
+    assert_continuous_writes_increasing(endpoints=endpoints, user=INTERNAL_USER, password=password)
+
```
Mehdi-Bendriss, line 195: same comment on list of checks to be ran here (applies to all subsequent tests)

tests/integration/ha/test_failover.py
```diff
@@ -0,0 +1,380 @@
...
+    # continue the etcd process
+    send_process_control_signal(
+        unit_name=leader_unit, model_full_name=ops_test.model_full_name, signal="SIGCONT"
+    )
+
+    # ensure the stopped unit is reachable again
+    time.sleep(10)
```
Mehdi-Bendriss: is this necessary?

tests/integration/ha/test_failover.py
```diff
@@ -0,0 +1,380 @@
...
+    leader_unit = initial_raft_leader.replace(app, f"{app}/")
+
+    # freeze the etcd process of the cluster/raft leader
+    send_process_control_signal(
+        unit_name=leader_unit, model_full_name=ops_test.model_full_name, signal="SIGSTOP"
+    )
+
+    # wait until the SIGSTOP fully takes effect
+    time.sleep(10)
+
```
Mehdi-Bendriss, line 176: we should also add a check here to confirm that the unit is not reachable anymore (as SIGSTOP could potentially take forever to take effect)

tests/integration/ha/test_failover.py
```diff
@@ -0,0 +1,380 @@
...
+    # make sure leadership was moved
+    new_raft_leader = get_raft_leader(endpoints=endpoints)
+    logger.info(f"new raft leader: {new_raft_leader}")
+    assert new_raft_leader != initial_raft_leader, (
+        "raft leadership not transferred after freeze of leader"
+    )
+
+    cluster_members = get_cluster_members(endpoints)
+    assert len(cluster_members) == NUM_UNITS, (
```
Mehdi-Bendriss: similar comment as above on explicit checks vs `NUM_UNITS` (applies to all subsequent tests)

tests/integration/ha/test_failover.py
```diff
@@ -0,0 +1,380 @@
...
+    # start writing data to the cluster
+    start_continuous_writes(endpoints=endpoints, user=INTERNAL_USER, password=password)
+    time.sleep(10)
+
+    # axe away the etcd process on all units
+    for unit in ops_test.model.applications[app].units:
+        send_process_control_signal(
+            unit_name=unit.name, model_full_name=ops_test.model_full_name, signal="SIGTERM"
+        )
+
```
Mehdi-Bendriss, line 333: we should also have an explicit check here that all processes on all nodes are down

tests/integration/ha/test_failover.py
```diff
@@ -0,0 +1,380 @@
...
+    await wait_until(ops_test, apps=[app], wait_for_exact_units=NUM_UNITS)
+
+    endpoints = get_cluster_endpoints(ops_test, app)
+    secret = await get_secret_by_label(ops_test, label=f"{PEER_RELATION}.{app}.app")
+    password = secret.get(f"{INTERNAL_USER}-password")
+
+    # start writing data to the cluster
+    start_continuous_writes(endpoints=endpoints, user=INTERNAL_USER, password=password)
+    time.sleep(10)
+
```
Mehdi-Bendriss, line 323: We should update here the restart delay on the systemd units of all nodes to something higher than the default restart delay, to ensure that all units are down within a certain period. 
Reasoning is that unit1 and unit2 may be down, while unit3 takes longer to be down, just enough for unit1 and unit2 to be up again and effectively having a functioning cluster.

The restart delay should then be reset after we confirm that all units are down.

(applies to this test and the full cluster crash one)

Please refer to [here ](https://github.com/canonical/opensearch-operator/blob/2/edge/tests/integration/ha/test_ha.py#L580-L583)

Mehdi-Bendriss (changes requested):
Thanks René! I left some comments

tests/integration/ha/test_failover.py
reneradoi: In case of etcd, the killed member will not be removed from the cluster (if the stop is not performed by our operator). It will still be part of the cluster, but with out-of-date data.

I've adjusted these steps to be:
1. kill process on current leader
2. immediately check for new raft leader (with retry, in case `election_timeout` has not passed)
3. assert continuous writes increasing

tests/integration/ha/test_failover.py
reneradoi: updated

tests/integration/ha/test_failover.py
reneradoi: updated

tests/integration/ha/test_failover.py
reneradoi: There already is a check for this, directly below. It makes sure we can't connect and write data against this endpoint anymore.

tests/integration/ha/test_failover.py
reneradoi: updated

tests/integration/ha/test_failover.py
reneradoi: updated

tests/integration/ha/test_failover.py
reneradoi: updated, check is now performed against the initial number of units that the test starts with (takes scaling up into account).

src/events/etcd.py
reneradoi: Agree, consistent status is still to be addressed. We have a ticket for it already in the backlog: https://warthogs.atlassian.net/browse/DPE-6430

tests/integration/ha/helpers.py
reneradoi: I'd like to stick with the explicit naming, as this was something that failed (with just the "short" name of the model) and I didn't know why.

tests/integration/ha/test_failover.py
reneradoi: Made sure it's always before the actual test starts. It still needs to be after scaling up to two units, because we need to know all relevant endpoints to write against before starting the continuous writes.

tests/integration/ha/test_failover.py
reneradoi: I've added the helper `is_endpoint_up()` to assert this.

tests/integration/ha/test_failover.py
reneradoi: has been added.

src/events/etcd.py
reneradoi: All other hooks should not leave a member stopped until after the hook was completed (also not TLS-related restarts). Therefore we're using the return-value of `restart_member()` to make sure to leave the member healthy.

The only reason a member could be stopped outside of a hook is in case of errors. The restart here would help to get out of these.

tests/integration/ha/test_failover.py
reneradoi: has been removed.

tests/integration/ha/test_failover.py
```diff
@@ -0,0 +1,416 @@
...
+    logger.info(f"Waiting {RESTART_DELAY_PATCHED}s for service restarts.")
+    time.sleep(RESTART_DELAY_PATCHED)
+
+    # now check the availability and formation of the cluster
+    cluster_members = get_cluster_members(endpoints)
+    assert len(cluster_members) == init_units_count, (
+        f"expected {init_units_count} cluster members, got {len(cluster_members)}"
+    )
+    logger.info(f"Cluster has come back, all {len(cluster_members)} members joined the cluster.")
+
```
Mehdi-Bendriss, line 349: We should reset the restart delay to the default value on the systemd units in this and the following test

Mehdi-Bendriss (commented):
Thanks René, looks good! 1 minor comment left from me

tests/integration/ha/test_failover.py
reneradoi: Thank you, I missed that.

Mehdi-Bendriss (approved):

skourta (approved):

PR #44, opened 2025-02-24
[DPE-6568] add self-healing for network cuts

reneradoi:
With this PR we add self-healing functionality for network cuts with changed ip-address to the etcd-operator.

**EtcdEvents:**
- the `config_changed` handler will update the etcd cluster configuration and request new TLS certificates if it detects a change of ip address
- the `peer_relation_changed` handler on the Juju leader unit will by notified by an updated ip address in the peer unit databag and will update the cluster state information in the application databag

**Integration tests:**
added a new integration test file `test_network_cut.py` to conduct these tests:
- `test_network_cut_on_raft_leader_without_ip_change`
- `test_network_cut_on_raft_leader_with_ip_change`

added the following test helpers: 
- `hostname_from_unit`, 
- `ip_address_from_unit` 
- `get_controller_hostname`
- `cut_network_from_unit_with_ip_change`
- `cut_network_from_unit_without_ip_change`
- `restore_network_for_unit_with_ip_change`
- `restore_network_for_unit_without_ip_change`
- `is_unit_reachable`

**Notes and fixes:**
- updated the `rolling_ops` lib to v0.8
- updated `broadcast_peer_url()` in `EtcdClient` to use the client's url property instead of the `endpoints` parameter passed to the method (for consistency with all other methods)
- updated `member` property of the `ClusterManager` to query the member list from all cluster endpoints, not just the current unit

### DIFF
--- a/src/common/client.py
+++ b/src/common/client.py
@@ -359,7 +359,7 @@ class EtcdClient:
         logger.debug("Health check passed.")
         return True
 
-    def broadcast_peer_url(self, endpoints: str, member_id: str, peer_urls: str) -> None:
+    def broadcast_peer_url(self, member_id: str, peer_urls: str) -> None:
         """Broadcast the peer URL to all units in the cluster.
 
         Args:
@@ -370,7 +370,7 @@ class EtcdClient:
         self._run_etcdctl(
             command="member",
             subcommand="update",
-            endpoints=endpoints,
+            endpoints=self.client_url,
             auth_username=self.user,
             auth_password=self.password,
             member=member_id,
--- a/src/common/client.py
+++ b/src/common/client.py
@@ -363,7 +363,6 @@ class EtcdClient:
         """Broadcast the peer URL to all units in the cluster.
 
         Args:
-            endpoints (str): The endpoints to run the command against.
             member_id (str): The member ID to broadcast the peer URL for.
             peer_urls (str): The peer URLs to broadcast.
         """
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -142,6 +142,9 @@ class EtcdEvents(Object):
             if not self.charm.cluster_manager.restart_member(move_leader=False):
                 raise HealthCheckFailedError("Failed to check health of the member after restart")
 
+            if self.charm.unit.is_leader():
+                self.charm.cluster_manager.update_cluster_member_state()
+
             # update tls certificates with new ip address
             if self.charm.state.unit_server.tls_client_state in (
                 TLSState.TO_TLS,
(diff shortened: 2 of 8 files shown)


reneradoi:
The integration test for network cut is currently failing because of an issue with the self-hosted runners, see [this link](https://warthogs.atlassian.net/browse/ISD-3026).

skourta (dismissed):
Thank you @reneradoi . LGTM.

src/common/client.py
```diff
@@ -370,7 +370,7 @@ def broadcast_peer_url(self, endpoints: str, member_id: str, peer_urls: str) ->
         self._run_etcdctl(
```
Mehdi-Bendriss, line 369: On the docstring, could you remove the reference to `endpoints`?

src/events/etcd.py
```diff
@@ -126,6 +127,31 @@ def _on_start(self, event: ops.StartEvent) -> None:
...
+            logger.info(f"New ip address: {ip_address}")
+            self.charm.state.unit_server.update({"ip": ip_address})
+
+            # update cluster configuration
+            self.charm.cluster_manager.broadcast_peer_url(self.charm.state.unit_server.peer_url)
+            self.charm.config_manager.set_config_properties()
+
+            # we need to update the client-urls by restarting etcd
+            # after ip change, this member is unavailable, no need to acquire restart lock
+            if not self.charm.cluster_manager.restart_member(move_leader=False):
```
Mehdi-Bendriss, line 142: Can you explain why we're restarting the unit before issuing a new certificate for it if TLS is enabled? 
Would this not make the unit unusable until the rotation happens?

Mehdi-Bendriss (commented):
Thank you René! good work! I have a question

src/events/etcd.py
reneradoi: That is a very good question. I though so, too. But when testing I realized that the peer TLS cert is still considered valid, even though there is a wrong IP in the SANS. Tested it with client TLS, same result.

I've tried to find out why, and if I understand [this function](https://github.com/etcd-io/etcd/blob/3a2cece119b098fe83fdaf36903dc7d9a934c658/client/pkg/transport/listener_tls.go#L192) correct, then ip address mismatch is ignored if DNS names in SANS are valid.

So the refresh of the TLS certs is there basically only for consistency reasons, but does not impact functionality.

src/common/client.py
reneradoi: Thank you, missed that.

src/events/etcd.py
Mehdi-Bendriss: Thanks for the explanation, that's rather surprising

Mehdi-Bendriss (approved):

skourta (approved):
Re-approval. Tests will be stabilized in a future PR.

PR #48, opened 2025-02-28
[DPE-6568] Raise error if enabling auth fails

reneradoi:
Currently the etcd operator does not raise an error if enabling authentication on cluster startup fails. This could happen for example if the `etcdctl` command times out because of high load on the machine.

To make sure we do not start an unprotected cluster, the `on_start` hook should fail if authentication can not be enabled.

The changes in this PR include:

**EtcdEvents:**
- fail the `on_start` hook for the first unit in the cluster, if the admin user can not be created or authentication can not be enabled in the cluster
- fail the `on_start` hook for subsequent units if authentication is not enabled

**Unit tests:**
- add new test cases for unhappy path (when enabling authentication fails) on startup

### DIFF
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -92,7 +92,7 @@ class EtcdEvents(Object):
             self.charm.set_status(Status.SERVICE_NOT_INSTALLED)
             return
 
-    def _on_start(self, event: ops.StartEvent) -> None:
+    def _on_start(self, event: ops.StartEvent) -> None:  # noqa: C901
         """Handle start event."""
         tls_transition_states = [TLSState.TO_TLS, TLSState.TO_NO_TLS]
         if (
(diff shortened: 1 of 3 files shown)


src/events/etcd.py
```diff
@@ -92,7 +92,7 @@ def _on_install(self, event: ops.InstallEvent) -> None:
             self.charm.set_status(Status.SERVICE_NOT_INSTALLED)
             return
 
-    def _on_start(self, event: ops.StartEvent) -> None:
+    def _on_start(self, event: ops.StartEvent) -> None:  # noqa: C901
```
Mehdi-Bendriss, line 95: heh, we're already getting there :D

src/events/etcd.py
```diff
@@ -124,13 +124,24 @@ def _on_start(self, event: ops.StartEvent) -> None:
...
         elif (
             self.charm.state.unit_server.member_endpoint
             in self.charm.state.cluster.cluster_members
         ):
             # this unit has been added to the etcd cluster
+            if not self.charm.state.cluster.auth_enabled:
+                if self.charm.unit.is_leader():
+                    # if enabling auth failed on first cluster startup, we want to retry now
+                    try:
+                        self.charm.cluster_manager.enable_authentication()
```
Mehdi-Bendriss, line 137: Could you add a 3-attempts retry with a 1sec backoff prior to definitively failing?

Mehdi-Bendriss (commented):
Thank you René! I have a small comment but LGTM

src/events/etcd.py
reneradoi: updated

Mehdi-Bendriss (approved):

skourta (approved):
Thank you @reneradoi

skourta:
Would you mind updating the PR title with the DPE number before merging?
Thank you

PR #49, opened 2025-03-03
[DPE-6700] stabilization and add test for forcefull restart with deleted data

reneradoi:
This PR brings fixes for a couple of edge cases, adjustments to some existing integration tests and a new integration test.

The main changes are the following:

**EtcdClient:**
- Only append client TLS file parameters to the `etcdctl` commands if the files actually exist. When transitioning to/from TLS, it can happen that some cluster members already have `https` endpoints, but a unit does not yet/not anymore have the client TLS files available. In these cases, the `etcdctl` commands failed with `Error: open /var/snap/charmed-etcd/common/tls/client.pem: no such file or directory`.

**EtcdEvents:**
- `on_start`: When reusing storage, a previously existing database file should only be deleted if the workload has not been started yet.
- `on_start`: Eliminate an edge case for storage reuse, when multiple units are started at exactly the same time. This could create a race condition where all of them try to initialize the cluster, which can happen only once. This is now prevented by only allowing the juju leader to initialize the cluster.
- `on_peer_relation_departed`: only allow updating the cluster members in the application databag if the leaders workload is still running. This mitigates an edge case where concurring unit removals bring juju leadership changes (even in the state of removal), which can lead to updates on already cleaned up data.
- `on_leader_elected`: run `update_cluster_member_state()` to reflect changes to the cluster membership due to unit removal

**ClusterManager:**
- remove a duplicate check for cluster availability in the `remove_member` logic
- only `update_cluster_member_state()` if the cluster has already been initialized

**Integration tests:**
- add a one-time retry to the `continuous writes` if the result was not `OK`. This eliminates flickering test results when a write can not be committed during raft leadership change. The retry is limited to one attempt after one second.
- add a retry to the `is_unit_reachable()` helper, which eliminates failing tests if the network disconnect takes longer than 5 `ping` executions
- add a new failover test `test_restart_raft_leader_after_deleting_database_file`
- temporarily skip `test_network_cut_on_raft_leader_without_ip_change` due to the issue with github runners causing frequent failures of the `lxc config` changes for the network cut (also added a comment with ticket reference)
- some other small fixes to make the integration test runs on CI more stable

### DIFF
--- a/src/common/client.py
+++ b/src/common/client.py
@@ -260,6 +260,7 @@ class EtcdClient:
             if use_input:
                 args.append("--interactive=False")
             # we append the TLS params whenever we find a client certificate
+            # todo: this is not substrate-agnostic
             if os.path.exists(f"{TLS_ROOT_DIR}/client.pem"):
                 args.append(f"--cert={TLS_ROOT_DIR}/client.pem")
                 args.append(f"--key={TLS_ROOT_DIR}/client.key")
--- a/src/common/client.py
+++ b/src/common/client.py
@@ -6,6 +6,7 @@
 
 import json
 import logging
+import os
 import subprocess
 from typing import Tuple
 
@@ -258,7 +259,8 @@ class EtcdClient:
                 args.append(f"-w={output_format}")
             if use_input:
                 args.append("--interactive=False")
-            if "https" in endpoints:
+            # we append the TLS params whenever we find a client certificate
+            if os.path.exists(f"{TLS_ROOT_DIR}/client.pem"):
                 args.append(f"--cert={TLS_ROOT_DIR}/client.pem")
                 args.append(f"--key={TLS_ROOT_DIR}/client.key")
                 args.append(f"--cacert={TLS_ROOT_DIR}/client_ca.pem")
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -132,8 +132,8 @@ class EtcdEvents(Object):
         ):
             # this unit has been added to the etcd cluster
             if not self.charm.state.cluster.auth_enabled:
+                # failed start hooks on cluster initialization will go here on retry
                 if self.charm.unit.is_leader():
-                    # if enabling auth failed on first cluster startup, we want to retry now
                     try:
                         self.charm.cluster_manager.enable_authentication()
                         self.charm.state.cluster.update({"authentication": "enabled"})
@@ -143,19 +143,21 @@ class EtcdEvents(Object):
                 else:
                     raise EtcdAuthNotEnabledError("Authentication not enabled.")
 
-            if self.charm.workload.exists(DATABASE_DIR):
-                logger.warning(f"Existing database file detected in {DATABASE_DIR}.")
-                # storage cannot be reused on non-leader members
-                try:
-                    self.charm.workload.remove_directory(DATABASE_DIR)
-                    logger.warning(
-                        f"Removed database file from {DATABASE_DIR} to join existing cluster."
-                    )
-                except OSError:
-                    # if removing fails, we cannot start the workload or the member would crash
-                    raise
+            if not self.charm.state.unit_server.is_started:
+                # database files should not be deleted on running units
+                if self.charm.workload.exists(DATABASE_DIR):
+                    logger.warning(f"Existing database file detected in {DATABASE_DIR}.")
+                    # storage cannot be reused on non-leader members
+                    try:
+                        self.charm.workload.remove_directory(DATABASE_DIR)
+                        logger.warning(
+                            f"Removed database file from {DATABASE_DIR} to join existing cluster."
+                        )
+                    except OSError:
+                        # if removing fails, we cannot start the workload or the member would crash
+                        raise
 
-            self.charm.cluster_manager.start_member()
+                self.charm.cluster_manager.start_member()
         else:
             # this unit that has not yet been added to the cluster
             # wait for leader to process `relation_joined` event and add the member to the cluster
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -232,13 +232,16 @@ class EtcdEvents(Object):
         if not self.charm.unit.is_leader():
             return
 
-        logger.debug(f"Removing {event.unit.name} from cluster state in peer relation.")
-        cluster_members = self.charm.state.cluster.cluster_members.split(",")
-        # re-assemble the string without the departing unit
-        updated_cluster_members = ",".join(
-            m for m in cluster_members if event.unit.name.replace("/", "") not in m
-        )
-        self.charm.state.cluster.update({"cluster_members": updated_cluster_members})
+        if self.charm.state.unit_server.is_started:
+            # this must not overwrite already cleaned up application databag
+            # it should only happen if at least this unit's workload is still running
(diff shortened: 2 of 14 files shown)


src/common/client.py
```diff
@@ -258,7 +259,8 @@ def _run_etcdctl(  # noqa: C901
                 args.append(f"-w={output_format}")
             if use_input:
                 args.append("--interactive=False")
-            if "https" in endpoints:
+            # we append the TLS params whenever we find a client certificate
+            if os.path.exists(f"{TLS_ROOT_DIR}/client.pem"):
```
skourta, line 264: This will depend on the workload type, but I don't see a quick way to make it workload agnostic.

src/events/etcd.py
```diff
@@ -230,13 +234,16 @@ def _on_peer_relation_departed(self, event: RelationDepartedEvent) -> None:
...
-        logger.debug(f"Removing {event.unit.name} from cluster state in peer relation.")
-        cluster_members = self.charm.state.cluster.cluster_members.split(",")
-        # re-assemble the string without the departing unit
-        updated_cluster_members = ",".join(
-            m for m in cluster_members if event.unit.name.replace("/", "") not in m
-        )
-        self.charm.state.cluster.update({"cluster_members": updated_cluster_members})
+        if self.charm.state.unit_server.is_started:
+            # this must not overwrite already cleaned up application databag
+            # it should only happen if at least this unit's workload is still running
```
skourta, line 239: Can you please explain why this unit's workload needs to be running? If we removed a unit but the workload of the juju leader crashed for some reason, we should still update the databag no?

tests/integration/helpers.py
```diff
@@ -99,12 +99,15 @@ def get_cluster_id(endpoints: str, tls_enabled: bool = False) -> str:
...
-        raise
+    result = subprocess.getoutput(etcd_command).split("\n")
+    for r in result:
+        member = json.loads(r)
+        try:
+            return member[0]["Status"]["header"]["cluster_id"]
+        except (TypeError, KeyError) as e:
+            logger.warning(e)
+
+    return ""
```
skourta: Can we raise an error instead? This way we are sure in the tests that if we don't get an error then the value is correct.

tests/integration/ha/test_storage_reuse.py
```diff
@@ -89,16 +90,7 @@ async def test_attach_storage_after_scale_down(ops_test: OpsTest) -> None:
 
     # ensure data can be written on the new unit
```
skourta: Outdates comment. Maybe change it to endpoint is healthy?

tests/integration/ha/test_network_cut.py
```diff
@@ -211,9 +220,14 @@ async def test_network_cut_on_raft_leader_with_ip_change(ops_test: OpsTest) -> N
     cut_network_from_unit_with_ip_change(leader_hostname)
 
     # make sure the unit is not reachable from the other units
-    for unit in ops_test.model.applications[app].units:
-        hostname = await hostname_from_unit(ops_test, unit.name)
-        assert not is_unit_reachable(hostname, leader_hostname)
+    remaining_unit_names = (
```
skourta: To avoid looping twice, you can `continue` if the current `unit.name == leader_unit`.

tests/integration/ha/test_network_cut.py
```diff
@@ -98,9 +102,14 @@ async def test_network_cut_on_raft_leader_without_ip_change(ops_test: OpsTest) -
     cut_network_from_unit_without_ip_change(leader_hostname)
 
     # make sure the unit is not reachable from the other units
-    for unit in ops_test.model.applications[app].units:
-        hostname = await hostname_from_unit(ops_test, unit.name)
-        assert not is_unit_reachable(hostname, leader_hostname)
+    remaining_unit_names = (
```
skourta: Same comment as below.

skourta (commented):
Thank you, @reneradoi. I left some comments.

src/common/client.py
reneradoi: Agree. It works for now, but on a different substrate, will require adjustments. I will add a `todo` comment.

src/events/etcd.py
reneradoi: Yes, if the workload of the Juju leader crashed, we do still update the databag. That's why I chose to check for the `is_started` property, not the actual workload status. This property will be removed from the unit databag on the `storage_detaching` hook, that means if it is not there, the unit was already purposefully stopped.

For context, I encountered the following scenario:
- 3 units, all stopped at the same time -> `planned_units == 0`
- the Juju leader runs `storage_detaching` first (no particular reason) and cleans the application databag because it sees `planned_units == 0`
- the Juju leader unit is removed before all of the others are gone, so one of the remaining units is appointed Juju leader
- the new Juju leader already ran its own `storage_detaching`, means has passed the cleanup part
- the new Juju leader then receives `relation_departed` for the previous leader while the workload on the other remaining unit is still running (not stopped yet) -> gets the current cluster members and updates them to the application databag

This issue is very much depending on the timing, but I saw it happen more than once. In a regular application removal scenario, it's not an issue. But if storage is about to be reused, or if it's just a scale-to-zero and back later, this can result in polluted databag on startup.

By checking `is_started` we make sure the update only happens if the leader will remain running or has not yet completed its own `storage_detaching` hook.

skourta: Thank you for the detailed explanation

tests/integration/ha/test_network_cut.py
reneradoi: Thank you for the hint, that makes it easier indeed!

tests/integration/ha/test_network_cut.py
reneradoi: updated

tests/integration/ha/test_storage_reuse.py
reneradoi: updated

tests/integration/helpers.py
reneradoi: updated

skourta (dismissed):

Mehdi-Bendriss (dismissed):
Thanks René!

reneradoi:
To avoid outdated cluster member information in the peer-relation databag, I added a call to `update_cluster_member_state()` to the `on_leader_elected` hook. This hook happens after the `relation_departed` hook, and thereby will ensure the application databag does get updated after the juju leader unit has been removed.

src/managers/cluster.py
```diff
@@ -300,16 +300,17 @@ def move_leader_if_required(self) -> None:
 
     def update_cluster_member_state(self) -> None:
         """Get up-to-date member information and store in cluster state."""
-        client = EtcdClient(
-            username=self.admin_user,
-            password=self.admin_password,
-            client_url=self.state.unit_server.client_url,
-        )
+        if self.state.cluster.cluster_state:
```
Mehdi-Bendriss: nit, I think it's better in this case to have the inverted condition and return - to reduce nestedness

Mehdi-Bendriss (commented):
Thanks René, left a small formatting comment, but looks good otherwise

src/managers/cluster.py
reneradoi: thank you!

Mehdi-Bendriss (approved):

skourta (approved):

PR #50, opened 2025-03-07
[DPE-6430] Add basic status handling

reneradoi:
With this PR, we add basic status handling for units and application. The implementation adheres to the guidelines provided in [Evaluating charm status across multiple components](https://discourse.charmhub.io/t/how-to-set-a-charms-status/11771#heading--evaluate-charm-status-across-multiple-components).

The changed behaviour can be nicely been seen in the integration test runs on CI, for example [this one](https://github.com/canonical/charmed-etcd-operator/actions/runs/13721304582/job/38377439306?pr=50#step:29:786) from line 785 onwards.

The main changes are:

**charm.py:**
- keeping a list of `pending_inactive_statuses`
- observing the `collect_unit_status` and `collect_app_status` events
- do not set status (`TLS_TRANSITION_FAILED`) if we raise an error in the event handler

**EtcdEvents:**
- do not set `ActiveStatus` as this is now part of `CollectStatusEvent` handler
- set status if deferring `on_start` because TLS is not ready or the unit did not join the cluster yet
- set status when encountering errors in `update_admin_password()`

**ClusterManager:**
- new method `compute_cluster_status()`

unrelated changes:
- `literals.py` re-arrange `Status` class in alphabetical order for better readability
- update `tls_certificates` lib from version `4.9` to `4.11`

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -18,7 +18,6 @@ from literals import (
     RESTART_RELATION,
     SUBSTRATE,
     DebugLevel,
-    EtcdClusterState,
     Status,
     TLSCARotationState,
     TLSState,
@@ -217,12 +216,8 @@ class EtcdOperatorCharm(ops.CharmBase):
         Component statuses should be computed in their respective priority.
         """
         # compute cluster status
-        if self.state.unit_server.is_started:
-            if not self.state.cluster.cluster_state == EtcdClusterState.EXISTING.value:
-                event.add_status(Status.CLUSTER_NOT_INITIALIZED.value.status)
-
-            if not self.state.cluster.auth_enabled:
-                event.add_status(Status.AUTHENTICATION_NOT_ENABLED.value.status)
+        for status in self.cluster_manager.compute_component_status():
+            event.add_status(status.value.status)
 
         # compute TLS status
         # todo: add compute logic here
--- a/src/charm.py
+++ b/src/charm.py
@@ -18,6 +18,7 @@ from literals import (
     RESTART_RELATION,
     SUBSTRATE,
     DebugLevel,
+    EtcdClusterState,
     Status,
     TLSCARotationState,
     TLSState,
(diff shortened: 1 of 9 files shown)


src/managers/cluster.py
```diff
@@ -316,3 +317,16 @@ def update_cluster_member_state(self) -> None:
         except Exception as e:
             # we should not have errors here, but if we do, we don't want the error to raise
             logger.warning(f"Error updating the cluster member state: {e}")
+
+    def compute_component_status(self) -> List[Status]:
+        """Compute the Cluster manager's statuses."""
+        status_list = []
+
+        if self.state.unit_server.is_started:
+            if not self.state.cluster.cluster_state == EtcdClusterState.EXISTING.value:
```
Mehdi-Bendriss:
```suggestion
            if self.state.cluster.cluster_state != EtcdClusterState.EXISTING.value:
```

src/charm.py
```diff
@@ -209,6 +208,27 @@ def _restart_clean_cas(self, _) -> None:
...
+    def _on_collect_status(self, event: ops.CollectStatusEvent) -> None:
+        """Compute the current status for this unit.
+
+        Ops framework will choose the highest-priority status and set that as the status.
+        If there are multiple statuses with the same priority, the first one added wins.
+        Component statuses should be computed in their respective priority.
+        """
+        # compute cluster status
+        for status in self.cluster_manager.compute_component_status():
+            event.add_status(status.value.status)
```
Mehdi-Bendriss, line 220: in case this component takes priority over the following ones, we should perhaps return on the first existing status?

Mehdi-Bendriss (commented):
Thanks René! I believe it's going on the right direction of the overall status vision we have

src/charm.py
reneradoi: Returning on the first existing status would mean that other, perhaps more important statuses would not be collected. I would try to avoid that.

According to the `CollectStatusEvent` [docs](https://ops.readthedocs.io/en/latest/reference/ops.html#ops.CollectStatusEvent), "If there are multiple statuses with the same priority, the first one added wins". Meaning:
- the order of status priorities (blocked, maintenance, waiting, active) is defined by ops
- the order of component priorities is defined by the order here in this method
- by collecting all component statuses, we make sure to not miss any whose status priority might be ranked higher

In my point of view, this is a very simple, yet comprehensive logic.

Mehdi-Bendriss: Okay, that sounds good. We can explore more advanced status prioritization once the status impl. spec lands

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):

skourta (approved):

PR #52, opened 2025-03-13
[DPE-6719] Implement s3 backup

reneradoi:
PR to merge functionality for "creating s3 backups" into the backup/restore feature branch.

Main changes:

**S3-interface:**
- add `s3-credentials` relation
- make use of `data_platform_libs.v0.s3` lib
- implement `S3Requirer` handlers

**BackupEvents:**
- add this new class for handling all events of the s3-relation as well as the actions `create-backup` and `list-backups`

**BackupManager:**
- add this new manager for the business logic of backups and restore
- this manager implements the actual interface to s3 object storage

**EtcdClient:**
- new method `create_database_snapshot()`


src/core/models.py
```diff
@@ -215,6 +216,11 @@ def learning_member(self) -> str:
         """
         return self.relation_data.get("learning_member", "")
 
+    @property
+    def s3_credentials(self) -> dict[str, str]:
+        """Get credentials and parameters to access s3 object storage."""
+        return json.loads(self.relation_data.get("s3-credentials", "{}"))
```
skourta, line 222: nit: I prefer `None` over an empty dict

src/events/backup.py
```diff
@@ -0,0 +1,120 @@
...
+
+        # make sure we have all required parameters for writing to the storage
+        required_parameters = ["bucket", "endpoint", "path", "access-key", "secret-key"]
+        s3_parameters = self.s3_requirer.get_s3_connection_info()
+
+        if missing_parameters := [p for p in required_parameters if p not in s3_parameters]:
+            raise KeyError(f"Parameters missing from S3 integrator: {missing_parameters}")
+
+        # Strip whitespaces from all parameters
+        for key, value in s3_parameters.items():
```
skourta, line 69: Just curious to know if this comes from an error you faced during development or just preventative.

src/events/backup.py
```diff
@@ -0,0 +1,120 @@
...
+        self.charm.backup_manager.store_tls_ca_chain(s3_parameters)
+        self.charm.backup_manager.create_bucket(s3_parameters)
+        self.charm.state.cluster.update({"s3-credentials": json.dumps(s3_parameters)})
+
+    def _on_s3_credentials_gone(self, event: CredentialsGoneEvent) -> None:
+        """Handle the removal of the relation with s3-integrator."""
+        if not self.charm.unit.is_leader():
+            return
+
+        self.charm.state.cluster.update({"s3-credentials": ""})
```
skourta: Shouldn't we also clean stored CA?

src/events/backup.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def _exists_preventing_reason(self) -> str:
+        """Check if an action can be executed, if not return error message."""
+        if not self.charm.unit.is_leader():
+            return "Action must be performed on the leader unit."
+
+        if not self.charm.state.cluster.s3_credentials:
+            return "No credentials for object storage available."
+
+        if not self.charm.state.unit_server.is_started:
```
skourta, line 136: Nice one!

src/managers/backup.py
```diff
@@ -0,0 +1,152 @@
...
+    def get_bucket_resource(self, s3_parameters: dict[str, str]) -> Bucket:
+        """Get the Bucket resource from the s3 connection.
+
+        Returns:
+            Bucket: the s3 bucket for uploading/downloading backups
+        """
+        region = s3_parameters.get("region")
+        s3_resource = boto3.resource(
+            "s3",
+            region_name=region,
```
skourta: Just use `s3_parameters.get("region")` here no need for the variable

src/managers/backup.py
```diff
@@ -0,0 +1,152 @@
...
+        return s3_resource.Bucket(s3_parameters["bucket"])
+
+    def create_bucket(self, s3_parameters: dict[str, str]) -> None:
+        """Create bucket if it does not exist yet."""
+        region = s3_parameters.get("region")
+        bucket = self.get_bucket_resource(s3_parameters)
+
+        try:
+            if region:
+                bucket.create(CreateBucketConfiguration={"LocationConstraint": region})
```
skourta, line 65: We can potentially remove the if else by using a param dict and only adding the region if it is not None

skourta (approved):

src/core/models.py
reneradoi: It can't be none, otherwise we can't load it as json.

src/events/backup.py
reneradoi: Tested various "wrong inputs" and realized this needs some cleaning, also for the `/`.

src/core/models.py
skourta: Yes, I meant returning None before loading the json. This is just my personal preference as both `None` and `{}` are falsy.

src/managers/backup.py
reneradoi: updated

src/events/backup.py
reneradoi: Yes, that is true. Thank you, good catch. Also reminds me that we need to store the ca file on all units, not just the leader, because it could happen that there is a leader change between establishing the relation and actually performing the backup/restore action.

src/managers/backup.py
reneradoi: Tested with this:
```
bucket_configuration = {"LocationConstraint": s3_parameters.get("region")}

        try:
            bucket.create(
                CreateBucketConfiguration=bucket_configuration
                if bucket_configuration["LocationConstraint"]
                else None
            )
```
Doesn't work because `Parameter validation failed: Invalid type for parameter CreateBucketConfiguration, value: None, type: <class 'NoneType'>, valid types: <class 'dict'>`.

skourta: I believe something like the following should work.
```
bucket_configuration = {"LocationConstraint": s3_parameters.get("region")} if s3_parameters.get("region") else {}

        try:
            bucket.create(
                CreateBucketConfiguration=bucket_configuration
            )
```
Also, this can be considered nitpicking. It is neither crucial nor important.

src/events/backup.py

src/managers/backup.py
reneradoi: We will keep the `if ... else` for better readability.

actions.yaml
```diff
@@ -0,0 +1,10 @@
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+create-backup:
+  description: Create a database backup and send it to an object storage.
+    S3 credentials are retrieved from a relation with the S3 integrator charm.
+
+list-backups:
+  description: List database backups. 
+    S3 credentials are retrieved from a relation with the S3 integrator charm.
```
Mehdi-Bendriss: nit:
```suggestion
    S3 credentials are retrieved from a relation with the S3 integrator charm.

```

src/events/backup.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def __init__(self, charm: "EtcdOperatorCharm"):
+        super().__init__(charm, key="backup")
+        self.charm = charm
+        self.s3_requirer = S3Requirer(self.charm, S3_RELATION_NAME)
+
+        self.framework.observe(
+            self.s3_requirer.on.credentials_changed, self._on_s3_credentials_changed
+        )
+        self.framework.observe(self.s3_requirer.on.credentials_gone, self._on_s3_credentials_gone)
```
Mehdi-Bendriss, line 39: I believe we also need to listen to both:
1. on_relation_joined: where we will do validation in case (in the future) the Azure integrator is also related
2. on_relation_broken: to remove the credentials while handling cases where a backup / restore is in progress

src/events/backup.py
```diff
@@ -0,0 +1,127 @@
...
+            logger.debug(f"No relation {S3_RELATION_NAME}")
+            return
+
+        # the TLS chain needs to be stored on all units in case of Juju leadership changes
+        self.charm.backup_manager.store_tls_ca_chain(s3_parameters)
+
+        if not self.charm.unit.is_leader():
+            return
+
+        if not self.charm.state.peer_relation:
```
Mehdi-Bendriss, line 58: when would this happen?

src/events/backup.py
```diff
@@ -0,0 +1,127 @@
...
+        for key, value in s3_parameters.items():
+            if isinstance(value, str):
+                s3_parameters[key] = value.strip()
+
+        # Clean up extra slash symbols to avoid issues on 3rd-party storages
+        s3_parameters["endpoint"] = s3_parameters["endpoint"].rstrip("/")
+        s3_parameters["path"] = s3_parameters["path"].strip("/")
+        s3_parameters["bucket"] = s3_parameters["bucket"].strip("/")
+
+        self.charm.backup_manager.create_bucket(s3_parameters)
```
Mehdi-Bendriss, line 78: It shouldn't be the job of the database charm to create a bucket if not exists, we should assume a bucket is already created and its ACLs well configured. 
We can test on the existence of the bucket and raise if not exists
(This is error prone, as we cannot assume we _can_ create a bucket in say s3, or what kind of ACLs the user may need in their bucket)

src/events/backup.py
```diff
@@ -0,0 +1,127 @@
...
+        self.charm.state.cluster.update({"s3-credentials": json.dumps(s3_parameters)})
+
+    def _on_s3_credentials_gone(self, event: CredentialsGoneEvent) -> None:
```
Mehdi-Bendriss, line 81: is this also invoked when a relation is broken?

src/events/backup.py
```diff
@@ -0,0 +1,127 @@
...
+        """Handle the removal of the relation with s3-integrator."""
+        self.charm.workload.remove_file(self.charm.workload.paths.tls.backup_ca)
```
Mehdi-Bendriss, line 88: I believe before doing that, we should check is a backup / restore is in progress and potentially defer in that case

src/events/backup.py
```diff
@@ -0,0 +1,127 @@
...
+        if self.charm.unit.is_leader():
+            self.charm.state.cluster.update({"s3-credentials": ""})
+
+    def _on_create_backup_action(self, event: ActionEvent) -> None:
+        """Create a backup and upload to object storage."""
+        if error := self._exists_preventing_reason():
+            event.set_results({"error": error})
+            event.fail(error)
+            return
+
```
Mehdi-Bendriss, line 99: Here, we should also check that a backup / restore operation is not already in progress and fail if that's the case

src/literals.py
```diff
@@ -19,6 +19,10 @@
 CONFIG_FILE = "/var/snap/charmed-etcd/current/etcd.conf.yml"
 TLS_ROOT_DIR = "/var/snap/charmed-etcd/common/tls"
 DATABASE_DIR = "/var/snap/charmed-etcd/common/var/lib/etcd/member"
+BACKUP_FILE_PATH = (
+    "/var/snap/charmed-etcd/common/var/lib/etcd/member/snap/charmed-etcd_snapshot.db"
```
Mehdi-Bendriss, line 23: why the `/snap/` in the last part of the path? i believe we could have a directory `snapshots` where we store this file

src/managers/backup.py
```diff
@@ -0,0 +1,166 @@
...
+                response_checksum_validation="when_required",
+            ),
+            verify=self.workload.paths.tls.backup_ca
+            if s3_parameters.get("tls-ca-chain")
+            else True,
+        )
+
+        return s3_resource.Bucket(s3_parameters["bucket"])
+
+    def create_bucket(self, s3_parameters: dict[str, str]) -> None:
```
Mehdi-Bendriss, line 58: related to a previous comment, we should rather check on its existence. 

Back to this - I see discrepancies in our charms, some attempt the bucket creation where others don't - let's confirm

src/managers/backup.py
```diff
@@ -0,0 +1,166 @@
...
+                # AWS returns these if the bucket was already created
+                "BucketAlreadyOwnedByYou" in e.args[0]
+                or "BucketAlreadyExists" in e.args[0]
+                # GCP returns this if the bucket was already created
+                or "BucketNameUnavailable" in e.args[0]
+            ):
+                logger.info(f"Using existing bucket {s3_parameters['bucket']}")
+                return
+            else:
+                # todo: do we want to raise to make the user aware of the error?
```
Mehdi-Bendriss: yes - we should definitely return it to the user in the action: `Backup failed. Check error logs for more details.`

src/managers/backup.py
```diff
@@ -0,0 +1,166 @@
...
+            client_url=self.state.unit_server.client_url,
+        )
+
+        if not etcd_client.create_database_snapshot():
+            raise EtcdBackupError("Failed to create database backup.")
+
+        bucket = self.get_bucket_resource(s3_parameters)
+
+        try:
+            bucket.upload_file(BACKUP_FILE_PATH, upload_target)
```
Mehdi-Bendriss: Could we have a tenacity retry here?

src/events/backup.py
```diff
@@ -0,0 +1,127 @@
...
+        try:
+            backup_id = self.charm.backup_manager.create_backup()
```
Mehdi-Bendriss, line 108: we should before doing that setting the charm in a maintenance status - "backup in progress"

src/events/backup.py
```diff
@@ -0,0 +1,127 @@
...
+        except EtcdBackupError as e:
+            event.set_results({"error": e})
+            event.fail(e)
```
Mehdi-Bendriss: Here I'm not so sure UX wise, I think a `event.fail("Backup failed. Check the error logs for more details")` would be better digested by the users

Mehdi-Bendriss (commented):
Thanks René! very clean! I left some comments and questions

src/events/backup.py
reneradoi: 1) Will take care of it when working on the integration of Azure backups. The s3 lib itself does not provide handling on the charm side for `relation-joined`.
2) The s3 lib emits `credentials_gone` in that case, which we already handle here.

src/events/backup.py
reneradoi: Just a safeguard in case the S3 relation is processed before the peer relation. We want to persist information in the peer relation databag, so it has to be available for it.

This case should not happen (hopefully), but I'm not aware that there is a guarantee for it.

src/events/backup.py
reneradoi: Yes, the s3 lib emits `credentials_gone` in that case.

src/events/backup.py
reneradoi: I am aware that this is handled differently in some products. In our case I chose to attempt to create the bucket, because it can easily be done and is more user-friendly.

As the user has to configure the bucket (as part of the config for s3-integrator) that they will think about it in advance and most probably set it up before. If not, our code will do.

The user still has the opportunity to adjust the ACLs before data is stored there (which will only happen when the `create-backup` action is run).

src/managers/backup.py
reneradoi: Offline conversation: We will keep this and add a section to the user docs, to make sure users don't forget to review the bucket's ACLs.

src/managers/backup.py
reneradoi: `create_bucket()` happens when the relation with s3-integrator is established, so we cannot fail a user action here. But we can fail the relation event and thereby give the user the opportunity to adjust permissions (or whatever caused the failure).

src/managers/backup.py
reneradoi: botocore has a retry already for certain errors, see https://boto3.amazonaws.com/v1/documentation/api/latest/reference/customizations/s3.html#boto3.s3.transfer.TransferConfig

I've added a general retry to cover for other cases.

src/events/backup.py
reneradoi: updated

src/events/backup.py
reneradoi: Has been added to the BackupManager's `create_backup()`.

src/events/backup.py

src/events/backup.py
reneradoi: updated

src/literals.py
reneradoi: Just a coincidence: `snap` is the name of etcd's data subdirectory where etcd stores it's database snapshot files, see [here](https://etcd.io/docs/v3.6/learning/persistent-storage-files/#long-leaving-files) for more information.

By default, snapshots created with `etcdctl snapshot save ...` are stored to this directory. I kept this default and just made it explicit.

src/events/backup.py
Mehdi-Bendriss: I don't think we log the error here

src/events/backup.py
Mehdi-Bendriss: I think this was missed

Mehdi-Bendriss (approved):
Thanks René!

PR #53, opened 2025-03-14
[DPE-6790] - TLS Status Handling

skourta:
This pull request includes several changes focused on improving the handling and computation of TLS statuses within the `etcd` charm. The changes involve updating the methods for setting and computing TLS statuses, as well as modifying the status levels and their logging levels.

Improvements to TLS status handling:

* [`src/charm.py`](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR224-R225): Added logic to compute and add TLS component statuses in the `_on_collect_status` method.
* [`src/events/etcd.py`](diffhunk://#diff-4d4b45129d72bf79a08e70009a59f4ebe0a016b4727117155249c864ae4a1b49L106): Removed the direct setting of TLS status and replaced it with deferring events and adding component statuses through the `tls_manager`. [[1]](diffhunk://#diff-4d4b45129d72bf79a08e70009a59f4ebe0a016b4727117155249c864ae4a1b49L106) [[2]](diffhunk://#diff-4d4b45129d72bf79a08e70009a59f4ebe0a016b4727117155249c864ae4a1b49L363-R362)
* [`src/events/tls.py`](diffhunk://#diff-3dfb4a104b407f7d4b2e07795a63745a09cde228a8dbe86602e7b72ea600fbe4L99-R105): Updated the initialization and event handling methods to use the `tls_manager` for adding component statuses instead of directly setting the status. [[1]](diffhunk://#diff-3dfb4a104b407f7d4b2e07795a63745a09cde228a8dbe86602e7b72ea600fbe4L99-R105) [[2]](diffhunk://#diff-3dfb4a104b407f7d4b2e07795a63745a09cde228a8dbe86602e7b72ea600fbe4L159-L162)

Status level modifications:

* [`src/literals.py`](diffhunk://#diff-959ba637043ab4ac88c6f9f9109022f78a7df1b9a2ab3a0407bff4f124c2db64L86-R90): Changed the status level for `TLS_NOT_READY` to `MaintenanceStatus` with a "DEBUG" log level and added new statuses for peer and client CA rotations.

Enhancements to `tls_manager`:

* [`src/managers/tls.py`](diffhunk://#diff-a2c036f44ae9bbab595aad84b23bc3106a342c5ab12f7b0cf3d26268b51cdc83L17-R17): Introduced a list to track component statuses, and added methods to compute component statuses and add new statuses to the list. [[1]](diffhunk://#diff-a2c036f44ae9bbab595aad84b23bc3106a342c5ab12f7b0cf3d26268b51cdc83L17-R17) [[2]](diffhunk://#diff-a2c036f44ae9bbab595aad84b23bc3106a342c5ab12f7b0cf3d26268b51cdc83R29) [[3]](diffhunk://#diff-a2c036f44ae9bbab595aad84b23bc3106a342c5ab12f7b0cf3d26268b51cdc83R205-R236)

### DIFF
--- a/src/managers/tls.py
+++ b/src/managers/tls.py
@@ -26,7 +26,6 @@ class TLSManager:
         self.state = state
         self.workload = workload
         self.substrate = substrate
-        self._component_status_list: list[Status] = []
 
     def set_tls_state(self, state: TLSState, tls_type: TLSType) -> None:
         """Set the TLS state.
@@ -205,7 +204,7 @@ class TLSManager:
 
     def compute_component_status(self) -> list[Status]:
         """Compute the component status."""
-        status_list = list(self._component_status_list)
+        status_list = []
 
         if self.state.unit_server.tls_peer_state == TLSState.TO_TLS:
             status_list.append(Status.TLS_ENABLING_PEER_TLS)
@@ -226,11 +225,3 @@ class TLSManager:
             status_list.append(Status.TLS_CLIENT_CA_ROTATING)
 
         return status_list
-
-    def add_component_status(self, status: Status) -> None:
-        """Add a component status.
-
-        Args:
-            status (Status): The status to add.
-        """
-        self._component_status_list.append(status)
--- a/src/managers/tls.py
+++ b/src/managers/tls.py
@@ -14,7 +14,7 @@ from charms.tls_certificates_interface.v4.tls_certificates import (
 
 from core.cluster import ClusterState
 from core.workload import WorkloadBase
-from literals import SUBSTRATES, TLSCARotationState, TLSState, TLSType
+from literals import SUBSTRATES, Status, TLSCARotationState, TLSState, TLSType
 
 logger = logging.getLogger(__name__)
 
@@ -26,6 +26,7 @@ class TLSManager:
         self.state = state
         self.workload = workload
         self.substrate = substrate
+        self._component_status_list: list[Status] = []
 
     def set_tls_state(self, state: TLSState, tls_type: TLSType) -> None:
         """Set the TLS state.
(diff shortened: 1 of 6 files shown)


src/managers/tls.py
```diff
@@ -26,6 +26,7 @@ def __init__(self, state: ClusterState, workload: WorkloadBase, substrate: SUBST
         self.state = state
         self.workload = workload
         self.substrate = substrate
+        self._component_status_list: list[Status] = []
```
reneradoi: Could you please explain why it is required to keep a separate list of statuses for TLS and not add them to the list of `pending_inactive_statuses` in `charm.py`?

I think we could go with one list of pending statuses for now, given that:
- we will be reworking the status handling later anyway
- the priority of the status type will be ordered by ops framework and the priority of `Status.TLS_INVALID_PRIVATE_KEY` is the highest anyway because of type `Blocked`

src/managers/tls.py
```diff
@@ -201,3 +202,35 @@ def is_cert_updated_on_all_servers(self, cert_type: TLSType) -> bool:
...
+
+        if self.state.unit_server.tls_client_state == TLSState.TO_NO_TLS:
+            status_list.append(Status.TLS_DISABLING_CLIENT_TLS)
+
+        if self.state.unit_server.tls_peer_ca_rotation_state != TLSCARotationState.NO_ROTATION:
+            status_list.append(Status.TLS_PEER_CA_ROTATING)
+
+        if self.state.unit_server.tls_client_ca_rotation_state != TLSCARotationState.NO_ROTATION:
+            status_list.append(Status.TLS_CLIENT_CA_ROTATING)
+
```
reneradoi, line 226: We could also have a check here for the `peer/client_certs_ready` property of the unit before returning.

reneradoi (commented):
Hey @skourta thank you for adding the TLS statuses. I've got one question to clarify and one little thing that could be added, but other than that, it looks good already.

src/managers/tls.py
skourta: I did it mostly to keep all logic related to status handling for TLS in the TLS manager. We can use the `pending_inactive_statuses` if you think that is a better design.

src/managers/tls.py
skourta: That is not needed because once we go to `TLSState.TLS` we know the certificate is ready. That is I only added checks for transition states.

src/managers/tls.py
reneradoi: Ok understand, thank you! I'd prefer to keep it simple and just use the `pending_inactive_statuses` list for now. Having in mind that we will at some point transition to "advanced status handling": the less places we need to adjust, the better.

Mehdi-Bendriss (approved):
Thanks!

PR #55, opened 2025-03-26
[DPE-6780] Implement s3 restore

reneradoi:
This PR adds functionality to restore backups in etcd. 

Backups can be restored to any cluster of `charmed-etcd`, no matter if the cluster is running or not or how many units are deployed.

The workflow for performing a database restore is the following:
- connect etcd to s3-storage: `juju integrate s3-integrator etcd`
- provide and configure the secret with `root` credentials
- list available backups: `juju run etcd/leader list-backups`
- restore the desired backup, for example: `juju run etcd/leader restore backup-id=2025-03-28T12:20:43Z`

The etcd operator will then perform these actions on each unit:
- download the backup from s3 storage
- stop and disable etcd 
- restore the backup with updated cluster membership configuration
- enable and start etcd
- delete the downloaded backup file

Details to the orchestration of this workflow are documented on this wiki page: https://github.com/canonical/charmed-etcd-operator/wiki/Backup-Restore:-Workflow

The main changes in the code are:

**EtcdClient:**
- re-ordered methods for better readability: `member_list()`, `is_healthy()`, `broadcast_peer_url()` and `create_database_snapshot()` moved _before_ the private method `_run_etcdctl()`
- add `_run_etcdutl()` for commands requiring to use `etcdutl` instead of `etcdctl`
- add `restore_database_snapshot()`

**BackupEvents:**
- add handler for `restore` action
- add handler for `peer_relation_changed` hook; this handler includes the main logic for orchestrating the restore workflow across all units

**BackupManager:**
- add method `download_backup_file()`
- add methods `stop_database_workload()` and `start_database_workload()` that not only stop/start etcd but also disable/enable the systemd-service
- add method `restore_backup()` to perform the actual restore operation
- add method `proceed_restore_workflow_if_possible()` that proceeds the restore-workflow after all units have completed the current step
- add method `compute_component_status()` for backup-related statuses

**Other event handlers:**
- defer `config_changed` events if restore is in progress
- defer `peer_relation_joined` events to avoid upscaling operation to the cluster while a restore is in progress
- ignore `peer_relation_changed` if restore is in progress (this event is then only handled in `BackupEvents`)
- ignore `update_status` if restore is in progress
- `on_storage_detaching`: do not remove members from the etcd cluster if restore is in progress (because the cluster member configuration will be updated after the restore anyway)
- defer `certificate_available` and `relation_broken` events for TLS if restore is in progress

**Unit State:**
- add property `restore_step` for the step of the restore workflow a unit is currently executing

**Cluster State:**
- add property `restore_instruction` for next step of the restore workflow that all units have to execute
- add property `restore_id` to store the parameter `backup-id` that is requested to be restored
- add property `is_restore_in_progress` that can be used to block operations/events while a backup is restored
- add property `can_restore_workflow_proceed` to check if all units have completed the current step of the restore workflow

**EtcdWorkload:**
- imports the lib `charms.operator_libs_linux.v1.systemd` for enabling and disabling the systemd-service for etcd


src/common/client.py
```diff
@@ -291,109 +417,70 @@ def _run_etcdctl(  # noqa: C901
...
-            snapshot_path=BACKUP_FILE_PATH,
-            endpoints=self.client_url,
-            auth_username=self.user,
-            auth_password=self.password,
-        ):
-            logger.debug(result)
-            return True
 
-        return False
+        return True if result == 0 else False
```
skourta: nit; `return result == 0`

src/events/backup.py
```diff
@@ -80,8 +91,11 @@ def _on_s3_credentials_changed(self, event: CredentialsChangedEvent) -> None:
 
     def _on_s3_credentials_gone(self, event: CredentialsGoneEvent) -> None:
         """Handle the removal of the relation with s3-integrator."""
-        if self.charm.state.cluster.is_backup_in_progress:
-            logger.debug("Backup in progress")
+        if (
+            self.charm.state.cluster.is_restore_in_progress
+            or self.charm.state.cluster.is_backup_in_progress
+        ):
+            logger.warning("Cannot s3-credentials while database backup/restore is in progress.")
```
skourta, line 101: missing `remove` in log message

src/managers/backup.py
```diff
@@ -150,6 +166,76 @@ def list_backups(self) -> list[str]:
...
+        bucket = self._get_bucket_resource(s3_parameters)
+
+        try:
+            bucket.download_file(download_source, f"{SNAP_CONFIG_PATH}/{RESTORE_FILE_NAME}")
+            logger.info(f"Backup {backup_id} downloaded to {SNAP_CONFIG_PATH}/{RESTORE_FILE_NAME}")
+        except ClientError as e:
+            logger.error(e)
+            return False
+
+        self.state.unit_server.update({"restore_step": RestoreStep.DOWNLOAD.value})
```
skourta: I would prefer to have a setter for the fields so that we avoid potential typos in setting and/or getting the fields.

src/managers/backup.py
```diff
@@ -150,6 +166,76 @@ def list_backups(self) -> list[str]:
...
+        logger.info("Restored backup successfully.")
+        self.state.unit_server.update({"restore_step": RestoreStep.RESTORE.value})
+
+    def start_database_workload(self) -> None:
+        """Enable and start the etcd database again after restoring."""
+        logger.info("Enabling and starting etcd workload.")
+        self.workload.enable_service()
+        self.workload.start()
+
+        self.state.unit_server.update({"restore_step": RestoreStep.RESTART.value})
```
skourta: Maybe rename step to `START` to fit the rest of the code?

tests/unit/test_backup.py
```diff
@@ -198,3 +212,464 @@ def test_list_backups_action():
         ctx.run(ctx.on.action("list-backups"), state_in)
 
     assert ctx.action_results == {"backups": "\n".join(expected_output)}
+
+
+def test_restore_action():
```
skourta, line 200: I noticed after writing the TLS tests that it is better to split these big tests into multiple smaller tests. Each case you're currently introducing by a comment can be an independent test.

skourta (approved):
Thank you, René. 
I left minor comments. The workflow is easy to understand and the implementation is clean.

src/events/backup.py
reneradoi: has been fixed in subsequent PR

src/common/client.py
```diff
@@ -291,109 +417,70 @@ def _run_etcdctl(  # noqa: C901
...
+                args.append(subcommand)
+            if snapshot_filename:
+                args.append(snapshot_filename)
+            if data_directory:
+                args.append("--data-dir")
+                args.append(data_directory)
+                # by default restore snapshots with revision bump
+                # see: https://etcd.io/docs/v3.6/op-guide/recovery/#restoring-with-revision-bump
+                args.append("--bump-revision")
+                args.append("1000000000")
```
Mehdi-Bendriss, line 457: Given that revisions are bumped with single digits, this seems rather excessive. Any reason to have such a high bump?

src/common/client.py
```diff
@@ -291,109 +417,70 @@ def _run_etcdctl(  # noqa: C901
...
+                args.append("--mark-compacted")
```
Mehdi-Bendriss, line 458: Nice

src/events/backup.py
```diff
@@ -136,4 +220,7 @@ def _exists_preventing_reason(self) -> str:
         if not self.charm.state.unit_server.is_started:
```
Mehdi-Bendriss: On backups (not restore) - why is the state of the daemon a preventing reason for taking a backup? given it's an `aws s3 cp ...`

src/events/backup.py
```diff
@@ -125,8 +138,79 @@ def _on_list_backups_action(self, event: ActionEvent) -> None:
...
+            event.fail(error)
+            return
+
+        if not (backup_id_to_restore := event.params.get("backup-id", "")):
+            event.fail("Must provide backup-id to restore.")
+            return
+
+        event.log(f"Initiating restore process for backup-id {backup_id_to_restore}")
+
+        if not self.charm.backup_manager.download_backup_file(backup_id_to_restore):
```
Mehdi-Bendriss, line 161: Could we explicitly check on the existence of the remote backup file and return a separate `Backup ID not found.` (validation that the provided backup id exists)

src/events/backup.py
```diff
@@ -136,4 +220,7 @@ def _exists_preventing_reason(self) -> str:
         if not self.charm.state.unit_server.is_started:
             return "Database is not started, cannot perform backup action."
```
Mehdi-Bendriss, line 229: I believe we're also missing a "Backup in progress"

src/events/backup.py
```diff
@@ -125,8 +138,79 @@ def _on_list_backups_action(self, event: ActionEvent) -> None:
...
+                        self.charm.cluster_manager.enable_authentication()
+                    except EtcdUserManagementError:
+                        logger.info("Auth already enabled")
+            case RestoreStep.COMPLETED, RestoreStep.RESTART:
+                if not self.charm.cluster_manager.is_healthy(cluster=False):
+                    # if the member is not healthy, we do not complete the restore process
+                    self.charm.set_status(Status.RESTORE_UNHEALTHY)
+                    return
+                self.charm.backup_manager.clean_up_after_restore()
+
```
Mehdi-Bendriss, line 210: How should we handle the "default" case where there is a mismatch in the relation data workflow due to a unit erroreing at some point during the restore workflow for ex?

src/events/etcd.py
```diff
@@ -298,19 +314,26 @@ def _on_secret_changed(self, event: ops.SecretChangedEvent) -> None:
...
     def _on_storage_detaching(self, event: ops.StorageDetachingEvent) -> None:
         """Handle removal of the data storage mount, e.g. when removing a unit."""
         if self.charm.app.planned_units() > 0:
-            try:
-                self.charm.cluster_manager.remove_member()
-            except (EtcdClusterManagementError, RaftLeaderNotFoundError):
-                # We want this hook to error out if we cannot remove the cluster member
-                # otherwise the cluster could become unavailable because of quorum loss
-                raise
+            if not self.charm.state.cluster.is_restore_in_progress:
```
Mehdi-Bendriss, line 329: I am not sure i understand the condition here, in the case where a restore is in progress - the current logic would **not** remove the current unit from the membership but still stop the workload later in this hook?

src/events/etcd.py
```diff
@@ -215,6 +220,9 @@ def _on_peer_relation_created(self, event: RelationCreatedEvent) -> None:
 
     def _on_peer_relation_changed(self, event: RelationChangedEvent) -> None:
         """Handle all events related to the cluster-peer relation."""
+        if self.charm.state.cluster.is_restore_in_progress:
+            return
```
Mehdi-Bendriss, line 224: should this defer?

src/literals.py
```diff
@@ -83,6 +85,13 @@ class Status(Enum):
     PASSWORD_UPDATE_FAILED = StatusLevel(BlockedStatus("failed to update password"), "ERROR")
     PEER_URL_NOT_SET = StatusLevel(MaintenanceStatus("peer-url not set"), "DEBUG")
     REMOVED = StatusLevel(BlockedStatus("unit removed from cluster"), "INFO")
+    RESTORE_FAILED = StatusLevel(BlockedStatus("failed to restore backup"), "ERROR")
+    RESTORE_IN_PROGRESS = StatusLevel(
+        MaintenanceStatus("Database restore is in progress"), "ERROR"
+    )
+    RESTORE_UNHEALTHY = StatusLevel(
+        BlockedStatus("cluster unhealthy after restoring backup - check credentials"), "ERROR"
```
Mehdi-Bendriss: is credentials mismatch the only possible failure scenario?

src/literals.py
```diff
@@ -122,3 +131,15 @@ class TLSCARotationState(Enum):
     NEW_CA_DETECTED = "new-ca-detected"
     NEW_CA_ADDED = "new-ca-added"
     CERT_UPDATED = "cert-updated"
+
+
+# enum for Backup state
+class RestoreStep(Enum):
```
Mehdi-Bendriss, line 137: nit: better use `StrEnum`

src/managers/backup.py
```diff
@@ -150,6 +166,76 @@ def list_backups(self) -> list[str]:
...
+            bucket.download_file(download_source, f"{SNAP_CONFIG_PATH}/{RESTORE_FILE_NAME}")
+            logger.info(f"Backup {backup_id} downloaded to {SNAP_CONFIG_PATH}/{RESTORE_FILE_NAME}")
+        except ClientError as e:
+            logger.error(e)
+            return False
+
+        self.state.unit_server.update({"restore_step": RestoreStep.DOWNLOAD.value})
+        return True
+
+    def stop_database_workload(self) -> None:
```
Mehdi-Bendriss: This doesn't seem like it should be in the backups manager, rather on the workload

src/managers/backup.py
```diff
@@ -150,6 +166,76 @@ def list_backups(self) -> list[str]:
...
+            raise EtcdBackupError("Failed to restore database backup.")
+
+        logger.info("Restored backup successfully.")
+        self.state.unit_server.update({"restore_step": RestoreStep.RESTORE.value})
+
+    def start_database_workload(self) -> None:
+        """Enable and start the etcd database again after restoring."""
+        logger.info("Enabling and starting etcd workload.")
+        self.workload.enable_service()
+        self.workload.start()
```
Mehdi-Bendriss: similar comment as above on the location of these 2 lines

src/managers/backup.py
```diff
@@ -173,3 +259,54 @@ def store_tls_ca_chain(self, s3_parameters: dict[str, str]) -> None:
...
+            case RestoreStep.RESTART:
+                return RestoreStep.COMPLETED
+
+    def proceed_restore_workflow_if_possible(self) -> None:
+        """Check workflow progress for all units and proceed to next step if possible."""
+        current_step = self.state.cluster.restore_instruction
+
+        # `cluster_state` must be reset before starting a restored cluster
+        if current_step == RestoreStep.RESTORE:
+            self.state.cluster.update({"cluster_state": EtcdClusterState.NEW.value})
```
Mehdi-Bendriss: should this be moved after the subsequent `if` block given the latter has a `return`

Mehdi-Bendriss (commented):
Thanks René! Excellent work, I really like the concept of the execution sequence orchestration through `RestoreStep`!

src/common/client.py
reneradoi: updated

src/managers/backup.py
reneradoi: updated

src/managers/backup.py
reneradoi: updated

src/common/client.py
reneradoi: Reasoning for this value taken from the etcd docs:
```
In order to ensure the revisions are never decreasing after a restore, you can supply the --bump-revision option.
This option takes a 64 bit integer, which denotes how many revisions to add to the current revision of the snapshot.
Since each write to etcd increases the revision by one, you may cover a week old snapshot with bumping by 1'000'000'000 assuming that etcd runs with less than 1500 writes per second.

In the context of Kubernetes controllers, it is important to also mark all the revisions, including the bump, as compacted using --mark-compacted.
This ensures that all watches are terminated and etcd does not respond to requests about revisions that happened after taking the snapshot - effectively invalidating its informer caches.
```

src/events/backup.py
reneradoi: updated

src/events/backup.py
reneradoi: updated

src/events/etcd.py
reneradoi: During the restore process, departing units cannot be removed from the cluster membership because the cluster itself is stopped and membership changes are online operation.

Instead, the departing unit is removed from the `cluster_members` in the databag. After restoring has completed, the "new" cluster membership is initialized from `cluster_members` from the databag, meaning the departed unit is no longer a cluster member.

src/events/etcd.py
reneradoi: Not necessary to defer here because there is a new `peer_relation` event coming after each step of the restore workflow.

New units that are added during the restore process are added as new cluster members after the restore has been completed.

src/literals.py
reneradoi: updated

src/literals.py
reneradoi: I'd like to avoid that because `StrEnum` requires Python 3.11 (at least for the standard lib), which we don't necessarily have. And I don't want to introduce the `StrEnum` [dependency](https://pypi.org/project/StrEnum/) because it is `NOT a drop-in replacement for the standard library implementation`.

src/managers/backup.py
reneradoi: updated

src/managers/backup.py
reneradoi: updated

src/managers/backup.py
reneradoi: has been changed into `elif` to keep the logical order of the workflow steps.

tests/unit/test_backup.py
reneradoi: Good point. Once all the already existing PR's for backup/restore have been merged into the feature branch, I will clean this up. I'd like to not mess around with it right now because that would cause a lot of merge conflicts / updates I'd have to do to all unit tests.

src/events/backup.py
reneradoi: I've changed the code so that we now allow for `offline` backups to be taken. This doesn't work with the whole `data` directory because the restore requires an empty `data-dir`, otherwise it fails. 

But I've successfully tested this (and enabled the operator code to work) with the database file [that includes all applied data of the cluster](https://etcd.io/docs/v3.5/learning/persistent-storage-files/#bbolt-btree-membersnapdb) - so we are fully able to take an `offline` snapshot and restore it on any cluster without losing any data that was applied through raft! That is a huge gain!

src/events/backup.py
reneradoi: If a step fails, the restore procedure will get stuck because not all units have completed the current step. The way forward (confirmed by testing) would be to `juju remove-unit ...` the faulty unit, which will trigger the process to continue.

src/managers/backup.py
```diff
@@ -271,20 +277,23 @@ def next_restore_step(current_step: RestoreStep) -> RestoreStep:
             case RestoreStep.STOP:
                 return RestoreStep.RESTORE
             case RestoreStep.RESTORE:
-                return RestoreStep.RESTART
-            case RestoreStep.RESTART:
+                return RestoreStep.START
+            case RestoreStep.START:
                 return RestoreStep.COMPLETED
 
+    def set_restore_step(self, restore_step: str) -> None:
```
skourta, line 284: I think this made the code significantly cleaner.

skourta (approved):
Great work @reneradoi !
I believe we can further enhance code coverage for the backup manager (currently 63%).

src/literals.py
Mehdi-Bendriss: Given that charmed-etcd is targeting the core24 base - it is guaranteed that we're running on python 3.11+

reneradoi: We would have to change the Python version of the project for this, currently we use `^3.10`. I'd rather do this separately, if that is ok for you?

Especially as this also impacts other `Enum` classes we currently have such as `TLSState`, `TLSType` and `TLSCARotationState`, which don't have anything to do with the backup/restore related PR we have here.

src/common/client.py
```diff
@@ -456,6 +456,8 @@ def _run_etcdutl(
                 args.append("--bump-revision")
                 args.append("1000000000")
                 args.append("--mark-compacted")
+                # allow restoring from an offline backup
+                args.append("--skip-hash-check")
```
Mehdi-Bendriss, line 460: This should only happen when it is an offline backup, not in all cases

Mehdi-Bendriss (commented):
Thanks René! Looks good! I have one final comment

src/literals.py
Mehdi-Bendriss: That's fine, let's do that in another PR along with juju version bump (v3.6.5) and other libs / packages

Mehdi-Bendriss (approved):
Thanks René! Excellent work! :)
(Discussed offline on the `skip-hash-check` part, metadata upload will happen in the future.)

src/common/client.py
reneradoi: Discussed offline, will be followed up.

PR #56, opened 2025-03-26
[DPE-6459] - External Client Relations

skourta:
This pull request introduces several changes to the `metadata.yaml`, `src/charm.py`, and other files to enhance the functionality and management of external clients, including the addition of new event handlers and methods for managing users and roles in etcd.

### Enhancements to External Client Management:

* [`src/charm.py`](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR16): Added `ExternalClientsManager` and `ExternalClientsEvents` to manage external client interactions and events. Updated the `_restart_ca_rotation` method to handle client CA rotation. [[1]](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR16) [[2]](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR29) [[3]](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR51-R58) [[4]](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR207-R221) [[5]](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR239-R241)

* [`src/events/external_clients.py`](diffhunk://#diff-41d3b0e68fba8bc44a83df27841cc482ba1df8a5cfe8e476397765673f46f446R1-R207): Introduced a new file to handle external clients related event handlers, including methods for updating client relations data and handling certificate events.

### Changes to `metadata.yaml`:

* Added `etcd-client` to `provides` and `client-cas` to `requires` to support new functionalities related to external clients. [[1]](diffhunk://#diff-d186204ea7abb85bbadd9d94ab29d5bcefcc43d6a0ed06015e98a71b2f099423R21-R24) [[2]](diffhunk://#diff-d186204ea7abb85bbadd9d94ab29d5bcefcc43d6a0ed06015e98a71b2f099423R34-R37)

### Enhancements to Etcd Client Utility:

* [`src/common/client.py`](diffhunk://#diff-f5c8dd5ff5ebd3473ee519d51e59ddb12e43e04365489c1c429256d3d20c3bc4R71-R103): Added methods for managing users and roles in etcd, including `get_user`, `get_role`, `add_role`, `grant_role`, `grant_permission`, `remove_role`, `remove_user`, and `get_version`. [[1]](diffhunk://#diff-f5c8dd5ff5ebd3473ee519d51e59ddb12e43e04365489c1c429256d3d20c3bc4R71-R103) [[2]](diffhunk://#diff-f5c8dd5ff5ebd3473ee519d51e59ddb12e43e04365489c1c429256d3d20c3bc4R235-R236) [[3]](diffhunk://#diff-f5c8dd5ff5ebd3473ee519d51e59ddb12e43e04365489c1c429256d3d20c3bc4R259-R260) [[4]](diffhunk://#diff-f5c8dd5ff5ebd3473ee519d51e59ddb12e43e04365489c1c429256d3d20c3bc4R302-R307) [[5]](diffhunk://#diff-f5c8dd5ff5ebd3473ee519d51e59ddb12e43e04365489c1c429256d3d20c3bc4R419-R564)

### Updates to Core Models:

* [`src/core/models.py`](diffhunk://#diff-e46c88bb687d77b94764a732b3e08a0a7bd544a64e606429c0b787c13e1c226aR219-R226): Added a `managed_users` property to the `ClusterState` class to manage user information in the cluster state.

### Event Handling Enhancements:

* [`src/events/etcd.py`](diffhunk://#diff-4d4b45129d72bf79a08e70009a59f4ebe0a016b4727117155249c864ae4a1b49R232-R233): Updated `_on_peer_relation_changed` to call `update_client_relations_data` from `ExternalClientsEvents` to ensure client relations data is updated when the peer relation changes.


src/events/external_clients.py
```diff
@@ -0,0 +1,207 @@
...
+        )
+
+        # validate leaf certificate
+        if not self.charm.external_clients_manager.is_leaf_certificate_valid(event.mtls_chain):
+            logger.error("Invalid end-entity certificate")
+            # clean the old user if exists
+            if old_common_name:
+                self._on_relation_broken(event)  # type: ignore
+                return
+            return
```
reneradoi, line 106: We should set the status to blocked here to signal the user something is wrong.

src/events/external_clients.py
```diff
@@ -0,0 +1,207 @@
...
+        self.framework.observe(
+            self.charm.on[EXTERNAL_CLIENTS_RELATION].relation_broken, self._on_relation_broken
+        )
+
+    def _on_mtls_chain_updated(self, event: MTLSChainUpdatedEvent):  # noqa: C901
+        """Handle the ca chain updated event."""
+        if not event.mtls_chain or not event.prefix:
+            logger.error("CA chain, keys prefix, or common name not provided")
+            self.charm.set_status(Status.EC_MISSING_CREDENTIALS)
+            event.defer()
```
reneradoi: Wouldn't it be possible to just return here and avoid the deferral? If one of these are missing, the fix would be to update the relation data from client side, which would then trigger another mtls_chain_updated event, or am I wrong?

src/events/external_clients.py
```diff
@@ -0,0 +1,207 @@
...
+        if self.charm.unit.is_leader():
+            if old_common_name != common_name:
+                logger.debug(f"Common name changed from {old_common_name} to {common_name}")
+
+                if old_common_name:
+                    logger.warning("Removing relation's old user")
+                    self.charm.cluster_manager.remove_managed_user(old_common_name)
+                    self.charm.external_clients_manager.remove_managed_user(event.relation.id)
+
+                if self.charm.cluster_manager.get_user(common_name) is not None:
```
reneradoi, line 122: Let's change the order of these two blocks and first check if the requested user already exists before deleting the currently used one. I know this is an edge case, but this would be more robust and allow to continue using the credentials, even if the newly requested ones cannot be used.

src/literals.py
```diff
@@ -89,6 +91,21 @@ class Status(Enum):
...
+    )
+    EC_USERNAME_EXISTS = StatusLevel(
+        MaintenanceStatus(
+            "The username provided already exists. Please provide a unique username"
+        ),
+        "ERROR",
+    )
+    EC_MISSING_CREDENTIALS = StatusLevel(
+        MaintenanceStatus("Missing certificate or prefix."), "ERROR"
+    )
```
reneradoi, line 114: Could you please maintain (roughly) alphabetical order for the status list for easier readability?

src/literals.py
```diff
@@ -89,6 +91,21 @@ class Status(Enum):
     TLS_CLIENT_CA_ROTATING = StatusLevel(MaintenanceStatus("Rotating client CA..."), "DEBUG")
     SERVICE_NOT_INSTALLED = StatusLevel(BlockedStatus("unable to install etcd snap"), "ERROR")
     SERVICE_NOT_RUNNING = StatusLevel(BlockedStatus("etcd service not running"), "ERROR")
+    EC_INVALID_CERTIFICATE = StatusLevel(
+        MaintenanceStatus(
+            "The certificate provided is a CA certificate. Please provide an end-entity certificate"
```
reneradoi: Let's add a prefix "Client relation: " to each of these status messages to make them more understandable.

src/managers/cluster.py
```diff
@@ -330,3 +330,151 @@ def compute_component_status(self) -> List[Status]:
...
+
+        Args:
+            username (str): The name of the user to add.
+            keys_prefix (str): The keys prefix to grant permission to.
+        """
+        self.add_user(username)
+        self.add_role(username)
+        self.grant_role(username, username)
+        self.grant_permission(username, keys_prefix)
+        logger.info(f"Added managed user {username}")
```
reneradoi: I believe we can make this a lot easier and avoid adding most of these methods to the `ClusterManager`.

We should only have the ones called from event handlers, which would be `get_user()`, `get_version()`, `remove_managed_user()` and `add_managed_user()`. These should then call the EtcdClient-methods directly, such as:

```
def add_managed_user():
    client = EtcdClient(...)
    client.add_user()
    client.add_role()
    client.grant_role()
    ...
```

This way we could avoid a lot of "boilerplate" code here that we don't really need.

src/events/external_clients.py
```diff
@@ -0,0 +1,207 @@
...
+        )
+
+        if relation_managed_user != common_name:
+            logger.error("New user not created yet")
+            event.defer()
+            return
+
+        if relation_managed_user and self.charm.tls_manager.is_new_ca(
+            event.mtls_chain, TLSType.CLIENT
+        ):
```
reneradoi: I'm confused: shouldn't the certificate provided here always be a leaf certificate? Then why would we check if it is a "new ca" and initiate a restart here?

src/events/external_clients.py
```diff
@@ -0,0 +1,207 @@
...
+                self.etcd_provides.set_version(relation.id, etcd_version)
+
+    def _on_certificates_available(self, event: CertificatesAvailableEvent):
+        """Handle the certificates available event."""
+        logger.debug("Certificates available event")
+        cas = self.certificate_transfer.get_all_certificates()
+        if self.certificate_transfer and self.charm.tls_manager.is_new_ca(
+            "\n".join(cas), TLSType.CLIENT
+        ):
+            self.charm.tls_events.clean_ca_event.emit(cert_type=TLSType.CLIENT)
```
reneradoi: I'm not a big fan of mixing responsibilities and emitting a "clean_ca"-event here:
- the name is misleading
- it's not easy to understand how these workflows are connected
- event handlers should not be depending on one another

Furthermore we don't need to organize this as a "CA rotation" workflow because the communication of etcd peers among each other should not be impacted. Instead, a unit can add the new client CA independently / on its own and just restart.

We could just do this:
```
if self.certificate_transfer and self.charm.tls_manager.is_new_ca():
    self.charm.tls_manager.add_trusted_ca()
    self.charm.rolling_restart()
```

src/events/external_clients.py
```diff
@@ -0,0 +1,207 @@
...
+        """Handle the relation broken event."""
+        relation_managed_user = self.charm.external_clients_manager.get_relation_managed_user(
+            event.relation.id
+        )
+
+        if self.charm.unit.is_leader() and relation_managed_user:
+            self.charm.cluster_manager.remove_managed_user(relation_managed_user)
+            self.charm.external_clients_manager.remove_managed_user(event.relation.id)
+
+        self.charm.tls_events.clean_ca_event.emit(cert_type=TLSType.CLIENT)
```
reneradoi: Same comment on `clean_ca_event` as in `_on_mtls_chain_updated()`: We don't need to do that here because the client certificate should not be a CA anyway.

src/events/external_clients.py
```diff
@@ -0,0 +1,207 @@
...
+        logger.debug("Certificates available event")
+        cas = self.certificate_transfer.get_all_certificates()
+        if self.certificate_transfer and self.charm.tls_manager.is_new_ca(
+            "\n".join(cas), TLSType.CLIENT
+        ):
+            self.charm.tls_events.clean_ca_event.emit(cert_type=TLSType.CLIENT)
+
+    def _on_certificates_removed(self, event: CertificatesRemovedEvent):
+        """Handle the certificates removed event."""
+        self.charm.tls_events.clean_ca_event.emit(cert_type=TLSType.CLIENT)
```
reneradoi: Same comment on `clean_ca_event` as in `_on_certificate_available()`: We should not mix workflows here, instead just:
- update the local CA file
- initiate a regular "rolling_restart"

src/events/external_clients.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def compute_component_status(self) -> list[Status]:
+        """Compute the component status."""
+        status_list = []
+
+        for relation in self.etcd_provides.relations:
+            mtls_chain = self.etcd_provides.fetch_relation_field(relation.id, "mtls-chain")
+            if not mtls_chain:
+                continue
```
reneradoi: We could have a status message here saying "Client certificate missing" or similar.

src/events/etcd.py
```diff
@@ -229,6 +229,8 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent) -> None:
             # reflect membership updates in the cluster state, e.g. ip change or tls switchover
             self.charm.cluster_manager.update_cluster_member_state()
 
+            self.charm.external_clients_events.update_client_relations_data()
```
reneradoi: Let's not call another event handlers methods here, to avoid dependencies between them. Instead we could have a method for updating the clients relation data in the `ExternalClientsManager`, similar to `update_cluster_member_state()` in the `ClusterManager`. We could then call `self.charm.external_clients_manager.update_client_state()` (or however it is called) from the event handlers here and in the `tls_events` as well.

src/events/tls.py
```diff
@@ -213,13 +202,19 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
...
 
         # TLS is enabled, New CA added to all servers, and cert updated -> no rolling restart needed until we clean up old CA
         if tls_state == TLSState.TLS and tls_ca_rotation_state == TLSCARotationState.NEW_CA_ADDED:
             logger.debug(f"Updating {cert_type.value} certificates with new CA")
             self.charm.tls_manager.set_ca_rotation_state(
                 cert_type, TLSCARotationState.CERT_UPDATED
             )
+            # Update the CA for external clients
+            if cert_type == TLSType.CLIENT:
+                self.charm.external_clients_events.update_client_relations_data()
```
reneradoi: Same comment on calling another event handler here: Let's add a method for updating the clients relation data in the ExternalClientsManager and then call `self.charm.external_clients_manager.update_client_state()` (or however it is called) from here.

src/events/external_clients.py
```diff
@@ -0,0 +1,207 @@
...
+            event.relation.id
+        )
+
+        if self.charm.unit.is_leader() and relation_managed_user:
+            self.charm.cluster_manager.remove_managed_user(relation_managed_user)
+            self.charm.external_clients_manager.remove_managed_user(event.relation.id)
+
+        self.charm.tls_events.clean_ca_event.emit(cert_type=TLSType.CLIENT)
+
+    def update_client_relations_data(self):
```
reneradoi: I understand that we can't call the `client_certificate.get_assigned_certificates()` from a manager. But most of this code is logic that should belong to the `ExternalClientsManager`. This would help to have a clearer responsibility on this logic, and help to decouple the event handlers from another.

src/charm.py
```diff
@@ -215,14 +221,21 @@ def _restart_ca_rotation(self, _) -> None:
...
 
+        # client CA rotation
         if self.state.unit_server.tls_client_ca_rotation_state == TLSCARotationState.CERT_UPDATED:
-            self.tls_manager.clean_cas(TLSType.CLIENT)
+            self.tls_manager.update_cas(self.tls_events.collect_client_cas(), TLSType.CLIENT)
             self.tls_manager.set_ca_rotation_state(TLSType.CLIENT, TLSCARotationState.NO_ROTATION)
 
+        # Event emitted by ECR
+        if self.state.unit_server.tls_client_ca_rotation_state == TLSCARotationState.NO_ROTATION:
+            self.tls_manager.update_cas(self.tls_events.collect_client_cas(), TLSType.CLIENT)
```
reneradoi: As commented at other places, we should not interfere the client CA workflow with our internal CA rotation workflow.

src/common/client.py
```diff
@@ -378,3 +416,149 @@ def broadcast_peer_url(self, member_id: str, peer_urls: str) -> None:
...
+            raise EtcdUserManagementError(f"Failed to remove user {username}.")
+
+    def get_version(self) -> str:
+        """Get the etcd version.
+
+        Returns:
+            str: The etcd version.
+        """
+        if result := self._run_etcdctl(
+            command="version",
```
reneradoi: I'm afraid this will give you the version of the `etcdctl` tool, not the one of the etcd server. This can instead be extracted from the `endpoint status` command and parsed from the json then.

Update: I have tested this, it indeed is just the version of the cli tool:
```
$ etcdctl version --endpoints=10.136.16.242:2379
etcdctl version: 3.4.30
API version: 3.4
$ etcdctl endpoint status --endpoints=10.136.16.242:2379 --cluster
http://10.136.16.186:2379, 3b84800fd24a4078, 3.5.18, 20 kB, true, false, 2, 83, 83, 
http://10.136.16.182:2379, 3d382d2bcfd3d387, 3.5.18, 20 kB, false, false, 2, 83, 83, 
http://10.136.16.242:2379, 46aeffd30b109895, 3.5.18, 20 kB, false, false, 2, 83, 83, 
```

reneradoi (changes requested):
Hey Smail! Thank you for providing this extensive PR! Functionality-wise it is looking pretty good. Regarding the code structure and implementation details there are some things that need to be reworked, esp. decoupling the event handlers, separating the client from the CA-rotation workflow and fixing a few details.

We can go over the comments / requested changes in a session together if you want to, please let me know if so.

src/charm.py
```diff
@@ -215,14 +221,21 @@ def _restart_ca_rotation(self, _) -> None:
...
+            self.tls_manager.update_cas([self.tls_events.collect_peer_ca()], TLSType.PEER)
             self.tls_manager.set_ca_rotation_state(TLSType.PEER, TLSCARotationState.NO_ROTATION)
 
+        # client CA rotation
         if self.state.unit_server.tls_client_ca_rotation_state == TLSCARotationState.CERT_UPDATED:
-            self.tls_manager.clean_cas(TLSType.CLIENT)
+            self.tls_manager.update_cas(self.tls_events.collect_client_cas(), TLSType.CLIENT)
             self.tls_manager.set_ca_rotation_state(TLSType.CLIENT, TLSCARotationState.NO_ROTATION)
 
+        # Event emitted by ECR
```
Mehdi-Bendriss:
```suggestion
        # Event emitted by the external client relations
```

src/events/external_clients.py
```diff
@@ -0,0 +1,207 @@
...
+        self.framework.observe(
+            self.certificate_transfer.on.certificates_removed, self._on_certificates_removed
+        )
+
+        self.framework.observe(
+            self.etcd_provides.on.mtls_chain_updated, self._on_mtls_chain_updated
+        )
+        self.framework.observe(
+            self.charm.on[EXTERNAL_CLIENTS_RELATION].relation_broken, self._on_relation_broken
+        )
```
Mehdi-Bendriss, line 59: We also need a handler on `relation_departed` and check if the departing unit is part of the ETCD app, and update the endpoints accordingly (remove the departed unit from the list)

src/events/external_clients.py
```diff
@@ -0,0 +1,207 @@
...
+        """Handle the ca chain updated event."""
+        if not event.mtls_chain or not event.prefix:
+            logger.error("CA chain, keys prefix, or common name not provided")
+            self.charm.set_status(Status.EC_MISSING_CREDENTIALS)
+            event.defer()
+            return
+
+        if not self.charm.state.unit_server.tls_client_state == TLSState.TLS:
+            logger.error("TLS is not enabled")
+            self.charm.set_status(Status.TLS_NOT_READY)
```
Mehdi-Bendriss: I believe we need to have more accurate / fine grained checks here, as the message will differ if tls is present but not ready VS tls not present at all

src/events/external_clients.py
Mehdi-Bendriss: Definitely, it should however be a `Maintenance` status (from the spec)

src/events/external_clients.py
```diff
@@ -0,0 +1,207 @@
...
+                    event.relation.id, common_name
+                )
+                self.etcd_provides.set_credentials(event.relation.id, common_name, "")
+                self.charm.external_clients_events.update_client_relations_data()
+
+        relation_managed_user = self.charm.external_clients_manager.get_relation_managed_user(
+            event.relation.id
+        )
+
+        if relation_managed_user != common_name:
```
Mehdi-Bendriss, line 141: is this to simply take into account non-juju-leader units?

src/events/external_clients.py
Mehdi-Bendriss: missing return type hint in this and subsequent methods and in the managers.cluster module

src/managers/external_clients.py
```diff
@@ -0,0 +1,105 @@
...
+from core.cluster import ClusterState
+from core.workload import WorkloadBase
+from literals import SUBSTRATES
+
+logger = logging.getLogger(__name__)
+
+WORKING_DIR = Path(__file__).absolute().parent
+
+
+class ExternalClientsManager:
```
Mehdi-Bendriss, line 23: missing return type hints in some methods of this class

src/managers/external_clients.py
```diff
@@ -0,0 +1,105 @@
...
+
+        Returns:
+            (bool): True if the certificate is not a CA.
+        """
+        # split the certificates by the end of the certificate marker and keep the marker in the cert
+        raw_cas = mtls_chain.split("-----END CERTIFICATE-----")
+        # add the marker back to the certificate
+        leaf_cert = raw_cas[0].strip() + "\n-----END CERTIFICATE-----"
+        logger.debug(f"Leaf certificate is a CA? {Certificate.from_string(leaf_cert).is_ca}")
+        return not Certificate.from_string(leaf_cert).is_ca
```
Mehdi-Bendriss: This is unfortunately not enough, I did some research, and though invalid - nothing stops a certificate from having `is_ca == False` and `key_cert_sign or ku.crl_sign == True`.
We should also check as follows:
```
try:
    key_usage = cert.extensions.get_extension_for_class(x509.KeyUsage).value
    return not (key_usage.key_cert_sign or key_usage.crl_sign)
except x509.ExtensionNotFound:
    return False
```

tests/integration/client_relations/requirer-charm/requirements.txt
```diff
@@ -0,0 +1 @@
+ops ~= 2.17
```
Mehdi-Bendriss:
```suggestion
ops ~= 2.20
```

tests/integration/client_relations/requirer-charm/charmcraft.yaml
```diff
@@ -0,0 +1,73 @@
...
+      # Ubuntu 22.04 archive. Install rustc and cargo using rustup instead of the Ubuntu archive
+      rustup set profile minimal
+      rustup default 1.83.0  # renovate: charmcraft-rust-latest
+
+      craftctl default
+      # Include requirements.txt in *.charm artifact for easier debugging
+      # cp requirements.txt "$CRAFT_PART_INSTALL/requirements.txt"
+
+actions:
+  update:
```
Mehdi-Bendriss:
```suggestion
  update-common-name:
```

tests/integration/client_relations/requirer-charm/src/charm.py
```diff
@@ -0,0 +1,348 @@
...
+logger = logging.getLogger(__name__)
+
+WORK_DIR = "./tmp"
+
+
+class RefreshTLSCertificatesEvent(ops.EventBase):
+    """Event for refreshing peer TLS certificates."""
+
+
+class RequirerCharmCharm(ops.CharmBase):
```
Mehdi-Bendriss, line 37: Missing return type hints in multiple methods / properties

tests/integration/client_relations/requirer-charm/src/charm.py
```diff
@@ -0,0 +1,348 @@
...
+        return bool(self.config.get("send-ca-cert", False))
+
+    def _on_start(self, event: ops.StartEvent):
+        """Handle start event."""
+        self.unit.status = ops.ActiveStatus()
+        # download etcdctl binary
+        urlretrieve(
+            "https://github.com/etcd-io/etcd/releases/download/v3.5.18/etcd-v3.5.18-linux-amd64.tar.gz",
+            "etcd-v3.5.18-linux-amd64.tar.gz",
+        )
```
Mehdi-Bendriss: Why not package etcdctl as part of the `prime-packages` ? This won't work in an airgapped deployment

tests/integration/client_relations/requirer-charm/src/charm.py
```diff
@@ -0,0 +1,348 @@
...
+        # extract etcdctl binary
+        with tarfile.open("etcd-v3.5.18-linux-amd64.tar.gz", "r:gz") as tar:
+            tar.extractall()
+        Path("etcd-v3.5.18-linux-amd64/etcdctl").rename("etcdctl")
+
+    def _on_update_action(self, event: ops.ActionEvent):
+        """Handle update common name action."""
+        # client relation
+        relation = self.model.get_relation("etcd-client")
+        if not relation:
```
Mehdi-Bendriss: walrus operator

tests/integration/client_relations/requirer-charm/src/charm.py
```diff
@@ -0,0 +1,348 @@
...
+        )
+
+        # TLSCertificatesRequiresV4 events
+        framework.observe(
+            self.certificates.on.certificate_available, self._on_certificate_available
+        )
+
+        # Charm events
+        framework.observe(self.on.start, self._on_start)
+        framework.observe(self.on.update_action, self._on_update_action)
```
Mehdi-Bendriss: update_common_name_action

tests/integration/client_relations/requirer-charm/src/charm.py
```diff
@@ -0,0 +1,348 @@
...
+            logger.error("No certificates available")
+            return
+
+        cert = certs[0]
+        Path(WORK_DIR).mkdir(exist_ok=True)
+        Path(f"{WORK_DIR}/client.pem").write_text(cert.certificate.raw)
+        Path(f"{WORK_DIR}/client.key").write_text(private_key.raw)
+
+        relation = self.model.get_relation("etcd-client")
+        if relation:
```
Mehdi-Bendriss: walrus operators in similar occurrences

tests/integration/client_relations/requirer-charm/src/charm.py
```diff
@@ -0,0 +1,348 @@
...
+        Path(f"{WORK_DIR}/ca.pem").write_text(event.tls_ca)
+
+    def _on_endpoints_changed(self, event: DatabaseEndpointsChangedEvent):
+        """Handle etcd client relation data changed event."""
+        logger.info("Endpoints changed: %s", event.endpoints)
+        if not event.endpoints:
+            logger.error("No endpoints available")
+            return
+        Path(WORK_DIR).mkdir(exist_ok=True)
+        Path(f"{WORK_DIR}/endpoints.txt").write_text(event.endpoints)
```
Mehdi-Bendriss: what is this file concretely for? we should ensure the endpoints are reachable as well and can write / read data into / from

Mehdi-Bendriss (changes requested):
Thanks Smail! Great work! 
I believe we can further improve the test application charm with connectivity checks and juju statuses

src/events/etcd.py
skourta: I started by trying to keep the update in the manager. However, the manager does not have access to all the state needed to compute the necessary knowledge. We need to get the server certificate (located inside the tls events handler) and also all certificates send by external relations (`etcd_provides` inside the events of the external clients). The only way to do it without a helper function in one of the event handlers is to fragment the logic and copy and paste it in both locations. I decided against that.

src/events/external_clients.py
skourta: We do not need to because we check and update all the client relations on peer relation changed.

src/events/external_clients.py
skourta: That is true.

src/events/external_clients.py
skourta: That is picked up by the `compute_component_status`

src/events/external_clients.py
skourta: I think the user of a relation should be deleted as soon as its common name is removed from the relation. That is why I set up the deletion before everything else.

src/events/external_clients.py
skourta: What do you mean?

src/events/external_clients.py
skourta: Yes, that is the case. I believe this remained from the previous implementation and should be safely switched to always restarting.

src/events/external_clients.py
skourta: The status message would be displayed in the `_on_mtls_changed` the chain is missing and that will be displayed.

tests/integration/client_relations/requirer-charm/src/charm.py
skourta: This is used as a synchronisation channel. We save the endpoints in the file and read them when we are making requests.

src/events/etcd.py
reneradoi: Copied from offline conversation:

We should adhere to the charm-structure-pattern. That means: relation data should be in and accessed through the charm context, also for non-peer relation data.

For the external clients, this can be achieved by:
- adding a class `EtcdExternalClient` to the `models.py` (similar to the `EtcdServer` class), that inherits from `RelationState` and returns all the relevant data from the client relation databag/secrets
- adding a `client_relation_interface` to the initialization of the `ClusterState`
- adding a property `client_relations` (similar to the peer_relation) to the `ClusterState`, that returns a set of all etcd_external_client relations from the model
- adding a property `external_clients` (similar to servers) to the `ClusterState` that returns a set of `EtcdExternalClients` with all their data from the relation

For the TLS CA's it can be done similarly to the S3-credentials (see the [S3 backup PR](https://github.com/canonical/charmed-etcd-operator/pull/52)).

tests/integration/client_relations/requirer-charm/src/charm.py
skourta: Handled by installing the charmed-etcd snap instead.

tests/integration/client_relations/requirer-charm/src/charm.py
skourta: Removed all files.

src/events/tls.py
```diff
@@ -323,3 +321,41 @@ def read_and_validate_private_key(
             return None
 
         return private_key
+
+    def collect_client_cas(self) -> list[str]:
```
reneradoi, line 326: This is something that should be handled by the TLS Manager too. I'd suggest to move it to `update_cas()` method of the TLS Manager and retrieve the cas from the cluster state.

That should be even easier now that we have both external clients and tls certificates in the state already, and `certificate_transfer` could be added there, too, in the same way you did it for the `tls_client_certificates`.

src/managers/external_clients.py
```diff
@@ -0,0 +1,161 @@
...
+
+        return not (key_usage.key_cert_sign or key_usage.crl_sign or basic_constraints.ca)
+
+    def update_client_relations_data(self, etcd_version: str) -> None:
+        """Update the ECR data."""
+        if not self.state.etcd_provides.relations:
+            return
+
+        uris = {server.client_url for server in self.state.servers}
+        endpoints = {f"{server.ip}:{CLIENT_PORT}" for server in self.state.servers}
```
reneradoi: To be exact we should take the cluster members from the cluster state to cover situations where units that are already part of the peer relation have not been added to the cluster as cluster members yet.

src/managers/external_clients.py
```diff
@@ -0,0 +1,161 @@
...
+            return False
+        # check if the certificate can sign other certificates
+        try:
+            key_usage = certificate.extensions.get_extension_for_class(x509.KeyUsage).value
+        except x509.ExtensionNotFound:
+            return not basic_constraints.ca
+
+        return not (key_usage.key_cert_sign or key_usage.crl_sign or basic_constraints.ca)
+
+    def update_client_relations_data(self, etcd_version: str) -> None:
```
reneradoi, line 122: Great that this is now handled by the manager!

src/managers/cluster.py
```diff
@@ -341,3 +341,97 @@ def compute_component_status(self) -> List[Status]:
...
+            client_url=self.state.unit_server.client_url,
+        )
+        return client.list_users()
+
+    def clean_users(self) -> None:
+        """Clean up users that errored on deletion."""
+        etcd_users = set(self.list_users())
+        active_users = set(self.state.cluster.managed_users.values())
+
+        for inactive_user in etcd_users - active_users:
```
reneradoi, line 427: Does this try to remove the `root` user then, because it is not a managed user?

src/managers/cluster.py
```diff
@@ -341,3 +341,97 @@ def compute_component_status(self) -> List[Status]:
...
+        client = EtcdClient(
+            username=self.admin_user,
+            password=self.admin_password,
+            client_url=self.state.unit_server.client_url,
+        )
+        client.add_user(username)
+        client.add_role(username)
+        client.grant_role(username, username)
+        client.grant_permission(username, keys_prefix)
+        logger.info(f"Added managed user {username}")
```
reneradoi, line 406: That is nice!

src/events/tls.py
```diff
@@ -213,13 +202,22 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
 
         # write certificates to disk
         self.charm.tls_manager.write_certificate(cert, private_key)  # type: ignore
+        # if there are client relations add their CAs to the trusted client CAs
+        if cert_type == TLSType.CLIENT and tls_state == TLSState.TO_TLS:
```
reneradoi, line 206: Why only `to-tls` and not also `tls`?

tests/integration/helpers.py
```diff
@@ -334,3 +336,77 @@ def get_storage_id(ops_test: OpsTest, unit_name: str, storage_name: str) -> str:
...
+
+
+def get_user(
+    endpoints: str,
+    username: str,
+    user: str | None = None,
+    password: str | None = None,
+    tls_enabled: bool = False,
+) -> dict[str, Any] | None:
+    """Write data to etcd using `etcdctl`."""
```
reneradoi: This comment needs adjustment.

tests/integration/helpers.py
```diff
@@ -334,3 +336,77 @@ def get_storage_id(ops_test: OpsTest, unit_name: str, storage_name: str) -> str:
...
+
+
+def get_role(
+    endpoints: str,
+    rolename: str,
+    user: str | None = None,
+    password: str | None = None,
+    tls_enabled: bool = False,
+) -> list[dict[str, str]] | None:
+    """Write data to etcd using `etcdctl`."""
```
reneradoi: Adjust comment please.

tests/integration/helpers.py
```diff
@@ -334,3 +336,77 @@ def get_storage_id(ops_test: OpsTest, unit_name: str, storage_name: str) -> str:
...
+                "key": base64.b64decode(perm["key"]).decode("utf-8"),
+                "range_end": base64.b64decode(perm["range_end"]).decode("utf-8"),
+            }
+            for perm in result
+        ]
+    except json.JSONDecodeError:
+        return None
+
+
+def separate_certificates(ca_chain: str) -> list[str]:
```
reneradoi: I didn't find usage for it, could be removed if not required.

reneradoi (commented):
Thank you Smail! The PR is now in a good shape, I only have some minor comments and one thing that should be moved to a manager instead of event handler.

Great Work btw.!

src/events/tls.py
skourta: This is what I mentioned yesterday about what I was going to do in a future PR as this is TLS related.

reneradoi: Ok that is fine then, sorry I missed it. Let's put in a `todo` comment here and create an issue to not forget about it.

src/managers/cluster.py
skourta: That is true. I missed that.

src/events/tls.py
skourta: In case TLS is already enabled, this is just a certificate update, so we do not need to update the client truststore.

This is to handle the case where client relations are added before TLS is enabled.

Mehdi-Bendriss (approved):
Thank you Smail! Great work!

reneradoi (approved):
Massive contribution, great work!

PR #58, opened 2025-04-01
[DPE-6782] add S3 integration tests

reneradoi:
This PR adds integration tests for creating and restoring a backup with S3-compatible storage.

The storage is deployed as a microceph instance with rados-gateway enabled. Please note: The setup of microceph was added as a test fixture because the solution provided by data-platform-workflows is [deprecated](https://github.com/canonical/data-platform-workflows/blob/main/python/pytest_plugins/microceph/README.md).

The PR is based on the branch for `s3-restore` and is supposed to be merged into the feature branch `backup-and-restore-implementation` **after** the `s3-restore` PR was merged into it.


tests/integration/backup/conftest.py
```diff
@@ -0,0 +1,105 @@
...
+    return {
+        "endpoint": f"http://{host_ip}",
+        "bucket": microceph.bucket,
+        "path": "etcd",
+        "region": "",
+    }
+
+
+@pytest.fixture(scope="session")
+def storage_credentials(microceph: ConnectionInformation) -> dict[str, str]:
```
skourta, line 152: This is called `cloud_credentials` in the other charms. For example [OS](https://github.com/canonical/opensearch-operator/blob/7c7d7b713003535e76b28caac8032bf9f35a900f/tests/integration/ha/test_backups.py#L155), and [MySQL](https://github.com/canonical/mysql-operator/blob/fce4a1fc7de6a7a654d13a7efd88966c3c862937/tests/integration/test_backup_ceph.py#L110).

tests/integration/helpers.py
```diff
@@ -68,7 +68,11 @@ def get_key(
             --cert client.pem \
             --key client.key"
 
-    return subprocess.getoutput(etcd_command).split("\n")[1]
+    try:
```
skourta: Can you please explain the reason for this change?
I prefer the helper to raise an error personally, to know exactly what happened.

skourta (approved):
Thank you @reneradoi.
I left one minor comment for consistency.
One way to enhance these tests is to add a couple of tests for misconfiguration just like in [opensearch](https://github.com/canonical/opensearch-operator/blob/7c7d7b713003535e76b28caac8032bf9f35a900f/tests/integration/ha/test_backups.py#L843)

tests/integration/backup/conftest.py
reneradoi: That is true. I've named it `storage_credentials`/`storage_config` because `cloud` is misleading in this case.

tests/integration/helpers.py
reneradoi: I agree, and I cannot remember anymore why I changed it - maybe as part of debugging. Anyway, have reverted it now.

tests/integration/backup/test_s3_backup_restore.py
```diff
@@ -0,0 +1,206 @@
...
+    await ops_test.model.deploy(CHARM_PATH, num_units=NUM_UNITS)
+    await wait_until(ops_test, apps=[APP_NAME], wait_for_exact_units=NUM_UNITS, idle_period=60)
+
+    logger.info("Configure admin credentials in etcd")
+    secret_name = "new_test_secret"
+
+    secret_id = await ops_test.model.add_secret(
+        name=secret_name, data_args=[f"{INTERNAL_USER}={PASSWORD}"]
+    )
+    await ops_test.model.grant_secret(secret_name=secret_name, application=APP_NAME)
```
Mehdi-Bendriss: May be worth externalizing this in a helper to avoid repeating it in new tests

tests/integration/backup/test_s3_backup_restore.py
```diff
@@ -0,0 +1,206 @@
...
+    logger.info("Remove existing etcd cluster and deploy a new one.")
+    await ops_test.model.remove_application(APP_NAME, block_until_done=True)
+    await ops_test.model.deploy(CHARM_PATH, num_units=NUM_UNITS)
+    await wait_until(ops_test, apps=[APP_NAME], wait_for_exact_units=NUM_UNITS, idle_period=60)
+
+    logger.info("Configure admin credentials in etcd")
+    secret_name = "new_test_secret"
+
+    secret_id = await ops_test.model.add_secret(
+        name=secret_name, data_args=[f"{INTERNAL_USER}={PASSWORD}"]
```
Mehdi-Bendriss: For the sake of testing, can we have a different root password?

Mehdi-Bendriss (commented):
Thanks René! Great work! I have one comment on the root password value but looks good 

ps: could we also add a test with the radosgw with TLS enabled?

tests/integration/backup/test_s3_backup_restore.py
reneradoi: This would result in a failed restore, because the root password is stored as part of the backup. Meaning the password needs to be known to be able to restore a backup correctly.

I will add an integration test with a different password as part of the [PR for verifying backup restore](https://github.com/canonical/charmed-etcd-operator/pull/85), in order to have a failing restore.

tests/integration/backup/test_s3_backup_restore.py
reneradoi: updated

Mehdi-Bendriss (approved):
Thank you René!

PR #59, opened 2025-04-02
[DPE-6850] improve UX with status messages

reneradoi:
This PR improves the UX with better status messages for the following cases:
- on installation
- on cluster initialization by the leader
- when reusing storage, if authentication fails because the credentials were not provided by the user

It also adds a retry to the `ClusterManager`'s `broadcast_peer_url()` method to account for failures when the etcd service has just been started and may take a few seconds to be available for updating the member configuration.

### DIFF
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -95,6 +95,9 @@ class EtcdEvents(Object):
 
     def _on_start(self, event: ops.StartEvent) -> None:  # noqa: C901
         """Handle start event."""
+        # check if data exists before doing any operation
+        storage_reuse = self.charm.workload.exists(DATABASE_DIR)
+
         tls_transition_states = [TLSState.TO_TLS, TLSState.TO_NO_TLS]
         if (
             self.charm.state.unit_server.tls_client_state in tls_transition_states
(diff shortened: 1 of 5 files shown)


skourta (dismissed):
Thank you @reneradoi LGTM.

src/events/etcd.py
```diff
@@ -141,13 +149,16 @@ def _on_start(self, event: ops.StartEvent) -> None:  # noqa: C901
...
+                    event.defer()
+                    return
 
             if not self.charm.state.unit_server.is_started:
                 # database files should not be deleted on running units
-                if self.charm.workload.exists(DATABASE_DIR):
+                if storage_reuse:
                     logger.warning(f"Existing database file detected in {DATABASE_DIR}.")
                     # storage cannot be reused on non-leader members
                     try:
```
Mehdi-Bendriss, line 164: Before starting a member we should also have "Waiting for ETCD to start..."
(few lines below the highlighted one)

Mehdi-Bendriss (commented):
Thanks René! Only 1 comment on the start service

src/events/etcd.py
reneradoi: updated

Mehdi-Bendriss (approved):
Thank you!

skourta (approved):
Re-approval

PR #63, opened 2025-04-23
Refactor internal links

a-velasco:
## Issue
The spellcheck workflow returns the following warning:

```
WARNING: more than one target found for 'myst' cross-reference manage-passwords: could be :std:ref:`How to manage passwords` or :std:doc:`How to manage passwords`
```

## Solution

- [x] Remove the `(manage-passwords)=` anchor to get rid of the warning.
- [ ] Remove all other H1 anchors and verify that existing references to the files still work


a-velasco:
These same link fixes will come from a different branch (`3.5/docs`).

PR #64, opened 2025-04-23
[DPE-6781] Add azure support

reneradoi:
This PR adds support for Azure object storage as backend for storing backup files. It is based on top of the [restore-PR](https://github.com/canonical/charmed-etcd-operator/pull/55) and merges into the feature-branch for backup/restore implementation.

The implementation is similar to the existing S3-implementation. The main points of interest are:
- credentials and configuration is shared with etcd via the `azure-storage-integrator`, implementing the `azure` relation interface
- the `azure_storage` lib is temporarily added to this repo via copy&paste, as it has not been published in Charmhub yet
- the implementation for azure uses the functionalities provided by `azure.storage.blob` Python package
- if both the `s3` and `azure` relation is configured, the relation-events will be handled without raising errors, but the charm will be in blocked status
- any backup-related user action will fail if both relations are configured


src/events/backup.py
```diff
@@ -107,6 +125,52 @@ def _on_s3_credentials_gone(self, event: CredentialsGoneEvent) -> None:
         if self.charm.unit.is_leader():
             self.charm.state.cluster.update({"s3-credentials": ""})
 
+    def _on_azure_credentials_changed(self, event: StorageConnectionInfoChangedEvent):
```
Mehdi-Bendriss: nit:
```suggestion
    def _on_azure_credentials_changed(self, event: StorageConnectionInfoChangedEvent)-> None:
```

src/events/backup.py
```diff
@@ -107,6 +125,52 @@ def _on_s3_credentials_gone(self, event: CredentialsGoneEvent) -> None:
...
+
+        # make sure we have all required parameters for writing to the storage
+        required_parameters = ["container", "storage-account", "path", "secret-key"]
+        if missing_parameters := [p for p in required_parameters if p not in azure_parameters]:
+            raise KeyError(f"Parameters missing from Azure integrator: {missing_parameters}")
+
+        # Strip whitespaces from all parameters
+        for key, value in azure_parameters.items():
+            if isinstance(value, str):
+                azure_parameters[key] = value.strip()
```
Mehdi-Bendriss, line 150: could you create an issue on the object storage integrator if that's the case?

src/events/backup.py
```diff
@@ -221,8 +285,22 @@ def _exists_preventing_reason(self) -> str:
         if not self.charm.unit.is_leader():
             return "Action must be performed on the leader unit."
 
-        if not self.charm.state.cluster.s3_credentials:
-            return "No credentials for object storage available."
+        if self.charm.model.get_relation(AZURE_RELATION_NAME) and self.charm.model.get_relation(
+            S3_RELATION_NAME
+        ):
+            return "Azure and S3 storage configured - please remove one."
```
Mehdi-Bendriss:
```suggestion
            return "Azure and S3 storages configured - please remove one."
```

src/literals.py
```diff
@@ -82,6 +83,9 @@ class Status(Enum):
     )
     HEALTH_CHECK_FAILED = StatusLevel(MaintenanceStatus("health check failed"), "DEBUG")
     NO_PEER_RELATION = StatusLevel(MaintenanceStatus("no peer relation available"), "DEBUG")
+    OBJECT_STORAGE_CONFLICT = StatusLevel(
+        BlockedStatus("Azure and S3 storage configured - please remove one"), "ERROR"
```
Mehdi-Bendriss:
```suggestion
        BlockedStatus("Azure and S3 storages configured - please remove one"), "ERROR"
```

src/events/backup.py
```diff
@@ -221,8 +285,22 @@ def _exists_preventing_reason(self) -> str:
         if not self.charm.unit.is_leader():
             return "Action must be performed on the leader unit."
```
Mehdi-Bendriss, line 287: I think we are missing a check on `no S3 and no Azure` relations

Mehdi-Bendriss (commented):
Thank you rené! I left a few nits and a comment

src/events/backup.py
skourta: Given the credentials are only added on relation changed of the storage integrators is that necessary since we have the checks on the credentials just after?

src/events/backup.py
```diff
@@ -107,6 +125,52 @@ def _on_s3_credentials_gone(self, event: CredentialsGoneEvent) -> None:
...
+                azure_parameters[key] = value.strip()
+
+        # Clean up extra slash symbols to avoid issues on 3rd-party storages
+        azure_parameters["path"] = azure_parameters["path"].strip("/")
+        azure_parameters["container"] = azure_parameters["container"].strip("/")
+
+        self.charm.backup_manager.create_container(azure_parameters)
+        self.charm.state.cluster.update({"azure-credentials": json.dumps(azure_parameters)})
+
+    def _on_azure_credentials_gone(self, event: StorageConnectionInfoGoneEvent):
```
skourta: Return type for function missing

tests/unit/test_backup.py
```diff
@@ -67,7 +68,53 @@ def test_s3_relation():
         assert isinstance(e.value.__cause__, KeyError)
 
 
-def test_create_backup_action():
+def test_azure_relation():
```
skourta, line 71: The unit test coverage is still lacking at only 52%. If this to be enhanced in a future PR can we create a jira ticket for it?

skourta (commented):
Thank you @reneradoi .
My only blocker is the test coverage.

src/events/backup.py

src/events/backup.py
reneradoi: No it's true, running actions should be blocked if no S3- or Azure-relation is configured. This was done so far by the `if no S3-relation` check, but as this has evolved, the `no relation` case got lost. Thanks for the find @Mehdi-Bendriss , I've added this and unit test coverage.

src/literals.py
reneradoi: updated

src/events/backup.py

src/events/backup.py
reneradoi: Will check again and do so.

src/events/backup.py

tests/unit/test_backup.py
reneradoi: Unit test coverage should mostly focus on stuff that makes sense to test independently, imho. And for the `BackupManager`, increasing unit test coverage would require mocking complete S3 / Azure interfaces. I don't think we should do this.

The functionality itself is covered by the integration test, which is what we should use to test interfaces. So unless you have a very strong desire to mock all of this out, I believe it's okay to leave it as it is.

tests/unit/test_backup.py
```diff
@@ -354,6 +354,32 @@ def test_support_only_one_object_storage():
         assert e.message == "Azure and S3 storage configured - please remove one."
 
 
+def test_ensure_at_least_one_object_storage():
+    ctx = testing.Context(EtcdOperatorCharm)
+    peer_relation = testing.PeerRelation(id=1, endpoint=PEER_RELATION)
+
+    # ensure backup cannot be created if both s3 and azure are related
```
skourta: Outdated comment

skourta (approved):
Thanks @reneradoi.

skourta (approved):

Mehdi-Bendriss (approved):
Thanks René!

PR #72, opened 2025-05-06
feat(DPE-7021): Terraform modules

skourta:
This pull request introduces Terraform support for deploying and managing the `charmed-etcd` operator and related resources. It includes workflow updates, a new Terraform module, and associated configuration files, enabling automated linting, testing, and deployment of the charm.

### New Terraform Module for `charmed-etcd`
* Added a new Terraform module (`terraform/charm`) for deploying the `charmed-etcd` operator. This includes resources for deploying the charm (`juju_application`) and setting up integrations (`juju_integration`).
* Defined input variables (`variables.tf`) for configuring the charm deployment, including options for application name, channel, model, units, constraints, and more.
* Added output variables (`outputs.tf`) to expose application names and integration endpoints for downstream use.

### Terraform Version and Provider Requirements
* Specified Terraform version (`>= 1.6`) and required provider (`juju >= 0.14.0`) in `versions.tf`.

### GitHub Actions Workflow Enhancements
* Added two new jobs to `.github/workflows/ci.yaml`: `terraform-lint` for linting Terraform code and `terraform-test` for running Terraform-based tests. These jobs include steps for setting up dependencies (e.g., `terraform`, `juju`, `lxd`) and deploying the charm using Terraform.
* Minor formatting update to the cron schedule syntax in the workflow file.


.github/workflows/ci.yaml
```diff
@@ -93,3 +93,74 @@ jobs:
...
+          df --human-readable
+      - name: (self hosted) Disk usage
+        timeout-minutes: 1
+        run: df --human-readable
+      - name: Install terraform snap
+        run: |
+          sudo snap install terraform --channel=latest/stable --classic
+      - name: juju set up
+        run: |
+          sudo snap install juju --channel=3.6/beta --classic
```
reneradoi: Let's install juju from `/stable` as we do in all other integration tests.

terraform/charm/variables.tf
```diff
@@ -0,0 +1,96 @@
...
+  description = "Map of endpoint bindings"
+  type        = map(string)
+  default     = {}
+}
+
+variable "expose" {
+  description = "Expose the application for external access."
+  type        = bool
+  default     = false
+}
```
reneradoi, line 78: I'm not sure why we need this, could you please explain?

terraform/charm/README.md
```diff
@@ -0,0 +1,54 @@
...
+The module offers the following configurable inputs:
+
+| Name          | Type        | Description                                               | Required |
+| ------------- | ----------- | --------------------------------------------------------- | -------- |
+| `app_name`    | string      | Application name                                          | False    |
+| `channel`     | string      | Channel that the charm is deployed from                   | False    |
+| `base`        | string      | The series to be used for this charm                      | False    |
+| `config`      | map(string) | Map of the charm configuration options                    | False    |
+| `model`       | string      | Name of the model that the charm is deployed on           | **True** |
+| `resources`   | map(string) | Map of the charm resources                                | False    |
```
reneradoi: The `resources` input is not used in the terraform code and can be removed here (as we also don't need it).

.github/workflows/ci.yaml
```diff
@@ -93,3 +93,74 @@ jobs:
...
+      - name: Juju setup
+        run: |
+          sg 'lxd' -c "juju bootstrap 'localhost' --config model-logs-size=10G"
+          juju model-defaults logging-config='<root>=INFO; unit=DEBUG'
+      - name: Terraform deploy
+        working-directory: ./terraform/charm
+        run: |
+          juju add-model test
+          terraform init
+          terraform apply -var "model=test" -auto-approve
```
reneradoi: Can we have an extra step here to display the output of `juju status`? I'd like to see if the deployment is actually coming up.

reneradoi (commented):
Thank you @skourta , nice work! I only left some minor suggestions and a question, other than that, looks good already!

terraform/charm/variables.tf
skourta: I believe the terraform modules are used in the public clouds extensively and if someone wants to expose etcd through the public network they will need this. Please correct me if I am wrong.

reneradoi: Ah so it's a general thing, not specific to our operator - that makes sense. I've found this: https://github.com/juju/terraform-provider-juju/blob/a78c38e2e3399dba83ecd0ab29a017e8deb6923f/docs/resources/application.md?plain=1#L109-L115

Let's keep it then.

skourta: Here's the [docs](https://registry.terraform.io/providers/juju/juju/latest/docs/resources/application#expose-1).
Let's leave this comment open for @Mehdi-Bendriss's opinion.

Mehdi-Bendriss: Yes, I think it's a sensible input variable to have.

.github/workflows/ci.yaml
```diff
@@ -93,3 +93,79 @@ jobs:
...
+      - name: Model Status
+        run: |
+          # wait for etcd to be ready
+          sleep 180
+          juju status --relations
```
Mehdi-Bendriss, line 172: consider using `juju wait-for`

.gitignore
```diff
@@ -14,3 +14,8 @@ client.pem
 client.key
 client_ca.pem
 coverage.xml
+.env
+.terraform
+.terraform.lock.hcl
+*.tfstate
+*.tfstate.*
```
Mehdi-Bendriss:
```suggestion
*.tfstate.*

```

terraform/charm/main.tf
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  endpoint_bindings = [
+    for k, v in var.endpoint_bindings : {
+      endpoint = k, space = v
+    }
+  ]
+}
+
+resource "juju_application" "self-signed-certificates" {
```
Mehdi-Bendriss, line 35: unlike opensearch, this is not a mandatory requirement for the etcd charm module
consider introducing a conditional such as: https://github.com/canonical/opensearch-dashboards-operator/blob/2/edge/terraform/main.tf#L35-L36

As well as in all below subsequent references to the tls operator

terraform/charm/main.tf
```diff
@@ -0,0 +1,91 @@
...
+  units       = 1
+  constraints = var.self-signed-certificates.constraints
+}
+
+
+#--------------------------------------------------------
+# 2. INTEGRATIONS
+#--------------------------------------------------------
+
+resource "juju_integration" "tls-etcd-peer" {
```
Mehdi-Bendriss, line 58: related to the above comment on having a conditional

terraform/charm/main.tf
```diff
@@ -0,0 +1,91 @@
...
+    endpoint = "peer-certificates"
+  }
+
+  depends_on = [
+    juju_application.self-signed-certificates,
+    juju_application.etcd,
+  ]
+}
+
+resource "juju_integration" "tls-etcd-client" {
```
Mehdi-Bendriss, line 80: related to the above comment on having a conditional

terraform/product/main.tf
```diff
@@ -0,0 +1,107 @@
...
+  charm {
+    name     = "data-integrator"
+    channel  = var.data-integrator.channel
+    revision = var.data-integrator.revision
+    base     = var.data-integrator.base
+  }
+  model  = var.etcd.model
+  config = var.data-integrator.config
+
+  constraints = var.data-integrator.constraints
```
Mehdi-Bendriss, line 39: missing placement directives (for the machines)

Mehdi-Bendriss (commented):
Thanks Smail! Great work! I left a few comments

terraform/product/main.tf
skourta: [Placement is deprecated](https://registry.terraform.io/providers/juju/juju/latest/docs/resources/application#placement-3). I added the `machines` variable but decided to go with units in the `main.tf`. You cannot specify both.

skourta: I have made both `units` and `machines` nullable and set the default value to `null`. If the type is `number`, you cannot specify `null` as an input value, so I set `units`'s default value to `null`.
Since the attributes are mutually exclusive by default in the provider, there is no need to add extra validation on the variables.

Mehdi-Bendriss: We would however need to have a validation that checks when:
- both are null 
- both are set

.github/workflows/ci.yaml
```diff
@@ -93,3 +93,79 @@ jobs:
...
+        run: |
+          terraform init
+          terraform fmt
+          terraform validate
+
+  terraform-test:
+    name: Terraform - Run terraform tests
+    needs:
+      - terraform-lint
+    runs-on: ubuntu-22.04
```
Mehdi-Bendriss: We can use ubuntu-latest here

terraform/charm/main.tf
```diff
@@ -0,0 +1,100 @@
...
+  charm {
+    name     = "charmed-etcd"
+    channel  = var.channel
+    revision = var.revision
+    base     = var.base
+  }
+  config             = var.config
+  model              = var.model
+  name               = var.app_name
+  units              = var.units != null ? var.units : null
```
Mehdi-Bendriss: This seems redundant. This ternary operator is effectively the same as simply assigning `var.units`

terraform/charm/main.tf
```diff
@@ -0,0 +1,100 @@
...
+  constraints        = var.constraints
+  storage_directives = var.storage
+  machines           = var.machines != null ? var.machines : null
```
Mehdi-Bendriss: This seems redundant. This ternary operator is effectively the same as simply assigning `var.units`

terraform/product/variables.tf
```diff
@@ -0,0 +1,98 @@
...
+    condition     = length(var.backups-integrator.machines) <= 1
+    error_message = "Machine count should be at most 1"
+  }
+}
+
+variable "tls" {
+  description = "Enable TLS for the application"
+  type        = bool
+  default     = false
+}
```
Mehdi-Bendriss: this variable is not needed in the context of the product module. Since external relations are tls based and we deploy the self-signed-certificates operator.

terraform/product/main.tf
```diff
@@ -0,0 +1,113 @@
...
+
+#--------------------------------------------------------
+# 2. INTEGRATIONS
+#--------------------------------------------------------
+
+
+# Integrator apps and grafana-agent
+resource "juju_integration" "data_integrator-etcd-integration" {
+  # This integration is only created if TLS is enabled
+  for_each = var.tls ? { "deployed" = true } : {}
```
Mehdi-Bendriss: We can drop the TLS check on the product module - a product module is what we consider to be deployed as a fully contained deployment able to take on client relations / applications.

Mehdi-Bendriss (commented):
Thanks Smail, I left a few comments.

terraform/product/main.tf
skourta: - Both null => machines is set to null and units default to 1. An etcd cluster with one unit is deployed.
- Both are set => you get an error from the provider telling you `Attribute "units" cannot be specified when "machines" is specified`

Mehdi-Bendriss (dismissed):

terraform/product/main.tf
```diff
@@ -0,0 +1,108 @@
...
+    name = var.etcd.app_name
+  }
+
+  depends_on = [
+    module.etcd,
+    juju_application.grafana-agent,
+  ]
+}
+
+# TODO add backup integrator <-> etcd integration once backups are merged in etcd
```
reneradoi: Is this still a todo? Or don't we want this by default?

reneradoi (dismissed):
Thank you Smail, looks good! I only have one question, other than that, consider it approved from my side.

terraform/product/main.tf
skourta: It is not. I have added and removed the comment.
Thank you @reneradoi .

reneradoi (approved):

Mehdi-Bendriss (approved):

PR #75, opened 2025-05-08
fix(DPE-7106): fix double mtls changed event being emitted

### LINKED ISSUE #73: Wrong "username" already exists, for new client relation users
For new users of the mTLS external client relations - an erroneous status gets displayed
```
The username provided already exists
```

When in fact no other user with the same name exists.

Image

skourta:
fixes #73 

## Context
The `mtls_cert_updated` is fired twice because relation changed is being fired twice:
- Once when requirer updates the data on `relation_created`.
- [2nd time right after `relation_joined`](https://arc.net/l/quote/hcevgqfy)

## Fix
Add conditions to detect when user is already being created and return

### DIFF
--- a/lib/charms/tls_certificates_interface/v4/tls_certificates.py
+++ b/lib/charms/tls_certificates_interface/v4/tls_certificates.py
@@ -9,7 +9,7 @@ interface.
 Pre-requisites:
   - Juju >= 3.0
   - cryptography >= 43.0.0
-  - pydantic >= 2.0
+  - pydantic >= 1.0
 
 Learn more on how-to use the TLS Certificates interface library by reading the documentation:
 - https://charmhub.io/tls-certificates-interface/
@@ -27,6 +27,7 @@ from datetime import datetime, timedelta, timezone
 from enum import Enum
 from typing import FrozenSet, List, MutableMapping, Optional, Tuple, Union
 
+import pydantic
 from cryptography import x509
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import rsa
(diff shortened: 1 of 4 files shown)


reneradoi (dismissed):
Looks good!

src/events/external_clients.py
```diff
@@ -119,20 +119,30 @@ def _on_mtls_cert_updated(self, event: MTLSCertUpdatedEvent) -> None:  # noqa: C
...
+                    )
                 )
+                if self.charm.cluster_manager.get_user(common_name) is not None:
+                    if common_name != relation_managed_user:
+                        logger.error("User already exists")
+                        self.charm.set_status(Status.EC_USERNAME_EXISTS)
+                        return
+                    if common_name == relation_managed_user:
+                        logger.debug("User is already being added for this relation")
+                        return
```
Mehdi-Bendriss:
```suggestion
                    if common_name == relation_managed_user:
                        logger.debug("User is already being added for this relation")
                    else:
                        logger.error("User already exists")
                        self.charm.set_status(Status.EC_USERNAME_EXISTS)
                    return
```

Mehdi-Bendriss (commented):
Thanks Smail, I have a small code style related comment.

lib/charms/tls_certificates_interface/v4/tls_certificates.py
```diff
@@ -9,7 +9,7 @@
 Pre-requisites:
   - Juju >= 3.0
   - cryptography >= 43.0.0
-  - pydantic >= 2.0
+  - pydantic >= 1.0
```
reneradoi, line 12: :open_mouth: no more pydantic 2 requirement! See https://github.com/canonical/tls-certificates-interface/pull/347 - we can remove our dependency pin for pydantic then!

skourta: Yes, I saw. Do you want to do it on this PR?

reneradoi: Now that you've updated the lib - yes :)

Mehdi-Bendriss (approved):

reneradoi (approved):
Thanks Smail!

PR #79, opened 2025-05-20
[DPE-7331] Update to 3.6

reneradoi:
This PR updates the etcd operator to upstream `v3.6`. It includes the following changes:
- deploy new snap revision
- adjust configuration to new/removed config parameters
- adjusts `get_endpoint_status()` method of `EtcdClient` class to always use credentials for authentication
- adjust several integration test helpers that rely on the `etcdctl endpoint status` or `etcdctl member list` commands to use authentication
- adjust CI workflows to handle `3.6/edge` as default branch and to publish to Charmhub channel `3.6/edge`
- adjust docs to deploy from channel `3.6/edge`

**After this PR was merged, the branch `3.6/edge` should become the default branch in this repository.**

It also includes several other, minor updates:
- charm libs
- `requests` dependency due to [security alert](https://github.com/canonical/charmed-etcd-operator/security/dependabot/7)
- juju agent version to `3.6.7` in CI


.github/workflows/ci.yaml
```diff
@@ -80,7 +80,7 @@ jobs:
       matrix:
         juju:
           # This runs on all runs
-          - agent: 3.6.0 # renovate: juju-agent-pin-minor
+          - agent: 3.6.5 # renovate: juju-agent-pin-minor
```
Mehdi-Bendriss:
```suggestion
          - agent: 3.6.7 # renovate: juju-agent-pin-minor
```

skourta (commented):
Great work @reneradoi 

You missed a change to `3.6/edge` in `.readthedocs.yaml`.

src/managers/config/etcd.conf.yml
```diff
@@ -7,16 +7,11 @@ snapshot-count: 10000
 heartbeat-interval: 100
 election-timeout: 1000
 quota-backend-bytes: 0
+max-learners: 1
```
Mehdi-Bendriss, line 10: Don't we add all nodes as learners first as they come up?

Mehdi-Bendriss (commented):
Thanks René, I have a question on the max-learners

reneradoi:
> You missed a change to `3.6/edge` in `.readthedocs.yaml`.

Thank you! Updated now. By the way I didn't update the terraform files on purpose because the `3.6/edge` channel doesn't have 
anything published yet and thus they would fail.

src/managers/config/etcd.conf.yml
reneradoi: Yes, but one by one to not overload the raft leader with data transfer operation. In 3.5 it was not configurable and always set to 1, now it can be configured and I want to make `max-learners=1` explicit.

.github/workflows/ci.yaml
reneradoi: updated

Mehdi-Bendriss (approved):

skourta (approved):

PR #84, opened 2025-05-28
[DPE-6996] Separate backup volume

reneradoi:
Backup/restore in etcd has to happen with locally stored files because streaming directly to/from object storage is not possible. This PR introduces a separate volume for holding the backup file in order to avoid outages because of full file system.

**UI/UX changes:**
- Users can specify a `archive` volume when deploying etcd: 
  - `juju deploy ./charmed-etcd_ubuntu@24.04-amd64.charm etcd -n 3 --storage data=etcd-storage,1G,1 --storage archive=etcd-storage,1G,1`
- Users can re-attach previously used `archive` volumes on deployment / scale-up:
  - `juju deploy ./charmed-etcd_ubuntu@24.04-amd64.charm etcd --attach-storage=data/1 --attach-storage=archive/0`
  - `juju add-unit etcd --attach-storage=data/1 --attach-storage=archive/0`

**Implementation:**
- define a new `archive` storage in `metadata.yaml`
- adjust the `BackupManager` class to use the `archive` storage when temporarily storing the backup-file before uploading to / when downloading from object storage
- adjust integration test for storage reuse


Mehdi-Bendriss (approved):

skourta (approved):
Thank you @reneradoi

PR #85, opened 2025-05-30
[DPE-6997] Verify backup before restore

reneradoi:
As part of the restore procedure, the data directories of all units will be purged. In a situation where the restore fails (e.g. wrong admin password configured), the cluster would ultimately be lost and all the data could not be recovered anymore. To avoid this, we will run a verification of the restore on the leader before actually executing the restore on all units.

The restore-verification will:
- purge the unit's data directory
- restore the backup file only on the leader unit
- immediately start etcd
- perform a health check on the leader unit

If the health check fails, the restore workflow will NOT proceed, so that the data on the remaining units will not be purged. Instead, the leader unit will reset its data directory and the restore-step will be skipped for all units. The cluster will be recovered as before the restore was initiated.

If the health check succeeds, the restore procedure can continue on all units.

**UI/UX:**
- add new status `Blocked: Restore verification failed - etcd cluster still running, restore cancelled, check debug-log`
- if a new `restore` action is executed by a user, the status will be reset

**Implementation:**
- add `RestoreStep.VERIFY`
- add new method `_verify_restore()` to the `BackupEvents` class that bundles the execution flow for the `verify` step
- adjust workflow execution for the `restore` step

**Testing:**
- adjust unit test coverage in `test_restore_workflow_synchronization()`
- add new integration test file `test_restore_verification_failed.py`

### DIFF
--- a/src/events/backup.py
+++ b/src/events/backup.py
@@ -261,12 +261,13 @@ class BackupEvents(Object):
                     self.charm.set_status(Status.RESTORE_FAILED)
             case RestoreStep.STOP, RestoreStep.DOWNLOAD:
                 self.charm.backup_manager.stop_database()
+                self.charm.backup_manager.set_restore_step(RestoreStep.STOP.value)
             case RestoreStep.VERIFY, RestoreStep.STOP:
                 self._verify_restore()
             case RestoreStep.RESTORE, RestoreStep.VERIFY:
                 if self.charm.state.cluster.restore_verification_failed:
                     logger.error(
-                        "Restore procedure not successful - start previous etcd database again"
+                        "Restore procedure not successful - the etcd cluster will restart with the previous state."
                     )
                     self.charm.backup_manager.set_restore_step(RestoreStep.RESTORE.value)
                 else:
@@ -283,6 +284,7 @@ class BackupEvents(Object):
                         self.charm.cluster_manager.enable_authentication()
                     except EtcdUserManagementError:
                         logger.info("Auth already enabled")
+                self.charm.backup_manager.set_restore_step(RestoreStep.START.value)
             case RestoreStep.COMPLETED, RestoreStep.START:
                 if not self.charm.cluster_manager.is_healthy(cluster=False):
                     # if the member is not healthy, we do not complete the restore process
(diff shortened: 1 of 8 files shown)


src/events/backup.py
```diff
@@ -251,11 +261,19 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent) -> None:  # noq
...
-                try:
-                    self.charm.backup_manager.restore_backup()
-                except EtcdBackupError:
-                    self.charm.set_status(Status.RESTORE_FAILED)
+            case RestoreStep.VERIFY, RestoreStep.STOP:
+                self._verify_restore()
+            case RestoreStep.RESTORE, RestoreStep.VERIFY:
+                if self.charm.state.cluster.restore_verification_failed:
+                    logger.error(
+                        "Restore procedure not successful - start previous etcd database again"
```
Mehdi-Bendriss: perhaps:
```suggestion
                        "Restore procedure not successful - the etcd cluster will restart with the previous state."
```

src/events/backup.py
```diff
@@ -276,6 +294,62 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent) -> None:  # noq
         if self.charm.unit.is_leader():
             self.charm.backup_manager.proceed_restore_workflow_if_possible()
 
+    def _verify_restore(self) -> None:
+        """Safeguard the restore-operation in order not to cause any data loss.
```
Mehdi-Bendriss, line 300: very nice docstrings

src/events/backup.py
```diff
@@ -276,6 +294,62 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent) -> None:  # noq
...
+                self.charm.backup_manager.set_restore_step(RestoreStep.VERIFY.value)
+                return
+            # if the verification was successful, stop etcd again and continue the workflow
+            logger.info(
+                f"Restore verification successful: Backup {self.charm.state.cluster.restore_id} can be restored."
+            )
+            self.charm.backup_manager.stop_database()
+            self.charm.backup_manager.set_restore_step(RestoreStep.VERIFY.value)
+        else:
+            self.charm.backup_manager.set_restore_step(RestoreStep.VERIFY.value)
```
Mehdi-Bendriss: nit: we could invert these conditions and return, in order to reduce the nestedness of the previous condition logic

src/managers/backup.py
```diff
@@ -273,6 +273,9 @@ def stop_database(self) -> None:
         # disable the service to avoid restart while the backup is restored
         self.workload.disable_database()
 
+        if self.state.cluster.restore_instruction == RestoreStep.VERIFY:
+            return
```
Mehdi-Bendriss: can you explain this condition?

Mehdi-Bendriss (commented):
Thanks René! looks really clean! I left a question

src/managers/backup.py
reneradoi: In case of the `verify` step, we don't want to immediately proceed to the next step of the workflow, as there are other actions to be done before that.

Thinking about it, I believe it's better to remove this from here and make it part of the event flow itself.

src/events/backup.py
reneradoi: updated.

src/events/backup.py

Mehdi-Bendriss (approved):
Thanks René!

tests/integration/backup/test_restore_verification_failed.py
```diff
@@ -0,0 +1,171 @@
...
+    # bucket should be created when integrating both
+    assert s3_bucket.meta.client.head_bucket(Bucket=s3_bucket.name)
+
+
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy"])
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_create_backup(ops_test: OpsTest) -> None:
+    """Create a backup and upload to s3-storage."""
+    global backup_id
```
skourta, line 78: We should avoid doing this in my opinion. It makes tests harder to debug and irreproducible if you are executing them one at a time.

skourta (approved):
Thank you @reneradoi.
I have one concern regarding the integration test.

tests/integration/backup/test_restore_verification_failed.py
reneradoi: I will address this in a separate PR, as it also impacts the other integration test.

PR #88, opened 2025-06-11
[DPE-7525] adjust integration test for mtls certificate in external client relations

reneradoi:
This PR fixes the integration test for external client relations, after a fix for the order of certificates in the TLS CA chain has been implemented on the self-signed-certificates operator (see https://github.com/canonical/self-signed-certificates-operator/pull/349).

**Changes:**
- no longer invert order of certificates in requirer-charm for external client relations
- run integration tests for external client relations against `1/edge` version of self-signed-certificates operator (same as all other TLS-related tests already)


Mehdi-Bendriss (approved):
Thanks René! Good catch

skourta (approved):
Thank you @reneradoi

PR #91, opened 2025-06-13
[DPE-7389] Remove inconsistent cluster members

reneradoi:
If a user removes a unit from an etcd cluster with `juju remove-unit ... --force --no-wait`, it can happen that the removing unit was not correctly removed from etcd cluster membership. In order to avoid quorum issues, the cluster membership configuration should be checked for inconsistency on update_status and units that are gone should be removed from cluster membership.

This PR adds a new method `remove_inconsistent_members_if_required()` that is executed by the Juju leader on `update_status`. It also adds unit and integration test coverage.

### DIFF
--- a/tests/integration/ha/test_disaster_recovery.py
+++ b/tests/integration/ha/test_disaster_recovery.py
@@ -51,6 +51,7 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
 async def test_membership_reconfiguration_after_unit_loss(ops_test: OpsTest) -> None:
     """Make sure a forcefully removed unit is removed as cluster member."""
     unit_to_remove = ops_test.model.applications[APP_NAME].units[-1]
+    removed_member_name = unit_to_remove.name.replace("/", "")
     logger.info(f"Forcefully removing unit {unit_to_remove.name}")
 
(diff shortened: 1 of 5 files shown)


skourta (dismissed):
Thank you @reneradoi .

I ran several manual tests. All good.

tests/integration/ha/test_disaster_recovery.py
```diff
@@ -0,0 +1,78 @@
...
+    # wait for the next `update_status` for the cluster membership to be updated
+    async with ops_test.fast_forward("5s"):
+        assert_continuous_writes_increasing(
+            endpoints=endpoints, user=INTERNAL_USER, password=password
+        )
+
+    cluster_members = get_cluster_members(endpoints)
+    assert len(cluster_members) == NUM_UNITS - 1, (
+        f"Expected {NUM_UNITS - 1} cluster members, got {len(cluster_members)}."
+    )
```
Mehdi-Bendriss: could we test against the unit names/ids instead of the length?

Mehdi-Bendriss (commented):
Thanks René! I have a minor comment

Mehdi-Bendriss (approved):

skourta (approved):

tests/integration/ha/test_disaster_recovery.py
reneradoi: updated

PR #93, opened 2025-06-16
small home page updates

a-velasco:
Added an external link to etcd's Charmhub page, since the theme doesn't currently support it in the "More resources" dropdown.

image

Updated Diatáxis table to include the Explanation section.


reneradoi (approved):
Thank you Andreia!

skourta (approved):

PR #95, opened 2025-06-19
[DPE-6998] Recover from majority failure

reneradoi:
If an etcd cluster experiences majority failure, it can not recover on its own. This PR adds a solution for these kind of error cases. It allows users to rebuild the existing cluster, but with new cluster membership configuration, while keeping the existing data.

It includes the following changes:

**UI/UX:**
- add a new `BlockedStatus` if cluster majority failure was detected: `Cluster failure - majority of cluster members lost. Run action rebuild-cluster to recover.`
- add a new action `rebuild-cluster` that allows users to recover a cluster from majority failure be re-initializing the cluster membership configuration. If the cluster is working fine, the action will fail to avoid shutting down the cluster unintentionally. This can be overridden with `force=True`.
- add a new `BlockedStatus` if cluster rebuild was executed by a user: `Rebuilding with new cluster configuration...`

**Implementation:**
- `EtcdClient`: new method `get_metric()` that allows querying the etcd metric server for a given metric (used by the `ClusterManager` to detect cluster failure)
- add check to detected failed clusters on `update_status`
- new handler `_on_rebuild_cluster_action` for `EtcdEvents`
- new properties for unit and cluster state to check the cluster rebuild progress
- block all potentially concurring event handlers if cluster rebuild is in progress, such as `config_changed`, `certificate_available` or others
- new method `_rebuild_cluster()` in `EtcdEvents` to orchestrate the workflow for rebuilding a cluster based on `peer_relation_changed` events

The workflow is also documented in this wiki page: https://github.com/canonical/charmed-etcd-operator/wiki/Rebuild-cluster:-Workflow

**Testing:**
- unit test coverage for the `rebuild-cluster` action
- unit test coverage for the workflow for rebuilding the cluster which is based on `peer_relation_changed` events
- integration test coverage for rebuilding a failed cluster by running the action in `test_disaster_recovery.py`

Unrelated change:
- updating `protobuf` dependency to `6.31.1` because of [CVE-2025-4565](https://github.com/canonical/charmed-etcd-operator/security/dependabot/8)

### DIFF
--- a/src/common/client.py
+++ b/src/common/client.py
@@ -660,7 +660,12 @@ class EtcdClient:
         Returns:
             str: value of the metric
         """
-        response = requests.get(self.client_url)
+        try:
+            response = requests.get(self.client_url)
+        except requests.exceptions.RequestException as e:
+            logger.error(e)
+            raise RuntimeError(e)
+
         # the metrics server always returns text format, no json available
         for line in response.text.split("\n"):
             if metric_name in line and "#" not in line:
--- a/src/common/client.py
+++ b/src/common/client.py
@@ -10,6 +10,7 @@ import os
 import subprocess
 from typing import Tuple
 
+import requests
 import tenacity
 
 from common.exceptions import (
@@ -646,3 +647,23 @@ class EtcdClient:
         ):
             return json.loads(result)["users"]
         return []
+
+    def get_metric(self, metric_name: str) -> str | None:
+        """Query the metric server.
+
+        Performs an http request to the internal metric server of etcd, searches the result for
+        the given metric and returns the first result found.
+
+        Args:
+            metric_name: name of the metric
+
+        Returns:
+            str: value of the metric
+        """
+        response = requests.get(self.client_url)
+        # the metrics server always returns text format, no json available
+        for line in response.text.split("\n"):
+            if metric_name in line and "#" not in line:
+                return line.split(" ")[1]
+
+        return None
--- a/src/common/client.py
+++ b/src/common/client.py
@@ -668,7 +668,7 @@ class EtcdClient:
 
         # the metrics server always returns text format, no json available
         for line in response.text.split("\n"):
-            if metric_name in line and "#" not in line:
+            if metric_name in line and not line.startswith("#"):
                 return line.split(" ")[1]
 
         return None
--- a/src/common/client.py
+++ b/src/common/client.py
@@ -664,7 +664,7 @@ class EtcdClient:
             response = requests.get(self.client_url)
         except requests.exceptions.RequestException as e:
             logger.error(e)
-            raise RuntimeError(e)
+            raise
 
         # the metrics server always returns text format, no json available
         for line in response.text.split("\n"):
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -435,6 +435,16 @@ class EtcdEvents(Object):
             event.fail(error)
             return
 
+        try:
+            # safeguard to avoid users wrecking fine clusters
+            if not self.charm.cluster_manager.is_cluster_failed and not event.params.get(
+                "force", ""
+            ):
+                event.fail("Cluster has not failed. Use `force` to rebuild anyway.")
+                return
+        except RuntimeError:
+            logger.warning("Could not determine if cluster failed - continue with cluster rebuild")
+
         logger.info("Cluster rebuild initiated.")
         logger.info("Stopping and disabling etcd workload.")
         # disable the service to avoid restart while workflow is in progress
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -41,6 +41,7 @@ from literals import (
     SNAP_USER,
     TLS_CLIENT_PRIVATE_KEY_CONFIG,
     TLS_PEER_PRIVATE_KEY_CONFIG,
+    EtcdClusterState,
     Status,
     TLSState,
     TLSType,
@@ -412,6 +413,11 @@ class EtcdEvents(Object):
             event.fail(error)
             return
 
+        logger.info("Cluster rebuild initiated.")
+        logger.info("Stopping and disabling etcd workload.")
+        # disable the service to avoid restart while workflow is in progress
+        self.charm.workload.disable_database()
+        self.charm.state.unit_server.update({"state": ""})
         self.charm.state.cluster.update({"rebuild_cluster": "True"})
         event.set_results({"result": "cluster rebuild in progress"})
 
@@ -499,7 +505,37 @@ class EtcdEvents(Object):
         with a `BlockedStatus`. Users should then investigate (e.g. force-remove a faulty unit)
         and re-run the action.
         """
-        pass
+        cluster_members = ",".join([unit.member_endpoint for unit in self.charm.state.servers])
+
+        if self.charm.unit.is_leader():
+            if not any(unit.is_started for unit in self.charm.state.servers):
+                logger.info("All units stopped - initialise new cluster configuration.")
+                self.charm.state.cluster.update({"cluster_state": EtcdClusterState.NEW.value})
+                self.charm.state.cluster.update({"cluster_members": cluster_members})
+                self.charm.config_manager.set_config_properties()
+
+                logger.info("Enabling and starting etcd again.")
+                self.charm.workload.enable_database()
+                self.charm.state.unit_server.update({"state": "started"})
+            elif (
+                all(unit.is_started for unit in self.charm.state.servers)
+                and self.charm.cluster_manager.is_healthy()
+            ):
+                logger.info("All units started again - cluster rebuild completed.")
+                self.charm.state.cluster.update({"cluster_state": EtcdClusterState.EXISTING.value})
+                self.charm.state.cluster.update({"rebuild_cluster": ""})
+
+            return
+
+        if self.charm.state.cluster.cluster_state == EtcdClusterState.EXISTING.value:
+            logger.info("Stopping and disabling etcd workload.")
+            self.charm.workload.disable_database()
+            self.charm.state.unit_server.update({"state": ""})
+        elif self.charm.state.cluster.cluster_state == EtcdClusterState.NEW.value:
+            logger.info("Enabling and starting etcd again.")
+            self.charm.config_manager.set_config_properties()
+            self.charm.workload.enable_database()
+            self.charm.state.unit_server.update({"state": "started"})
 
     def _exists_preventing_reason(self) -> str:
         """Check if an action can be executed, if not return error message.
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -11,6 +11,7 @@ from typing import TYPE_CHECKING
 import ops
 from ops import Object
 from ops.charm import (
+    ActionEvent,
     LeaderElectedEvent,
     RelationChangedEvent,
     RelationCreatedEvent,
@@ -83,6 +84,9 @@ class EtcdEvents(Object):
         )
         self.framework.observe(
             self.charm.on[DATA_STORAGE].storage_attached, self._on_storage_attached
+        )
+        self.framework.observe(
+            self.charm.on.rebuild_cluster_action, self._on_rebuild_cluster_action
         )
 
     def _on_storage_attached(self, event: ops.StorageAttachedEvent) -> None:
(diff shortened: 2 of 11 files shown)


src/common/client.py
```diff
@@ -646,3 +647,28 @@ def list_users(self) -> list[str]:
...
+            metric_name: name of the metric
+
+        Returns:
+            str: value of the metric
+        """
+        try:
+            response = requests.get(self.client_url)
+        except requests.exceptions.RequestException as e:
+            logger.error(e)
+            raise RuntimeError(e)
```
Mehdi-Bendriss: I don't think this warrants a `RuntimeError`, we can be more specific here. From the [docs](https://docs.python.org/3/library/exceptions.html#RuntimeError):
> Raised when an error is detected that doesn’t fall in any of the other categories. The associated value is a string indicating what precisely went wrong.

src/events/etcd.py
```diff
@@ -240,6 +251,16 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent) -> None:
...
 
+        if self.charm.state.cluster.rebuild_cluster_in_progress:
+            try:
+                self._rebuild_cluster()
+                return
+            except EtcdClusterManagementError as e:
+                # if adding a member or promoting a learner fails, we want to re-run this again
+                logger.warning(e)
+                event.defer()
+                return
```
Mehdi-Bendriss: nit:
```suggestion
            except EtcdClusterManagementError as e:
                # if adding a member or promoting a learner fails, we want to re-run this again
                logger.warning(e)
                event.defer()
            return
```

src/events/etcd.py
```diff
@@ -189,8 +197,11 @@ def _on_start(self, event: ops.StartEvent) -> None:  # noqa: C901
 
     def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
         """Handle config_changed event."""
-        if self.charm.state.cluster.is_restore_in_progress:
-            logger.warning("Cannot update config while database restore is in progress.")
+        if (
+            self.charm.state.cluster.is_restore_in_progress
+            or self.charm.state.cluster.rebuild_cluster_in_progress
+        ):
+            logger.warning("Cannot update config while another operation is in progress.")
```
Mehdi-Bendriss: I would perhaps be more specific here, something along the lines of:
```suggestion
            logger.warning("Cannot update config while a restore or cluster-rebuild operation is in progress.")
```

src/events/etcd.py
```diff
@@ -290,9 +311,11 @@ def _on_peer_relation_departed(self, event: RelationDepartedEvent) -> None:
...
     def _on_peer_relation_joined(self, event: RelationJoinedEvent) -> None:
         """Handle event received by all units when a new unit joins the cluster relation."""
-        if self.charm.state.cluster.is_restore_in_progress:
-            logger.warning("Cannot add cluster member while database restore is in progress.")
-            event.defer()
+        if (
+            self.charm.state.cluster.is_restore_in_progress
+            or self.charm.state.cluster.rebuild_cluster_in_progress
+        ):
+            logger.warning("Cannot add cluster member while another operation is in progress.")
```
Mehdi-Bendriss: similar comment on message (and in subsequent occurrences)

src/events/etcd.py
```diff
@@ -323,24 +346,49 @@ def _on_leader_elected(self, event: LeaderElectedEvent) -> None:
 
             self.charm.state.cluster.update({f"{INTERNAL_USER}-password": password})
 
+        try:
+            if self.charm.cluster_manager.is_cluster_failed:
```
Mehdi-Bendriss, line 355: Can you explain the choice of hook for this check?

src/events/etcd.py
```diff
@@ -368,19 +416,52 @@ def _on_secret_changed(self, event: ops.SecretChangedEvent) -> None:
...
             event.defer()
             return
 
         if admin_secret_id := self.charm.config.get(INTERNAL_USER_PASSWORD_CONFIG):
             if admin_secret_id == event.secret.id:
                 self.update_admin_password(admin_secret_id)
 
+    def _on_rebuild_cluster_action(self, event: ActionEvent) -> None:
+        """Recover from majority failure by rebuilding the cluster membership configuration."""
+        if error := self._exists_preventing_reason():
```
Mehdi-Bendriss: We should probably rename this one and the method used in the restore flow to be more specific, perhaps: 
```suggestion
        if error := self._exists_rebuild_preventing_reason():
```

src/events/etcd.py
```diff
@@ -368,19 +416,52 @@ def _on_secret_changed(self, event: ops.SecretChangedEvent) -> None:
...
+            event.set_results({"error": error})
+            event.fail(error)
+            return
+
+        try:
+            # safeguard to avoid users wrecking fine clusters
+            if not self.charm.cluster_manager.is_cluster_failed and not event.params.get(
+                "force", ""
```
Mehdi-Bendriss:
```suggestion
                "force", False
```

src/events/etcd.py
```diff
@@ -440,3 +521,102 @@ def update_private_key(self, private_key_id: str) -> None:
...
+                self.charm.state.cluster.update({"cluster_state": ""})
+                self.charm.config_manager.set_config_properties()
+
+                logger.info("Enabling and starting etcd again.")
+                self.charm.workload.enable_service()
+                self.charm.cluster_manager.start_member()
+                self.charm.state.unit_server.update({"rebuild_completed": "True"})
+            elif (
+                all(unit.rebuild_completed for unit in self.charm.state.servers)
+                and not self.charm.state.cluster.learning_member
```
Mehdi-Bendriss, line 562: I'm not sure I understand this middle-condition in the flow of the check

src/events/etcd.py
```diff
@@ -440,3 +521,102 @@ def update_private_key(self, private_key_id: str) -> None:
...
+                        self.charm.cluster_manager.add_member(unit.unit_name)
+                        break
+
+            return
+
+        # this is the workflow for non-leader units
+        if not self.charm.state.cluster.cluster_members:
+            # the action was executed on the leader, cluster member configuration was cleared
+            # clean up in case a previous run failed
+            self.charm.state.unit_server.update({"rebuild_completed": ""})
```
Mehdi-Bendriss, line 585: isn't this part only called when there is `rebuild_in_progress`?

src/events/etcd.py
```diff
@@ -440,3 +521,102 @@ def update_private_key(self, private_key_id: str) -> None:
...
+            and not self.charm.state.unit_server.is_started
+        ):
+            # startup phase
+            logger.info("Enabling and starting etcd again.")
+            self.charm.config_manager.set_config_properties()
+            self.charm.workload.enable_service()
+            self.charm.cluster_manager.start_member()
+            self.charm.state.unit_server.update({"rebuild_completed": "True"})
+
+    def _exists_preventing_reason(self) -> str:
```
Mehdi-Bendriss: nit: `exists_` implies that this returns a bool

tests/integration/ha/test_disaster_recovery.py
```diff
@@ -75,10 +75,71 @@ async def test_membership_reconfiguration_after_unit_loss(ops_test: OpsTest) ->
...
     assert_continuous_writes_consistent(endpoints=endpoints, user=INTERNAL_USER, password=password)
+
+
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy"])
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_recover_from_majority_failure(ops_test: OpsTest) -> None:
+    """When the majority of the cluster is lost, users can run `rebuild-cluster`."""
+    await wait_until(ops_test, apps=[APP_NAME], wait_for_exact_units=NUM_UNITS - 1)
+
```
Mehdi-Bendriss, line 95: Better in a separate test:
1. Could we add a check for the action when the cluster is healthy and ensure that the action fails
2. pass the `force` param and ensure that the cluster _actually_ rebuilds itself and becomes healthy again

tests/integration/ha/test_disaster_recovery.py
```diff
@@ -75,10 +75,71 @@ async def test_membership_reconfiguration_after_unit_loss(ops_test: OpsTest) ->
...
+    rebuild_action = await leader_unit.run_action("rebuild-cluster")
+    rebuild_response = await rebuild_action.wait()
+    assert rebuild_response.results.get("return-code") == 0, "rebuild failed"
+
+    # wait for the rebuild to be performed
+    await wait_until(ops_test, apps=[APP_NAME], wait_for_exact_units=2)
+
+    endpoints = get_cluster_endpoints(ops_test, APP_NAME)
+    cluster_members = get_cluster_members(endpoints)
+    member_names = [member["name"] for member in cluster_members]
```
Mehdi-Bendriss, line 132: Could we check explicitely the content of the member_names to ensure that it contains the 3 expected units?

Mehdi-Bendriss (commented):
Thanks René, great work! I left some comments and questions

src/managers/cluster.py
```diff
@@ -68,6 +68,25 @@ def leader(self) -> str:
...
+    def is_cluster_failed(self) -> bool:
+        """Check if the cluster is experiencing majority failure.
+
+        Returns:
+            bool: True if the cluster has failed, False if not.
+        """
+        client = EtcdClient(
+            username=self.admin_user,
+            password=self.admin_password,
+            client_url=f"http://{self.state.unit_server.ip}:{METRICS_PORT}/metrics",
```
skourta, line 81: The metrics endpoint is also exposed under the `CLIENT_PORT` so we can use the default client url and move the `/metrics` route inside the `get_metric` function.

src/common/client.py
reneradoi: replaced with `RequestException`.

src/events/etcd.py
reneradoi: updated

src/events/etcd.py
reneradoi: updated

src/events/etcd.py
reneradoi: updated

src/events/etcd.py
reneradoi: It's a nice-to-have that I discovered when testing: If you force-remove the Juju leader or the unit is lost because of network cut, a new leader will be elected. We could use the opportunity to perform the check and don't have to wait for the next update status to detect cluster failure.

src/events/etcd.py
reneradoi: has been renamed

src/events/etcd.py
reneradoi: has been renamed

src/events/etcd.py
reneradoi: The Juju leader loops over all units and adds them one-by-one as cluster member. The unit then starts and sets `unit.rebuild_completed`. This has to happen here because the promotion to full-voting-member happens on the Juju leader, the unit itself is not participating in this step and cannot set `unit.rebuild_completed` there.

When all units are added and started (reading: `all(unit.rebuild_completed for unit in self.charm.state.servers)`, the last one could still be a learner and the cluster would not yet be healthy. That's why we skip the health check at this moment and wait for the last unit to be promoted to complete the whole workflow.

src/events/etcd.py
reneradoi: Yes that's true. This condition is to allow users to run `rebuild-cluster` multiple times if it failed on a previous execution.

Let's say a unit could not complete the rebuild process, for whatever reason. The flag `rebuild_completed` would already be set on any other unit that did complete the workflow. So if we want to run it again (before the whole workflow is done and all units cleaned up their flags), we need to purge the flag from the units that did already complete it, because otherwise we would not add them to the cluster again.

tests/integration/ha/test_disaster_recovery.py
reneradoi: We do, right below:
- check that all units of the deployment are in the `member_names`
- check that the removed units are not in the `member_names` anymore

Or do you mean something else?

skourta (commented):
Great work @reneradoi. Very well thought-out synchronisation.
Besides Mehdi's I left a comment about the client url for the metrics endpoint.

tests/integration/ha/test_disaster_recovery.py
reneradoi: Updated the integration test. We also have unit test coverage for this ([here](https://github.com/canonical/charmed-etcd-operator/blob/recover-from-majority-failure/tests/unit/test_charm.py#L842-L861)), that's why I didn't put this into integration test initially.

src/managers/cluster.py
reneradoi: I think we should not do that, mainly for two reasons:
- The metrics port is always `http` by our design, the client port might be `https` - we would have to consider TLS in the whole logic - unnecessary complexity here
- What if the etcd project decided to no longer expose metrics on the client port?

skourta: This is not a hard opinion for me, but I think it is a bit misleading to pass the endpoint URL inside the `client_url`.
- For TLS, we already have that in `_run_etcdctl`, we can do something similar, no?
- And for the etcd project, I think the [`listen-metrics-urls`](https://etcd.io/docs/v3.6/op-guide/monitoring/#metrics-endpoint) are optional, while the `client-url` one is not. The default metrics endpoint is exposed on the `client_url`, so any change there would have to be a breaking change.

reneradoi: offline conversation: we keep it as is because it's simple and making things explicit by using the metrics url.

skourta (approved):
Thank you @reneradoi

Mehdi-Bendriss (approved):

PR #104, opened 2025-06-27
[DPE-7642] add user documentation for recovery from majority failure

reneradoi:
With https://github.com/canonical/charmed-etcd-operator/pull/95 the charmed etcd operator was enhanced to be able to recover from majority failure. This PR adds user documentation for the feature.

### DIFF
--- a/docs/how-to/disaster-recovery.md
+++ b/docs/how-to/disaster-recovery.md
@@ -42,7 +42,7 @@ Charmed etcd provides a mechanism to recover from majority failure. It can be ex
 This will:
 - stop the cluster
 - keep existing data
-- re-initialize the cluster members 
+- re-initialise the cluster members 
 - form a new cluster with the remaining, functional units and the existing data
 
 ```{caution}
--- a/docs/how-to/disaster-recovery.md
+++ b/docs/how-to/disaster-recovery.md
@@ -1,4 +1,5 @@
 # How to perform disaster recovery
+
 An etcd cluster automatically recovers from temporary failures of cluster members. If cluster members are lost permanently,
 etcd is able to withstand up to `(N-1)/2` lost members.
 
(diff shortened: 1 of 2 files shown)


docs/how-to/disaster-recovery.md
```diff
@@ -0,0 +1,116 @@
...
+```{caution}
+Before executing the action `rebuild-cluster`, make sure to remove the failed Juju units.
+```
+All units that are permanently lost, for example because their machine crashed, need to be removed first. In our example, 
+these are the units `etcd/1` and `etcd/2`. If they can not be reached anymore by the Juju controller, they might need to
+be removed with the `--force` option.
+
+Remove them by running:
+
+`juju remove-unit etcd/1 etcd/2 --force --no-wait`
```
a-velasco: Is there a particular reason why you used inline notation instead of a code block?

a-velasco (changes requested):
Excellent guide! Just one request:

Insert a new line **after headings** and **before code blocks**.

````
# Some heading

Text begins. Here's some code:

```
juju do stuff
```
````

We'll implement an automatic check soon to help with this, but there'll be less debt to deal with later if we get into the habit now.

File: docs/how-to/disaster-recovery.md (hunk 1 above)
Comment (reneradoi) [reply docs/how-to/disaster-recovery.md]: Initially I wanted to use this code block inline indeed, but decided otherwise later, because it is a bit to long of a command. Just forgot to change the notation then, fixed now.

**Review by a-velasco — DISMISSED** [review]

File: docs/how-to/disaster-recovery.md [hunk 2]
```diff
@@ -0,0 +1,128 @@
+# How to perform disaster recovery
+
+An etcd cluster automatically recovers from temporary failures of cluster members. If cluster members are lost permanently,
+etcd is able to withstand up to `(N-1)/2` lost members.
+
```
Mehdi-Bendriss, line 5: Perhaps worth adding a link to the docs? e.g: As stated by the [official etcd docs](http://...):

Mehdi-Bendriss (dismissed):
Thanks René! Only a minor non-blocking comment

docs/how-to/disaster-recovery.md
reneradoi: good idea, updated

Mehdi-Bendriss (approved):

a-velasco (approved):

PR #110, opened 2025-07-07
[DPE-7633] - add public ip to client certs

skourta:
### Workload IP Management Enhancements:
* [`src/core/workload.py`](diffhunk://#diff-4d1ed63ca48f694e77680b1434284f1d8d896cffe1842fb6b8b8e616c192b696R173-R190): Added abstract methods `get_private_ip` and `get_public_ip` to define interfaces for retrieving IP addresses in workload classes.
* [`src/workload.py`](diffhunk://#diff-38281d02638d07eddd511b40cb2aa3cc1877e72f7ce4caacfe2a2a0c34cfd75fR135-R178): Implemented `get_private_ip` and `get_public_ip` methods using subprocess calls to retrieve IP addresses from the unit, with error handling and logging for robustness.

### TLS Certificate Management Improvements:
* [`src/managers/tls.py`](diffhunk://#diff-a2c036f44ae9bbab595aad84b23bc3106a342c5ab12f7b0cf3d26268b51cdc83R267-R289): Added the `get_sans_ip` method to dynamically generate SANs IPs based on workload's private and public IPs, with specific handling for peer and client TLS types. Includes detailed logging for debugging.
* [`src/events/tls.py`](diffhunk://#diff-3dfb4a104b407f7d4b2e07795a63745a09cde228a8dbe86602e7b72ea600fbe4L102-R102): Updated certificate request generation to use the new `get_sans_ip` method for dynamically determining SANs IPs based on TLS type. [[1]](diffhunk://#diff-3dfb4a104b407f7d4b2e07795a63745a09cde228a8dbe86602e7b72ea600fbe4L102-R102) [[2]](diffhunk://#diff-3dfb4a104b407f7d4b2e07795a63745a09cde228a8dbe86602e7b72ea600fbe4L116-R116)


src/core/workload.py
```diff
@@ -170,3 +170,21 @@ def disable_database(self) -> None:
...
+    def get_private_ip(self) -> str | None:
+        """Get the private IP of the workload.
+
+        Returns:
+            str: The private IP address.
+        """
+        pass
+
+    @abstractmethod
+    def get_public_ip(self) -> str | None:
```
Mehdi-Bendriss: Shouldn't these rather be in the `managers/cluster.py`?

src/workload.py
```diff
@@ -132,3 +132,47 @@ def disable_database(self) -> None:
...
+
+        if output.returncode != 0:
+            return None
+
+        return output.stdout.strip()
+
+    @override
+    def get_private_ip(self) -> str | None:
+        """Get the Private IP address of the current unit."""
+        cmd = "unit-get private-address"
```
Mehdi-Bendriss: I don't think we need to use unit-get for this one, do we?
how about 
```
import socket

s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
try:
    # doesn't have to be reachable
    s.connect(("10.255.255.255", 1))
    ip = s.getsockname()[0]
except Exception:
    ip = "127.0.0.1"
finally:
    s.close()
```

And can you check the difference between this way and `charm.model.get_binding(peer_relation_name).network.bind_address` and whether any of these will be deprecated by juju

src/workload.py
```diff
@@ -132,3 +132,47 @@ def disable_database(self) -> None:
...
+            logger.error(f"Error executing command '{cmd}': {e}")
+            return None
+
+        if output.returncode != 0:
+            return None
+
+        return output.stdout.strip()
+
+    @override
+    def get_private_ip(self) -> str | None:
```
Mehdi-Bendriss: how about we also change: `get_host_mapping` to be more fine grained, and report both the private and public address?

Mehdi-Bendriss (commented):
Thanks Smail, I left a few comments and questions

src/workload.py
skourta: I was going to discuss this with @reneradoi because I think that function ultimately needs to be refactored. Its place should not be in the cluster manager as it is unit-related.

src/workload.py
skourta: We do not. I used it to keep both private and public IP functions similar.

src/core/workload.py
skourta: They should not be in the cluster manager as these are unit-related. Since we are using juju they are workload independent we might want to move them from the workload file. I will think further about it.

src/workload.py
reneradoi: I agree, `get_host_mapping()` should be moved out of the cluster manager. I'm fine with adding the public ip to it.

src/workload.py
reneradoi: I like the approach of using the Juju hook command to get the address. This helps us keeping the code similar for both functions and, furthermore, helps with consistency. 

Imagine Juju gets sorted out ipv4 vs. ipv6 in mixed mode, we would have to make adjustments on our code to stay inline with whatever Juju uses for private ip then. By using the hook command, we don't have to worry about that.

In addition to that, we could have the same implementation for vm and k8s (some day...) using the hook command.

src/core/workload.py
reneradoi: I agree, these methods should be in either the `src/workload.py` or `src/core/workload.py`, with that latter being preferred by me but only possible if we chose to implement substrate-agnostic (meaning: using hook commands instead of socket).

reneradoi (dismissed):
I really like that change, thank you Smail!

src/workload.py
skourta: @reneradoi How about we move the  `get_host_mapping` to the workload too?

reneradoi: Yes, I agree with that.

Mehdi-Bendriss (approved):
Thanks Smail!

reneradoi (approved):
Nice change, thanks Smail!

PR #117, opened 2025-07-21
[DPE-7733] fix TLS switchover

reneradoi:
**Issue**

During the TLS switchover for client connections, it might happen that the operator tries to connect with TLS to etcd servers without TLS (or vice versa). This can happen due to a mismatch between:
- the client TLS state of another server in the databag at the beginning of the current event hook
- the actual TLS state of that server at the time an `etcdctl` command is executed by the operator code

If that server has restarted and changed client TLS mode during the event execution, the connection from operator code might fail.

This behavior mostly happens when the operator tries to update the peer url of it's local etcd server when switching TLS mode. It can be seen in various CI runs, for example [here](https://github.com/canonical/charmed-etcd-operator/actions/runs/16406465883/job/46361823833#step:13:1172) and [here](https://github.com/canonical/charmed-etcd-operator/actions/runs/16432100325/job/46435494512#step:13:2696) and [here](https://github.com/canonical/charmed-etcd-operator/actions/runs/16432100325/job/46435494535#step:13:3137). To mitigate the risk of failed connections, it would be better to only execute `etcdctl` commands against the local etcd server when updating the peer url settings.

**Solution**

This PR changes the behavior of the following methods to only execute the command against the local etcd server instead of all cluster endpoints:
- `cluster_manager.member`
- `cluster_manager.broadcast_peer_url()`

This reduces the risk of errors for the operator code in situations where a) client TLS mode is switched over or b) network disruptions happen.

Due to the changed behavior, it is required to rearrange the operations executed after a change of IP address (`config_changed`), because the peer url can only be updated against the local server after the server has been restarted and picked up its new IP address (it happened before the restart until this PR).

**Other changes:**
- add error handling for `certificate_available` event in case it is not possible to get the assigned certificates; this occasionally fails in integration tests, as can be seen  [here](https://github.com/canonical/charmed-etcd-operator/actions/runs/16406465883/job/46361823622#step:13:19668)
- while working on it: minor adjustments to integration test for `ca_rotation`: 
  - less waiting time because secret expiration for expiring certificates now already happens before the actual expiration (see: [secret expiration time in TLS lib](https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/lib/charms/tls_certificates_interface/v4/tls_certificates.py#L1202-L1204))
  - running on large runner (for performance reasons)

### DIFF
--- a/src/events/tls.py
+++ b/src/events/tls.py
@@ -179,7 +179,6 @@ class TLSEvents(Object):
             logger.error(
                 f"No assigned certificate found for the received certificate {event.certificate}"
             )
-            event.defer()
             return
 
         tls_state = (
--- a/src/events/tls.py
+++ b/src/events/tls.py
@@ -179,6 +179,7 @@ class TLSEvents(Object):
             logger.error(
                 f"No assigned certificate found for the received certificate {event.certificate}"
             )
+            event.defer()
             return
 
         tls_state = (
--- a/src/events/tls.py
+++ b/src/events/tls.py
@@ -172,8 +172,14 @@ class TLSEvents(Object):
             self.peer_certificate if cert_type == TLSType.PEER else self.client_certificate
         )
 
-        certs, private_key = relation_requirer.get_assigned_certificates()
-        cert = certs[0]
+        try:
+            certs, private_key = relation_requirer.get_assigned_certificates()
+            cert = certs[0]
+        except IndexError:
+            logger.error(
+                f"No assigned certificate found for the received certificate {event.certificate}"
+            )
+            return
 
         tls_state = (
             self.charm.state.unit_server.tls_peer_state
--- a/src/events/tls.py
+++ b/src/events/tls.py
@@ -169,19 +169,25 @@ class TLSEvents(Object):
         )
         peer_certificates, peer_private_key = self.peer_certificate.get_assigned_certificates()
 
-        if client_certificates and client_certificates[0].certificate == cert:
-            cert_type = TLSType.CLIENT
-            cert = client_certificates[0]
-            private_key = client_private_key
-            tls_state = self.charm.state.unit_server.tls_client_state
-            tls_ca_rotation_state = self.charm.state.unit_server.tls_client_ca_rotation_state
-        elif peer_certificates and peer_certificates[0].certificate == cert:
-            cert_type = TLSType.PEER
-            cert = peer_certificates[0]
-            private_key = peer_private_key
-            tls_state = self.charm.state.unit_server.tls_peer_state
-            tls_ca_rotation_state = self.charm.state.unit_server.tls_peer_ca_rotation_state
-        else:
+        try:
+            if client_certificates and client_certificates[0].certificate == cert:
+                cert_type = TLSType.CLIENT
+                cert = client_certificates[0]
+                private_key = client_private_key
+                tls_state = self.charm.state.unit_server.tls_client_state
+                tls_ca_rotation_state = self.charm.state.unit_server.tls_client_ca_rotation_state
+            elif peer_certificates and peer_certificates[0].certificate == cert:
+                cert_type = TLSType.PEER
+                cert = peer_certificates[0]
+                private_key = peer_private_key
+                tls_state = self.charm.state.unit_server.tls_peer_state
+                tls_ca_rotation_state = self.charm.state.unit_server.tls_peer_ca_rotation_state
+            else:
+                logger.error(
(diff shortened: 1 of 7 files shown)


src/managers/cluster.py
```diff
@@ -121,7 +121,7 @@ def member(self) -> Member:
         client = EtcdClient(
             username=self.admin_user,
             password=self.admin_password,
-            client_url=",".join(e for e in self.cluster_endpoints),
+            client_url=self.state.unit_server.client_url,
```
Mehdi-Bendriss, line 124: What would happen in the event where the current node is down while other nodes are up? 
Should we perhaps add a check to know if we _can_ use the current hosts and if not fallback on `cluster_endpoints`?

src/managers/cluster.py
```diff
@@ -150,7 +150,7 @@ def broadcast_peer_url(self, peer_urls: str) -> None:
...
```
Mehdi-Bendriss, line 153: similar question

tests/integration/tls/test_ca_rotation.py
```diff
@@ -258,9 +258,9 @@ async def test_ca_rotation_by_expiration(ops_test: OpsTest) -> None:
     )
     assert current_client_certificate, "Failed to get the current client certificate"
 
-    logger.info("Waiting ~6m for expiration of certificates - renewed certs will have a new CA")
-    # 6m + 30s for renewal to start
-    time.sleep(390)
+    logger.info("Waiting ~5.4m for expiration of certificates - renewed certs will have a new CA")
+    # the juju secret expires at 90% of the certificate validity; 360s * 0.9 = 324s
```
Mehdi-Bendriss, line 262: I like this 😄

Mehdi-Bendriss (commented):
Thanks René, good investigation! I have one main question

src/managers/cluster.py
skourta: My opinion is that we need to self-heal the current node in that situation. Even if we query the other members for info they will not have the info of this member (which this function tries to get). And the same goes for the `broadcast_peer_url`. If the node is down, it serves nothing to broadcast its new URL.

skourta (dismissed):
Thank you @reneradoi. 
I would like us to run the CI a couple more times for the tests to check the stability of the test results.

src/managers/cluster.py
reneradoi: Good question, happy that you're raising it. I've thought about it carefully and opted against having a fallback to all cluster endpoints in case of failure. (Just didn't write it into the already long PR description...)

The main reason for not doing so: It is not required and would complicate things. Not required because the places where the `member` method is used either have error handling in place in case `member` raises a `ValueError` (which it does if it fails), or as @skourta pointed out, the execution should fail if the member is down.

I've checked all places where these are called and I'm confident we don't need to add a fallback. Are you ok with that @Mehdi-Bendriss?

Mehdi-Bendriss (dismissed):
Thanks René!

skourta (dismissed):

skourta (approved):

Mehdi-Bendriss (approved):

src/events/tls.py
```diff
@@ -169,19 +169,25 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
...
+                cert = peer_certificates[0]
+                private_key = peer_private_key
+                tls_state = self.charm.state.unit_server.tls_peer_state
+                tls_ca_rotation_state = self.charm.state.unit_server.tls_peer_ca_rotation_state
+            else:
+                logger.error(
+                    f"Received certificate does not match any assigned certificates: {cert}"
+                )
+                return
+        except IndexError:
```
Mehdi-Bendriss, line 190: When would this ever get raised since we are already checking on `client_certificates` not being None nor empty?

Mehdi-Bendriss (commented):
Thanks René, I have a comment

src/events/tls.py
reneradoi: This is what is referred to here in the description:

> add error handling for certificate_available event in case it is not possible to get the assigned certificates; this occasionally fails in integration tests, as can be seen [here](https://github.com/canonical/charmed-etcd-operator/actions/runs/16406465883/job/46361823622#step:13:19668)

In a very rare edge case, when the received certificate is already expired (which should not happen in reality...), the TLS lib will not return it anymore in the `get_assigned_certificates()` call, which returns an uncaught error for the operator code, failing with `IndexError` when getting the first item of the expected list.

PR #118, opened 2025-07-24
chore: fixes detected with k8s integration

skourta:
* [`src/events/etcd.py`](diffhunk://#diff-4d4b45129d72bf79a08e70009a59f4ebe0a016b4727117155249c864ae4a1b49L196-R200): Added logic to open the client port (`CLIENT_PORT`) when the workload starts successfully, along with a corresponding log message. 

* [`src/managers/external_clients.py`](diffhunk://#diff-7cf18beef6c6df0f0e1a46e25a3b1bdb9aec8f5a433fb89bd09549560bfca7bfL135-R138): Updated the `update_client_relations_data` method to filter cluster servers based on their `tls_client_state`, ensuring only servers with `TLSState.TLS` are included. 

* [`src/managers/external_clients.py`](diffhunk://#diff-7cf18beef6c6df0f0e1a46e25a3b1bdb9aec8f5a433fb89bd09549560bfca7bfR147-R150): Added a check to skip relations that do not have the "prefix" and "mtls-cert" fields set, ensuring that only fully initialised relations are processed in the `update_client_relations_data` method.

### DIFF
--- a/src/managers/external_clients.py
+++ b/src/managers/external_clients.py
@@ -141,6 +141,10 @@ class ExternalClientsManager:
         server_ca = self.state.tls_client_certificate.ca.raw
 
         for relation in self.state.etcd_provides.relations:
+            if not self.state.etcd_provides.fetch_relation_field(relation.id, "prefix"):
+                # Skip relations that are not ready yet
+                continue
+
             relation_data = self.state.etcd_provides.fetch_my_relation_data(
                 [relation.id], ["uris", "endpoints", "tls-ca", "version"]
             )[relation.id]
--- a/src/managers/external_clients.py
+++ b/src/managers/external_clients.py
@@ -144,8 +144,10 @@ class ExternalClientsManager:
         server_ca = self.state.tls_client_certificate.ca.raw
 
         for relation in self.state.etcd_provides.relations:
-            if not self.state.etcd_provides.fetch_relation_field(relation.id, "prefix"):
-                # Skip relations that have not requested a prefix yet
+            if not self.state.etcd_provides.fetch_relation_field(
+                relation.id, "prefix"
+            ) or not self.state.etcd_provides.fetch_relation_field(relation.id, "mtls-cert"):
+                # Skip relations with invalid paylods
                 continue
 
             relation_data = self.state.etcd_provides.fetch_my_relation_data(
--- a/src/managers/external_clients.py
+++ b/src/managers/external_clients.py
@@ -147,7 +147,8 @@ class ExternalClientsManager:
             if not self.state.etcd_provides.fetch_relation_field(
                 relation.id, "prefix"
             ) or not self.state.etcd_provides.fetch_relation_field(relation.id, "mtls-cert"):
-                # Skip relations with invalid paylods
+                # Skip relations with invalid payloads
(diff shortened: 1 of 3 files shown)


src/managers/external_clients.py
```diff
@@ -141,6 +141,10 @@ def update_client_relations_data(self, etcd_version: str) -> None:
         server_ca = self.state.tls_client_certificate.ca.raw
 
         for relation in self.state.etcd_provides.relations:
+            if not self.state.etcd_provides.fetch_relation_field(relation.id, "prefix"):
+                # Skip relations that are not ready yet
+                continue
```
Mehdi-Bendriss: I have a few questions here:
1. Why restrict it to the "prefix" field and not the cert as well?
2. Do we not not have validation as part of the requirer lib? 
3. The comment should be more generic, as ETCD is oblivious to other charms being ready or not: e.g: `Skip relations with invalid payloads.`

Mehdi-Bendriss (commented):
Thanks Smail, I have a couple of small questions

reneradoi (dismissed):
I really like that we are finding these things early before the software gets deployed by others!

src/managers/external_clients.py
skourta: 1. That is what is being used to check in the data interfaces lib (RESOURCE_FIELD).
2. This actually comes from an error raised in the requirer lib. If the client did not request the RESOURCE_FIELD yet a `PrematureDataAccessError` error is raised.

skourta: Updated

src/managers/external_clients.py
```diff
@@ -132,7 +132,10 @@ def update_client_relations_data(self, etcd_version: str) -> None:
             uri.split("=")[0] for uri in self.state.cluster.cluster_members.split(",")
         }
         cluster_servers = {
-            server for server in self.state.servers if server.member_name in cluster_server_names
+            server
+            for server in self.state.servers
+            if server.member_name in cluster_server_names
+            and server.tls_client_state == TLSState.TLS
```
Mehdi-Bendriss, line 138: This is nice

src/managers/external_clients.py
```diff
@@ -141,6 +144,12 @@ def update_client_relations_data(self, etcd_version: str) -> None:
         server_ca = self.state.tls_client_certificate.ca.raw
 
         for relation in self.state.etcd_provides.relations:
+            if not self.state.etcd_provides.fetch_relation_field(
+                relation.id, "prefix"
+            ) or not self.state.etcd_provides.fetch_relation_field(relation.id, "mtls-cert"):
+                # Skip relations with invalid paylods
```
Mehdi-Bendriss:
```suggestion
                # Skip relations with invalid payloads
```

Mehdi-Bendriss: Could you please add an error log here?

Mehdi-Bendriss (commented):
Thanks Smail! Tiny comments and we should be good

src/managers/external_clients.py
skourta: I added it as a warning since this is not technically an error

Mehdi-Bendriss (approved):

reneradoi (approved):
Re-approval after minor changes

PR #119, opened 2025-07-24
[DPE-7630] make leader/timeout parameters configurable

reneradoi:
In high-latency network environments, it could happen that the fixed configuration values we currently use for `heartbeat-interval` and `election-timeout` cause many leader changes. Therefore we should make these values configurable to users.

**UI/UX changes:**
- introduce config options `election-timeout` and `heartbeat-interval`
- add a `BlockedStatus` in case the config options are set to invalid values

**Implementation:**
- add `compute_component_status()` method to the `ConfigManager`
- when running the `config_changed` hook, initiate a rolling restart if the configuration values are valid and different from what is currently configured
- add a `tuning_parameters_valid` property to the `ConfigManager`; this checks the validity of the config values and is used for the `config_changed` event as well as the status computation
- add a `requires_restart()` method to the `ConfigManager` which loads the current config file from the unit and checks if one of the tuning parameters was updated
- adjust the `config_properties()` method of the `ConfigManager` to only apply the config values if they are considered valid
- add a `load_yaml_file()` method to the `EtcdWorkload` class that can be used to load the current config file and returns a dictionary with the configuration parameters

**Architecture decisions:**
- config validation is done by a simple method in the `ConfigManager` instead of introducing a "structured config class" or config model; reasons: keep it simple, avoid possible future backwards incompatibilities for config validation
- config validation is executed (and thereby applied) whenever configuration is written to the unit; this helps to group restarts for different kind of changes (e.g. config change because of new ip address after network disruption + adjusted config values)
- `restart_required` validation happens against the current config file because there is no other way of querying the currently applied values for the tuning parameters (even though "live" query would be best e.g. via metrics, but this is not possible)

**Testing:**
- add unit testing for the `config_changed` event
- add integration testing to `test_ha_on_rolling_restart.py` to ensure the rolling restart is performed maintaining HA when updating these config values

### DIFF
--- a/src/managers/config.py
+++ b/src/managers/config.py
@@ -85,6 +85,11 @@ class ConfigManager:
             f"http://{self.state.unit_server.ip}:{METRICS_PORT}"
         )
 
+        if self.tuning_parameters_valid:
+            for option in TuningOptions:
+                # take over the config value set by users
+                config_properties[option.value] = self.config.get(option.value)
+
         if self.state.unit_server.tls_client_state in [TLSState.TO_TLS, TLSState.TLS]:
             # replace http with https in listen-client-urls and advertise-client-urls
             config_properties["listen-client-urls"] = self.state.unit_server.client_url.replace(
--- a/src/managers/config.py
+++ b/src/managers/config.py
@@ -85,7 +85,7 @@ class ConfigManager:
             f"http://{self.state.unit_server.ip}:{METRICS_PORT}"
         )
 
-        if self.tuning_parameters_valid:
+        if self.are_tuning_parameters_valid():
             for option in TuningOptions:
                 # take over the config value set by users
                 config_properties[option.value] = self.config.get(option.value)
@@ -155,8 +155,7 @@ class ConfigManager:
             file=self.config_file,
         )
 
-    @property
-    def tuning_parameters_valid(self) -> bool:
+    def are_tuning_parameters_valid(self) -> bool:
         """Validate configuration values for tuning parameters.
 
         Returns:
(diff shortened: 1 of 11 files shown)


src/managers/config.py
```diff
@@ -141,6 +155,47 @@ def set_config_properties(self) -> None:
             file=self.config_file,
         )
 
+    @property
+    def tuning_parameters_valid(self) -> bool:
```
Mehdi-Bendriss: nit, I don't think this qualifies as a property, better as a method 
```suggestion
    def is_tuning_parameters_valid(self) -> bool:
```

tests/integration/ha/test_ha_on_rolling_restart.py
```diff
@@ -83,3 +83,31 @@ async def test_disable_and_enable_peer_tls(ops_test: OpsTest) -> None:
     assert_continuous_writes_increasing(endpoints=endpoints, user=INTERNAL_USER, password=password)
     stop_continuous_writes()
     assert_continuous_writes_consistent(endpoints=endpoints, user=INTERNAL_USER, password=password)
+
+
```
Mehdi-Bendriss, line 87: Could you please add a test to check for the status for invalid config values? and confirm that etcd is still running despite the blocked status / invalid config

Mehdi-Bendriss (commented):
Thanks René! Looks good, left a small comment

src/managers/config.py
reneradoi: Would be possible to argument for both sides here I'd say, no strong opinion though from my side -> changed it.

tests/integration/ha/test_ha_on_rolling_restart.py
reneradoi: Have added it to ensure the workload keeps running. We also have unit test coverage for validation and status.

Mehdi-Bendriss (approved):
Thanks René, great work!

skourta (approved):
Thank you @reneradoi. Great work!
For the unit tests we can also patch the `load_yaml_file` in conftest for all tests and then override its values only for its tests.

Issue #122, opened 2025-07-25
Region should be restricted for bucket creation


Gu1nness:
When testing a code similar to the one you have for bucket creation, I came to the painful realisation that setting the region to `us-east-1` was leading to an error when creating the bucket on AWS.
This is due to some decisions made to refuse the specification of region if it is the default value (got some information [here](https://github.com/aws/aws-sdk-js/issues/3647) and [here](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateBucket.html#API_CreateBucket_RequestSyntax) and [here](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#createBucket-property).

I have not checked yet if that also applies to GCS.

Gu1nness:
Final comment on this one: We might face the situation where the bucket exists but the credentials we're provided with do not allow bucket creation, so it's better to first check for bucket existence (that we can do) before trying to create it.
Postgresql has a nice pice of code for that [here](https://github.com/canonical/postgresql-operator/blob/main/src/backups.py#L271-L326)

reneradoi:
Thank you @Gu1nness for bringing this up. It has been solved in https://github.com/canonical/charmed-etcd-operator/pull/123.

Issue #121, opened 2025-07-25
Unprotected call to create_bucket


Gu1nness:
If the bucket creation fails during the s3-credentials-relation-changed event fails, the charm ends up in error.

The call to create_bucket [here](https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/src/events/backup.py#L117) is not protected like it is in the `create-backup` action.

There should be an error catching of `EtcdBackupError` and set a dedicated status.

reneradoi:
Thank you @Gu1nness for bringing this up. It has been solved in https://github.com/canonical/charmed-etcd-operator/pull/123.

PR #123, opened 2025-07-28
[DPE-7811] Error handling for backup/restore setup

reneradoi:
**Issues**
- If setting up object storage for backup/restore fails, the operator currently does not catch the error but leaves the charm in error state.
- When using S3 storage, it is not possible to provide a `LocationConstraint` when using the default S3 region of AWS.

Solves #121 and #122 

**Solution**
- add error handling for errors when creating an S3 bucket or Azure storage container
- do not set a `LocationConstraint` when setting up the S3 bucket if the region is set to AWS default `us-east-1`

### DIFF
--- a/src/events/backup.py
+++ b/src/events/backup.py
@@ -114,7 +114,12 @@ class BackupEvents(Object):
         s3_parameters["path"] = s3_parameters["path"].strip("/")
         s3_parameters["bucket"] = s3_parameters["bucket"].strip("/")
 
-        self.charm.backup_manager.create_bucket(s3_parameters)
+        try:
+            self.charm.backup_manager.create_bucket(s3_parameters)
+        except EtcdBackupError:
+            logger.error("Cannot setup s3 object storage - check credentials and permissions")
+            return
+
         self.charm.state.cluster.update({"s3-credentials": json.dumps(s3_parameters)})
 
     def _on_s3_credentials_gone(self, event: CredentialsGoneEvent) -> None:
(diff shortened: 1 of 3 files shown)


src/managers/backup.py
```diff
@@ -122,6 +124,9 @@ def create_container(self, azure_parameters: dict[str, str]) -> None:
             logger.info(f"Container {azure_parameters['container']} created")
         except ResourceExistsError:
             logger.info(f"Container {azure_parameters['container']} already exists")
+        except AzureError as e:
```
skourta, line 127: Can we be a bit more specific on the Azure error?

src/events/backup.py
```diff
@@ -114,7 +114,12 @@ def _on_s3_credentials_changed(self, event: CredentialsChangedEvent) -> None:
         s3_parameters["path"] = s3_parameters["path"].strip("/")
         s3_parameters["bucket"] = s3_parameters["bucket"].strip("/")
 
-        self.charm.backup_manager.create_bucket(s3_parameters)
+        try:
+            self.charm.backup_manager.create_bucket(s3_parameters)
+        except EtcdBackupError:
+            logger.error("Cannot setup s3 object storage - check credentials and permissions")
```
Mehdi-Bendriss, line 120: We should probably add a blocked status reflecting that the object storage is not configured due to missing requirements (this should play nicely in the status PR @skourta is working on)

Mehdi-Bendriss (commented):
Thanks René, looks good! I have one comment

src/managers/backup.py
reneradoi: The list of possible exceptions is very long (see: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/exceptions.py#L61), and I'd prefer to not handle them specifically.

src/events/backup.py
reneradoi: This is covered, as we do not take over the credentials to our databag and by that trigger `"Missing or invalid s3/azure credentials"` on status computation.

Mehdi-Bendriss (approved):

skourta (approved):

PR #124, opened 2025-07-28
[DPE-7666]: advanced statuses

skourta:
This pull request the advanced statuses lib in etcd.

### Status Handling Improvements:
* Integrated the `StatusHandler` from `data_platform_helpers` into the charm to centralize and prioritize status management. (`src/charm.py`: [[1]](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cL13-R13) [[2]](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR58-R66)
* Replaced legacy status methods (`set_status`, `_on_collect_status`) with the new `StatusHandler`, removing redundant code. (`src/charm.py`: [[1]](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cL85-L95) [[2]](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cL278-L311)
* Updated backup and cluster event handlers to use the new status management system (`state.statuses`) for adding and deleting statuses. (`src/events/backup.py`: [[1]](diffhunk://#diff-ad959188bbaf55316575b2e26d4d155b09e71de16ce05ea31339fe6089fea1e1L277-R283) [[2]](diffhunk://#diff-ad959188bbaf55316575b2e26d4d155b09e71de16ce05ea31339fe6089fea1e1R298-R308) [[3]](diffhunk://#diff-ad959188bbaf55316575b2e26d4d155b09e71de16ce05ea31339fe6089fea1e1L307-R332); `src/events/etcd.py`: [[4]](diffhunk://#diff-4d4b45129d72bf79a08e70009a59f4ebe0a016b4727117155249c864ae4a1b49L104-R108) [[5]](diffhunk://#diff-4d4b45129d72bf79a08e70009a59f4ebe0a016b4727117155249c864ae4a1b49L188-R224)

### Dependency Management:
* Added `data-platform-helpers` as a new dependency in `pyproject.toml` to support advanced status handling. [[1]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R16) [[2]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R66)

### Code Simplification and Cleanup:
* Removed the legacy `Status` and `DebugLevel` enums, along with outdated status-related attributes and methods. (`src/charm.py`: [[1]](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cL25-L26) [[2]](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cL48)
* Deleted the `common/secrets.py` utility file and relocated its functionality to the `ClusterState` class for better encapsulation. (`src/common/secrets.py`: [[1]](diffhunk://#diff-af2274dbc7e6a3eac83daf0f9e75a5f75acb6adf42667e0eba21098bcd0b3819L1-L30) `src/core/cluster.py`: [[2]](diffhunk://#diff-728670c1b94eb83ab287f8abf7b8c88c7c80b237e1ea8a8191a942976dcb42dcR171-R189)

### Cluster State Enhancements:
* Extended the `ClusterState` class to implement the `StatusesStateProtocol` interface, enabling it to manage statuses via the `statuses` attribute. (`src/core/cluster.py`: [[1]](diffhunk://#diff-728670c1b94eb83ab287f8abf7b8c88c7c80b237e1ea8a8191a942976dcb42dcL39-R41) [[2]](diffhunk://#diff-728670c1b94eb83ab287f8abf7b8c88c7c80b237e1ea8a8191a942976dcb42dcR52-R54)
* Added a new `is_leader` property to the `EtcdCluster` model for determining leadership within the cluster. (`src/core/models.py`: [src/core/models.pyR187-R191](diffhunk://#diff-e46c88bb687d77b94764a732b3e08a0a7bd544a64e606429c0b787c13e1c226aR187-R191))

### Metadata and Action Updates:
* Introduced a new `status-peers` relation in `metadata.yaml` and a `status-detail` action in `actions.yaml` to support enhanced status querying and management. (`metadata.yaml`: [[1]](diffhunk://#diff-d186204ea7abb85bbadd9d94ab29d5bcefcc43d6a0ed06015e98a71b2f099423R23-R24) `actions.yaml`: [[2]](diffhunk://#diff-ac0f9c9b30ee833f0e17ba061702fbe3cb698f1e3e08806b45a95c144e5a8e71R29-R36)


src/core/models.py
```diff
@@ -184,6 +184,11 @@ def restore_step(self) -> RestoreStep:
         """Get the current progress of the restore workflow."""
         return RestoreStep(self.relation_data.get("restore_step", ""))
 
+    @property
+    def is_leader(self) -> bool:
```
reneradoi: This could be misunderstood as the Raft leader. Let's rename it to either `is_leader_unit` or `is_juju_leader`, please.

reneradoi (commented):
Hey Smail! Thank you for this PR!

I don't have major concerns regarding the implementation. Separating the statuses per component makes it even more easy to have an overview, compared to the list in more or less alphabetical order we had before. That is nice.

Status handling in the operator code itself is also looking good from my side. It does get a bit more complicated with resetting statuses, but that's the downside of maintaining a cache of statuses.

One thing that I don't really like is how the status info cloggs the log with `info` messages on every hook, example [here](https://github.com/canonical/charmed-etcd-operator/actions/runs/16568497194/job/46854822183?pr=124#step:13:1623). Is there a way to configure this to `debug` instead?

Integration testing with `wait_until` gets a bit more tricky now as well. But from this first review I'd say this is the only main thing that needs to be adjusted, other than that, it's already in pretty good shape.

pyproject.toml
```diff
@@ -55,14 +56,15 @@ ops = "^2.17.0"
...
-pytest-operator = "^0.28.0"
+pytest-operator = "^0.43.1"
 juju = "==3.6.0"
 allure-pytest = "*"
 allure-pytest-default-results = "^0.1.2"
 tenacity = "*"
 pydantic = ">= 1.0"
 boto3 = "^1.37.12"
 mypy-boto3-s3 = "^1.37.0"
+data-platform-helpers = ">=0.1.6"
```
Mehdi-Bendriss:
```suggestion
data-platform-helpers = ">=0.1.7"
```

src/events/etcd.py
```diff
@@ -99,7 +98,12 @@ def _on_storage_attached(self, event: ops.StorageAttachedEvent) -> None:
     def _on_install(self, event: ops.InstallEvent) -> None:
         """Handle install event."""
         if not self.charm.workload.install():
-            self.charm.set_status(Status.SERVICE_NOT_INSTALLED)
+            self.charm.status.set_running_status(
```
Mehdi-Bendriss, line 102: I don't think this should be a running status, a running status implies an operation in progress during the hook execution.

As a matter of fact, we should crash this hook so it prevents subsequent hooks from running and cascading the failure.

src/events/etcd.py
```diff
@@ -183,20 +187,39 @@ def _on_start(self, event: ops.StartEvent) -> None:  # noqa: C901
...
+                component=self.charm.cluster_manager.name,
+            )
+            self.charm.state.statuses.delete(
+                EtcdServiceStatuses.SERVICE_STARTING.value,
+                scope="unit",
+                component=self.charm.cluster_manager.name,
+            )
         else:
-            self.charm.set_status(Status.SERVICE_NOT_RUNNING)
+            self.charm.status.set_running_status(
```
Mehdi-Bendriss, line 222: Similar comment

src/events/external_clients.py
```diff
@@ -63,18 +62,15 @@ def _on_mtls_cert_updated(self, event: MTLSCertUpdatedEvent) -> None:  # noqa: C
         """Handle the ca chain updated event."""
         if not event.mtls_cert or not event.prefix:
             logger.error("CA chain, keys prefix, or common name not provided")
-            self.charm.set_status(Status.EC_MISSING_CREDENTIALS)
```
Mehdi-Bendriss, line 66: I'm not clear why these statuses are no longer shown? Are they offloaded to another module? in which case how are the ones specific to this one handled, like `EC_USERNAME_EXISTS` etc?

Mehdi-Bendriss (commented):
Thanks Smail, good work! 
I would like us to take a second look at: 
- what is currently set as a running status 
- the statuses that we deleted like `EC_USERNAME_EXISTS` and how exactly are they handled now

src/events/external_clients.py
skourta: These are offloaded to the `get_statuses` function of the `ExternalClientsManager`

metadata.yaml
```diff
@@ -20,6 +20,8 @@ peers:
     interface: etcd_peers
   restart:
     interface: rolling_op
+  status-peers:
+    interface: status_peers
```
Gu1nness, line 24: Nit: should be `status-peers` to be fully compliant with the spec but it seems you have a naming scheme here so I understand the choice.

src/events/etcd.py
```diff
@@ -183,20 +187,39 @@ def _on_start(self, event: ops.StartEvent) -> None:  # noqa: C901
...
+            self.charm.state.statuses.delete(
+                EtcdServiceStatuses.SERVICE_NOT_RUNNING.value,
+                scope="unit",
+                component=self.charm.cluster_manager.name,
+            )
+            self.charm.state.statuses.delete(
+                EtcdServiceStatuses.SERVICE_STARTING.value,
+                scope="unit",
+                component=self.charm.cluster_manager.name,
+            )
```
Gu1nness: Maybe it would be easier to clear here instead of removing the statuses one by one ?

src/events/etcd.py
```diff
@@ -183,20 +187,39 @@ def _on_start(self, event: ops.StartEvent) -> None:  # noqa: C901
...
         else:
-            self.charm.set_status(Status.SERVICE_NOT_RUNNING)
+            self.charm.status.set_running_status(
+                EtcdServiceStatuses.SERVICE_NOT_RUNNING.value,
+                scope="unit",
+                component_name=self.charm.cluster_manager.name,
+                statuses_state=self.charm.state.statuses,
+            )
```
Gu1nness, line 227: I have the feeling that status should be a regular status:
 * It can wait the end of the event to display (the end being right after)
 * It's a status that can be set independently in the `get_statuses` function
 * There is no need to immediately tell the operator, it does not require urgent action it seems ?
You can have a look at the [UX spec](https://docs.google.com/document/d/1SV11ct-flQkc5BOYOeXgmPeglL8bVs-mDVkGjG20K48/edit?tab=t.0#heading=h.ny83sa5d7c8e) if you want more details on when running statuses should be set, they should be kept for long running actions (lifecycle actions like restarting a database or removing a replica, because a lack of information provided to the user could lead to thinking the event is 'stuck') or juju actions (immediate information provided by the action like displaying a backup id in the status.)

src/events/etcd.py
```diff
@@ -355,7 +381,12 @@ def _on_leader_elected(self, event: LeaderElectedEvent) -> None:
 
         try:
             if self.charm.cluster_manager.is_cluster_failed:
-                self.charm.set_status(Status.CLUSTER_FAILED)
+                self.charm.status.set_running_status(
```
Gu1nness: I think this should be a regular status: the event finishes right after that status being set.

src/events/etcd.py
```diff
@@ -375,12 +406,22 @@ def _on_update_status(self, event: ops.UpdateStatusEvent) -> None:
 
         if not self.charm.workload.alive():
             if not self.charm.cluster_manager.restart_member():
-                self.charm.set_status(Status.SERVICE_NOT_RUNNING)
+                self.charm.status.set_running_status(
```
Gu1nness, line 398: As I said earlier for that status, I think it should be a regular status.

src/managers/backup.py
```diff
@@ -392,26 +397,31 @@ def proceed_restore_workflow_if_possible(self) -> None:
             logger.info(f"Next restore step: {next_step.value}")
             self.state.cluster.update({"restore_instruction": next_step.value})
 
-    def compute_component_status(self) -> List[Status]:
+    def get_statuses(self, scope: Scope, recompute: bool = False) -> list[StatusObject]:
         """Compute the Backup manager's statuses."""
-        status_list = []
+        if recompute:
+            self.state.statuses.clear(scope=scope, component=self.name)
```
Gu1nness: That shouldn't be necessary: If it's recomputed to be stored (eg: update-status event) then the handler takes care of doing so.
Otherwise, you could just request the statuses to be recomputed without having a side effect on the databag.

src/managers/cluster.py
```diff
@@ -353,31 +361,44 @@ def update_cluster_member_state(self) -> None:
             # we should not have errors here, but if we do, we don't want the error to raise
             logger.warning(f"Error updating the cluster member state: {e}")
 
-    def compute_component_status(self) -> List[Status]:
+    def get_statuses(self, scope: Scope, recompute: bool = False) -> list[StatusObject]:
         """Compute the Cluster manager's statuses."""
-        status_list = []
+        status_list: list[StatusObject] = self.state.statuses.get(
+            scope=scope, component=self.name
+        ).root
```
Gu1nness: If you're recomputing, you're recomputing and you should start from scratch.
If you're not recomputing, then you just return the content of that list.
Example: https://github.com/canonical/mongo-single-kernel-library/blob/6/edge/single_kernel_mongo/managers/mongo.py#L503-L553

A risk of what you're doing here is for example to end up with the same status multiple times in the list.

src/managers/config.py
```diff
@@ -186,14 +191,14 @@ def requires_restart(self) -> bool:
...
         """Compute the Cluster manager's statuses."""
-        status_list = []
+        status_list: list[StatusObject] = []
 
         if not self.are_tuning_parameters_valid():
-            status_list.append(Status.TUNING_CONFIG_INVALID)
+            status_list.append(ConfigStatuses.TUNING_CONFIG_INVALID.value)
 
-        return status_list
+        return status_list if status_list else [CharmStatuses.ACTIVE_IDLE.value]
```
Gu1nness, line 202: It feels weird to read it that way in the first place but it's good for the future if you have more cases to add, so I love it!

src/statuses.py
```diff
@@ -0,0 +1,157 @@
...
+    """Collection of cluster related statuses."""
+
+    CLUSTER_INITIALIZING = StatusObject(
+        status="maintenance", message="Initializing etcd cluster..."
+    )
+    CLUSTER_FAILED = StatusObject(
+        status="blocked",
+        message="Cluster failure - majority of cluster members lost",
+        running="async",
+        action="rebuild-cluster",
```
Gu1nness: It's maybe not very clear but you can be a bit more verbose on the "action" field, because it represents an action to take (as an operator). It could be to run a juju action, or check some configuration, etc.

```suggestion
        action="Run action rebuild-cluster",
```

Gu1nness (changes requested):
Excellent PR !
A few suggestions here and there and some more important points: I have the feeling that some statuses are set as running statuses while they should not be running statuses.
I have linked the spec of the UX of statuses explaining what should be running statuses if it can help.

src/managers/cluster.py
skourta: This is necessary to pick up on some of the statuses that need to be set and removed in charm code and not at the `get_statuses`. An example is `PASSWORD_UPDATE_FAILED`. Maybe I should add a clear when the charm is initialized?

Gu1nness: Ah ! then what you can do is `get(scope=scope, component=self.name, running_status_only=False, running_status_type="async")` to select only the running statuses. I think that password_update_failed should be a running status (maybe even an approved_critical_status) so that you can display it right away and it'll stay on top of everything.

skourta: That was just an example. There are a few of these that can only be set through the code. We do not have a way to detect them from the state. Other examples: `CLUSTER_MANAGEMENT_ERROR`, and `NO_PEER_RELATION`. Should we have them all as running_statuses?

Gu1nness (dismissed):

tests/integration/ha/test_disaster_recovery.py
```diff
@@ -141,10 +142,15 @@ async def test_recover_from_majority_failure(ops_test: OpsTest) -> None:
...
-            leader_unit = unit
+        leader_unit = None
+        while leader_unit is None:
+            for unit in ops_test.model.applications[APP_NAME].units:
+                if await unit.is_leader_from_status():
+                    leader_unit = unit
+
+            if leader_unit is None:
+                logger.info("Waiting for a leader to be elected")
+                time.sleep(10)
```
Mehdi-Bendriss, line 153: Could you explain this?

Mehdi-Bendriss (commented):
Thanks Smail! All good from my side, besides 1 question on the tests

tests/integration/ha/test_disaster_recovery.py
reneradoi: We discussed about it, seems like there was a test failure on a local run where there was no Juju leader. I haven't seen this so far, but this is just a safeguard to make sure the test doesn't fail unnecessarily.

This could theoretically happen here because we axe away two units (without considering if one of them is the Juju leader), it may take a while until a new Juju leader is appointed.

If we ever see this at another place, we should consider it a potential bug and investigate further.

skourta: This wasn't just locally. It happened multiple times on the CI as well.
Example: https://github.com/canonical/charmed-etcd-operator/actions/runs/16650262145/job/47121446343?pr=124#step:8:4575

src/events/tls.py
```diff
@@ -355,7 +380,7 @@ def update_private_key(self, private_key_id: str) -> None:
         logger.debug("Updating TLS private key.")
 
         if self.read_and_validate_private_key(private_key_id) is None:
```
reneradoi: As this method was moved to the TLS manager, this call needs to be adjusted and `read_and_validate_private_key()` should be removed from the TLS events class.

reneradoi (changes requested):
Only one thing remaining: cleaning up the moved method `read_and_validate_private_key()` to avoid duplicate code.

reneradoi (dismissed):
Great! Thank you Smail!

reneradoi (approved):
Re-approval after fixing integration tests.

Mehdi-Bendriss (approved):
Thank you Smail!

PR #126, opened 2025-07-29
[DPE-7732] add config option for certificate-extra-sans

reneradoi:
**UI/UX changes:**

- introduce config option `certificate-extra-sans` to configure DNS names or IP addresses to be added as `SubjectAlternativeName` to the certificates used by etcd for client- and peer-communication
- add a BlockedStatus in case the config option is set to an invalid value (invalid IP address or invalid DNS name)

**Implementation:**
- when running the `config_changed` hook, refresh TLS certificates if TLS is enabled, the configuration value is valid and the `SubjectAlternativeNames` are different from the current certificates
- add methods to `TLSManager`: `extra_sans_config_is_valid`, `build_sans_dns`, `get_current_sans` and `certificate_sans_updated` to handle the processing and update-check for certificate sans
- add helper methods `_is_ip_address()` and `_is_hostname()` to `TLSManager` for validation of the config values
- check validity of `certificate-extra-sans` configuration value in TLS status computation

**Testing:**
- add unit test coverage for the various validation checks on DNS names and IP addresses
- add integration test coverage for setting the `certificate-extra-sans` config to valid and invalid values as well as resetting it

**Refactoring:**
- move TLS related parts of `config_changed` and `secret_changed` event handlers from `EtcdEvents` class to `TLSEvents` class
- rename TLS-manager's `get_sans_ip()` to `build_sans_ip` and refactor it internally
- add a return value to `EtcdWorkload.exec()` to capture the output of `openssl` commands needed to inspect the currently installed TLS certificates
- added the fallback for retrieval of the private ip to `EtcdWorkload.get_private_ip()` instead of handling it outside of the method

### DIFF
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -41,8 +41,6 @@ from literals import (
     SNAP_GROUP,
     SNAP_LOG_PATH,
     SNAP_USER,
-    TLS_CLIENT_PRIVATE_KEY_CONFIG,
-    TLS_PEER_PRIVATE_KEY_CONFIG,
     Status,
     TLSState,
     TLSType,
@@ -200,7 +198,7 @@ class EtcdEvents(Object):
         else:
             self.charm.set_status(Status.SERVICE_NOT_RUNNING)
 
-    def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:  # noqa: C901
+    def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
         """Handle config_changed event."""
         if (
             self.charm.state.cluster.is_restore_in_progress
@@ -235,12 +233,6 @@ class EtcdEvents(Object):
                 TLSState.TLS,
             ) or self.charm.state.unit_server.tls_peer_state in (TLSState.TO_TLS, TLSState.TLS):
                 self.charm.tls_events.refresh_tls_certificates_event.emit()
-
-        if tls_peer_private_key_id := self.charm.config.get(TLS_PEER_PRIVATE_KEY_CONFIG):
-            self.update_private_key(tls_peer_private_key_id)
-
-        if tls_client_private_key_id := self.charm.config.get(TLS_CLIENT_PRIVATE_KEY_CONFIG):
-            self.update_private_key(tls_client_private_key_id)
 
         if (
             self.charm.config_manager.are_tuning_parameters_valid()
@@ -420,17 +412,6 @@ class EtcdEvents(Object):
 
     def _on_secret_changed(self, event: ops.SecretChangedEvent) -> None:
         """Handle the secret_changed event."""
-        if tls_peer_private_key_id := self.charm.config.get(TLS_PEER_PRIVATE_KEY_CONFIG):
-            if tls_peer_private_key_id == event.secret.id:
-                self.update_private_key(tls_peer_private_key_id)
-
-        if tls_client_private_key_id := self.charm.config.get(TLS_CLIENT_PRIVATE_KEY_CONFIG):
-            if tls_client_private_key_id == event.secret.id:
-                self.update_private_key(tls_client_private_key_id)
-
-        if not self.charm.unit.is_leader():
-            return
-
         if (
             self.charm.state.cluster.is_restore_in_progress
             or self.charm.state.cluster.rebuild_cluster_in_progress
@@ -439,6 +420,9 @@ class EtcdEvents(Object):
                 "Cannot update credentials while a restore or cluster-rebuild operation is in progress."
             )
             event.defer()
+            return
+
+        if not self.charm.unit.is_leader():
             return
 
         if admin_secret_id := self.charm.config.get(INTERNAL_USER_PASSWORD_CONFIG):
@@ -528,16 +512,6 @@ class EtcdEvents(Object):
         except (ModelError, SecretNotFoundError) as e:
             logger.error(e)
             self.charm.set_status(Status.PASSWORD_UPDATE_FAILED)
-
-    def update_private_key(self, private_key_id: str) -> None:
-        """Update the private key in etcd."""
-        logger.debug("Updating TLS private key.")
-
-        if self.charm.tls_events.read_and_validate_private_key(private_key_id) is None:
-            self.charm.set_status(Status.TLS_INVALID_PRIVATE_KEY)
-            return
-
-        self.charm.tls_events.refresh_tls_certificates_event.emit()
 
     def _rebuild_cluster(self) -> None:  # noqa: C901
         """Rebuild cluster with new membership configuration, to recover from majority failure.
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -412,6 +412,9 @@ class EtcdEvents(Object):
 
     def _on_secret_changed(self, event: ops.SecretChangedEvent) -> None:
         """Handle the secret_changed event."""
+        if not self.charm.unit.is_leader():
+            return
+
         if (
             self.charm.state.cluster.is_restore_in_progress
             or self.charm.state.cluster.rebuild_cluster_in_progress
@@ -420,9 +423,6 @@ class EtcdEvents(Object):
                 "Cannot update credentials while a restore or cluster-rebuild operation is in progress."
             )
             event.defer()
-            return
-
-        if not self.charm.unit.is_leader():
             return
 
         if admin_secret_id := self.charm.config.get(INTERNAL_USER_PASSWORD_CONFIG):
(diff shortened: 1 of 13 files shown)


src/events/etcd.py
```diff
@@ -200,7 +198,7 @@ def _on_start(self, event: ops.StartEvent) -> None:  # noqa: C901
         else:
             self.charm.set_status(Status.SERVICE_NOT_RUNNING)
 
-    def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:  # noqa: C901
```
Mehdi-Bendriss, line 203: nice

src/events/etcd.py
```diff
@@ -441,6 +422,9 @@ def _on_secret_changed(self, event: ops.SecretChangedEvent) -> None:
             event.defer()
             return
 
+        if not self.charm.unit.is_leader():
```
Mehdi-Bendriss: does it not make sense to keep it as it was? so non-juju-leader units don't need to defer an event they don't have to handle anyway

src/managers/tls.py
```diff
@@ -268,30 +270,138 @@ def compute_component_status(self) -> list[Status]:
...
             frozenset[str]: The SANs IP.
         """
-        private_ip = self.workload.get_private_ip()
-        if not private_ip:
+        sans_ip = set()
+        if self.extra_sans_config_is_valid() and (
+            extra_sans_config := self.state.config.get("certificate-extra-sans")
+        ):
+            extra_sans = [san.strip() for san in extra_sans_config.split(",")]
+            sans_ip = {san for san in extra_sans if len(san.split(".")) == 4}
```
Mehdi-Bendriss: This will include all subdomains in the form of `sub.domain.example.com` as IPs. 
We should rather use a regex to extract valid IPv4 addresses. Here is one from [stackoverflow](https://stackoverflow.com/questions/5284147/validating-ipv4-addresses-with-regexp) (I tested it in various valid / invalid inputs) 
```
^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}$
```

src/managers/tls.py
```diff
@@ -268,30 +270,138 @@ def compute_component_status(self) -> list[Status]:
...
+
+        if private_ip := self.workload.get_private_ip():
+            sans_ip.add(private_ip)
+        else:
```
Mehdi-Bendriss: When would this happen? if it does, any reason to not have this fallback logic in `get_private_ip` ?

src/managers/tls.py
```diff
@@ -268,30 +270,138 @@ def compute_component_status(self) -> list[Status]:
...
+        """
+        sans_dns = set()
+        if self.extra_sans_config_is_valid() and (
+            extra_sans_config := self.state.config.get("certificate-extra-sans")
+        ):
+            extra_sans = [san.strip() for san in extra_sans_config.split(",")]
+            sans_dns = {
+                san.replace("{unit}", str(self.state.unit_server.unit_id))
+                for san in extra_sans
+                if len(san.split(".")) != 4
```
Mehdi-Bendriss: Similar comment on the accuracy of the matching, this could be `all_sans - sans_ip`

src/managers/tls.py
```diff
@@ -268,30 +270,138 @@ def compute_component_status(self) -> list[Status]:
...
+        """Validate configuration value for certificate-extra-sans option.
+
+        Returns:
+            bool: True if config value is valid, False if invalid.
+        """
+        if not (extra_sans_config := self.state.config.get("certificate-extra-sans")):
+            return True
+
+        extra_sans = [san.strip() for san in extra_sans_config.split(",")]
+        allowed = re.compile(r"(?!-)[A-Z0-9-{}]{1,63}(?<!-)$", re.IGNORECASE)
```
Mehdi-Bendriss: A few points here:

1. For clarity, could you add a comment about this regex, the 63 character limits etc. and rename the `allowed` variable to something along the lines of `allowed_domain_names` or similar

2. I think all the occurrences of the `{unit}` placeholder will count towards the `63` limit. Now realistically, this wouldn't be a frequent case we come across - but i believe we can safeguard around it by first evaluating the interpolated string - then applying the regex without the `{}` characters.

src/managers/tls.py
```diff
@@ -268,30 +270,138 @@ def compute_component_status(self) -> list[Status]:
...
+
+        for san in extra_sans:
+            # validation for ip addresses
+            if len(san.split(".")) == 4:
```
Mehdi-Bendriss: similar comment on the IPv4 matching pattern

src/managers/tls.py
```diff
@@ -268,30 +270,138 @@ def compute_component_status(self) -> list[Status]:
...
+                san_type, san_value = sans.split(":")
+
+                if san_type.strip() == "DNS":
+                    sans_dns.add(san_value)
+                if san_type.strip() == "IP Address":
+                    sans_ip.add(san_value)
+
+        return {"sans_ip": sans_ip, "sans_dns": sans_dns}
+
+    def certificate_sans_updated(self, tls_type: TLSType) -> bool:
```
Mehdi-Bendriss: nit: I believe this should be renamed to something implying whether the sans should be updated or not, perhaps? (my initial comprehension, before reading the method, was whether the sans were successfully updated in the certificates)

Mehdi-Bendriss (commented):
Thank you René, really nice work! I left a few comments on some edge cases / or refinement of validation

src/managers/tls.py
reneradoi: This logic has already been here before (see [TLSManager](https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/src/managers/tls.py#L279-L282)). I've only changed it to not return immediately in this case, because we want to add the `extra-sans` IP addresses.

src/events/etcd.py
reneradoi: True, it does make sense to keep this part as it was - will move it up.

src/managers/tls.py
reneradoi: Will rename it to `certificate_sans_require_update` to make it more clear what this is supposed to be.

src/managers/tls.py
reneradoi: I've replaced the "magic" number and regex, hostname validation is delegated to `validators.hostname` now.

src/managers/tls.py
reneradoi: Validation of IP addresses is now done via the `ipaddress` lib.

src/managers/tls.py
reneradoi: Validation of IP addresses is now done via the `ipaddress` lib.

src/managers/tls.py
reneradoi: Validation of IP addresses is now done via the `ipaddress` lib.

src/managers/tls.py
Mehdi-Bendriss: My point here was whether it would rather make more sense to always return `socket.gethostbyname(socket.gethostname())` in `self.workload.get_private_ip()` if the juju get ip doesn' yield anything (in a specific state that we're not aware of). 
Otherwise this will force to repeat this check everywhere in the code whenever we're looking for a private_ip

reneradoi: Ok got it. I've now refactored the `workload.get_private_ip()` to be a bit simpler and have that fallback, by that we also got rid of `socket` in the TLS manager.

Mehdi-Bendriss (approved):
Thank you René!

src/events/tls.py
```diff
@@ -319,6 +322,44 @@ def _on_clean_ca(self, event: CleanCAEvent) -> None:
             )
             event.defer()
 
+    def _on_config_changed(self, event: ConfigChangedEvent) -> None:
```
skourta, line 325: I really like this!

src/managers/tls.py
```diff
@@ -268,30 +269,150 @@ def compute_component_status(self) -> list[Status]:
...
+        return frozenset(sans_ip)
+
+    def build_sans_dns(self) -> frozenset[str]:
+        """Build the SANs DNS for the TLS certificate.
+
+        Returns:
+            frozenset[str]: The SANs DNS.
+        """
+        sans_dns = set()
+        if self.extra_sans_config_is_valid() and (
```
skourta, line 332: nit:
We should invert the order of the condition only to check if it's valid in case it exists. Small performance gain

src/managers/tls.py
```diff
@@ -268,30 +269,150 @@ def compute_component_status(self) -> list[Status]:
...
+        Returns:
+            bool: True if config value is valid, False if invalid.
+        """
+        if not (extra_sans_config := self.state.config.get("certificate-extra-sans")):
+            return True
+
+        extra_sans = [san.strip() for san in extra_sans_config.split(",")]
+
+        for san in extra_sans:
+            if not self._is_ip_address(san):
```
skourta, line 358: nit: group conditions

skourta (approved):

src/managers/tls.py
reneradoi: Agree, but it won't matter much here, as the validity check immediately returns if the config doesn't exist.

PR #133, opened 2025-08-05
[DPE-7831] - Fix integration test for CA rotation

skourta:
This pull request updates the CA and certificate rotation integration test to use longer validity periods and adjusts the waiting logic accordingly. 

## Context
- The cluster would register a full failure upon rotation CA certificates
- The certificate secret would expire at 90% of the validity of the certificate, leaving 10% of its lifetime to acquire a new one and do a rolling restart
- Empirically, a full rolling restart for CA rotation takes around 1m30s

## Problem
- The certificates were expiring before the units acquired the lock, resulting in the cluster members losing communications with the other members, which would cause a cluster failure

## Fix
* Increased the CA validity to 20 minutes and certificate validity to 10 minutes.
* Updated the waiting period after CA renewal to 12m to reflect the new certificate validity settings, and revised the log message to explain the timing.
* Changed `apps_full_statuses` to `units_full_statuses` in `wait_until` calls to wait for the idle_time. [[1]](diffhunk://#diff-a8362a8580830e574ae0f1e2fbd35adbf5ea51ed2ae3a85e78c5a2eebaa4d0c7L243-R250) [[2]](diffhunk://#diff-a8362a8580830e574ae0f1e2fbd35adbf5ea51ed2ae3a85e78c5a2eebaa4d0c7R282-R292)


src/events/etcd.py
```diff
@@ -404,6 +405,10 @@ def _on_update_status(self, event: ops.UpdateStatusEvent) -> None:
             not self.charm.state.cluster.cluster_state
             or self.charm.state.cluster.is_restore_in_progress
             or self.charm.state.cluster.rebuild_cluster_in_progress
+            or self.charm.state.unit_server.tls_client_ca_rotation_state
+            != TLSCARotationState.NO_ROTATION
+            or self.charm.state.unit_server.tls_peer_ca_rotation_state
+            != TLSCARotationState.NO_ROTATION
```
reneradoi, line 411: Good find!

tests/integration/tls/test_ca_rotation.py
```diff
@@ -240,14 +265,24 @@ async def test_ca_rotation_by_expiration(ops_test: OpsTest) -> None:
     # cert validity should be enough time for the rotation to happen
     # even with health checks failing because of invalid certs
     # CA validity must be 2x cert validity to be considered valid
-    logger.info("Adjusting validity of the CA to 12 min and certificates to 6 min")
-    tls_config = {"root-ca-validity": "12m", "certificate-validity": "6m"}
+    ca_root_validity = 20  # in minutes
+    certificate_validity = 10  # in minutes
+    logger.info(
+        "Adjusting validity of the CA to %s min and certificates to %s min",
```
reneradoi, line 242: How come you don't use f-strings anymore?

tests/integration/tls/test_ca_rotation.py
```diff
@@ -279,15 +314,24 @@ async def test_ca_rotation_by_expiration(ops_test: OpsTest) -> None:
...
+    logger.info("Waiting for the certificates to expire after CA renewal")
+    while (
+        remaining_time := (time_after_config_set + (certificate_validity * 60 * 0.6 * 2) + 30)
+        - time.time()
+    ) > 0:
+        logger.info(
+            "Waiting %.0f seconds for the certificates to expire after CA renewal",
+            remaining_time,
+        )
+        time.sleep(10)
```
reneradoi: Personally, that's a bit too verbose to me. If you want to have that information while the test is running, let's rather do something like:

```
waiting_time = (certificate_validity * 60 * 0.6 * 2) + 30
cert_expiration_time = time_after_config_set + waiting_time
logger.info(f"Certificates will expire at {cert_expiration_time}")
time.sleep(waiting_time)
```

That is much nicer to read and understand, and I'd prefer to avoid while-loops in tests wherever possible.

tests/integration/tls/test_ca_rotation.py
```diff
@@ -228,6 +228,31 @@ async def test_ca_rotation_by_config_change(ops_test: OpsTest) -> None:
...
 
+@pytest.mark.abort_on_fail
+async def test_prepare_units_for_ca_expiration_test(ops_test: OpsTest) -> None:
+    """Prepare the units for the CA expiration test."""
+    # add debug log to self signed certificates app
+    ssc_unit = ops_test.model.applications[TLS_NAME].units[0]
+    logger.info("Adding debug log to self-signed-certificates charm")
+    action = await ssc_unit.run(
+        'sudo sed -i \'/^            self._dump_provider_certificates(relation=relation, certificates=provider_certificates)/a\\            logger.debug("Revoked all certificates in relation %s", relation.id)\\n            logger.debug("Relation data %s", relation.data[self.charm.app])\' /var/lib/juju/agents/unit-self-signed-certificates-0/charm/lib/charms/tls_certificates_interface/v4/tls_certificates.py',
+        block=True,
```
reneradoi: Do you think we still need that after TLS v4 has been fixed?

tests/integration/tls/test_ca_rotation.py
```diff
@@ -228,6 +228,31 @@ async def test_ca_rotation_by_config_change(ops_test: OpsTest) -> None:
...
+    )
+    assert action.status == "completed", f"Failed to add debug log: {action}"
+
+    etcd_app: Application = ops_test.model.applications[APP_NAME]  # type: ignore
+    for unit in etcd_app.units:
```
reneradoi, line 214: Let's please move this to a helper. It impacts readability of the test a lot. The `run` command could also be formatted a bit better into something like:

```
search_expression = ...
replace_expression = ...
file = ...
command = sudo sed -i s/{search_expression}/{replace_expression}/{file}
```

reneradoi (changes requested):
Thanks Smail for investigating and fixing this! I'd like to see some formatting changes to the test, to account for better read-/understandability. Shouldn't we also update the TLS lib to include the fix that you contributed there?

tests/integration/tls/test_ca_rotation.py
skourta: As pointed out to me by Gulsum, using f-string while logging is a [code smell](https://docs.astral.sh/ruff/rules/logging-f-string/).

tests/integration/tls/test_ca_rotation.py
skourta: I added this so the users would have more visibility on how longer we will still wait progressively. When I was executing the test frequently I found it easier to follow the execution this way. 

I think since we should no longer need to frequently check on the test it makes sense to make the changes you proposed.

tests/integration/tls/test_ca_rotation.py
skourta: Yes this would make the test more stable. If we keep the default 90% then the certificates might actually expire before we finish the rolling restart. We can either make the validity durations longer or keep this one. WDYT?

tests/integration/tls/test_ca_rotation.py
reneradoi: Nice to know, didn't hear of that yet.

tests/integration/tls/test_ca_rotation.py
reneradoi: Sorry for being not precise enough. I agree on the 0.6 change, I was rather referring to the extensive debug logging that is added on the certificate provider side. Do we still need this logging output?

tests/integration/tls/test_ca_rotation.py
```diff
@@ -228,6 +208,21 @@ async def test_ca_rotation_by_config_change(ops_test: OpsTest) -> None:
     ), "Failed to read new key"
 
 
+async def _prepare_units_for_ca_expiration_test(ops_test: OpsTest) -> None:
+    """Prepare the units for the CA expiration test."""
+    etcd_app: Application = ops_test.model.applications[APP_NAME]  # type: ignore
+    for unit in etcd_app.units:
+        logger.info("Updating renewal relative time to 0.6 for unit %s", unit.name)
+        await unit.run(
+            f"sudo sed -i 's|\\(refresh_events=\\[self.refresh_tls_certificates_event\\],\\)|\\1renewal_relative_time=0.6,|' /var/lib/juju/agents/unit-{unit.name.replace('/', '-')}/charm/src/events/tls.py",
```
reneradoi: I would only like to have these run commands formatted a bit better (as mentioned above), into something like:

search_expression = ...
replace_expression = ...
file = ...
command = sudo sed -i s/{search_expression}/{replace_expression}/{file}

tests/integration/tls/test_ca_rotation.py
```diff
@@ -228,6 +208,21 @@ async def test_ca_rotation_by_config_change(ops_test: OpsTest) -> None:
...
+            block=True,
+        )
+        await unit.run(
+            f"sudo sed -i '/provider_relation_data = _ProviderApplicationData.load(relation.data\\[relation.app\\])/a\\            logger.debug(\"Remote relation data bag: %s\", relation.data[relation.app])' /var/lib/juju/agents/unit-{unit.name.replace('/', '-')}/charm/lib/charms/tls_certificates_interface/v4/tls_certificates.py",
```
reneradoi: Here too: formatting

reneradoi (commented):
Only the `run` commands that need some better formatting for readability, other than that I'm good.

reneradoi (approved):
Thank you for all the adjustments, Smail!

Mehdi-Bendriss (approved):
Thanks Smail, great work!

PR #138, opened 2025-08-08
[DPE-7889] Fix cluster-status computation

reneradoi:
**Issues**

With #124 the computation of cluster statuses has been reworked. This leads to lots of errors when deploying charmed etcd, because on every hook, the etcd-internal metrics server is queried for the `cluster_failed` metric, even though the etcd server has not been started yet.

Furthermore, if the cluster is not failed, the operator tries to delete the status, even though it wasn't set as a running status previously.

Another inconsistency is on enabling or disabling peer TLS, where `Waiting to join cluster` is displayed as a status message even though the unit is already a cluster member.

**Fixes**

- The check for `cluster_failed` should only be executed if the etcd server has already been started.
- This status does not have to be deleted as it is not a running status.
- Consider peer TLS transition for checking cluster membership status.

### DIFF
--- a/src/managers/cluster.py
+++ b/src/managers/cluster.py
@@ -384,6 +384,18 @@ class ClusterManager(ManagerStatusProtocol):
             if not self.state.cluster.auth_enabled:
                 status_list.append(ClusterStatuses.AUTHENTICATION_NOT_ENABLED.value)
 
+            try:
+                if self.is_cluster_failed:
+                    status_list.append(ClusterStatuses.CLUSTER_FAILED.value)
+                else:
+                    self.state.statuses.delete(
+                        ClusterStatuses.CLUSTER_FAILED.value,
+                        scope=scope,
+                        component=self.name,
+                    )
+            except RequestException as e:
+                logger.error(f"Could not determine if cluster failed: {e}")
+
         if not self.state.peer_relation:
             status_list.append(EtcdServiceStatuses.SERVICE_INSTALLING.value)
         else:
@@ -398,18 +410,6 @@ class ClusterManager(ManagerStatusProtocol):
 
             if self.state.cluster.learning_member and self.state.unit_server.is_juju_leader:
                 status_list.append(ClusterStatuses.CLUSTER_MEMBER_NOT_PROMOTED.value)
-
-            try:
-                if self.is_cluster_failed:
-                    status_list.append(ClusterStatuses.CLUSTER_FAILED.value)
-                else:
-                    self.state.statuses.delete(
-                        ClusterStatuses.CLUSTER_FAILED.value,
-                        scope=scope,
-                        component=self.name,
-                    )
-            except RequestException as e:
-                logger.error(f"Could not determine if cluster failed: {e}")
 
         return status_list if status_list else [CharmStatuses.ACTIVE_IDLE.value]
 
--- a/src/managers/cluster.py
+++ b/src/managers/cluster.py
@@ -396,7 +396,11 @@ class ClusterManager(ManagerStatusProtocol):
             if not self.state.cluster.cluster_state:
                 status_list.append(ClusterStatuses.CLUSTER_INITIALIZING.value)
 
-            if self.state.unit_server.member_endpoint not in self.state.cluster.cluster_members:
+            if (
+                self.state.unit_server.member_endpoint not in self.state.cluster.cluster_members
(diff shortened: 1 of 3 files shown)


skourta (dismissed):
Thanks @reneradoi .
To clarify by errors you mean the log error `Could not determine if cluster failed` not an error status, correct?

reneradoi:
> Thanks @reneradoi . To clarify by errors you mean the log error `Could not determine if cluster failed` not an error status, correct?

Yes, these are the ones:
- metrics call failed, for example [here](https://github.com/canonical/charmed-etcd-operator/actions/runs/16868738111/job/47789211405#step:13:911)
- status cannot be deleted because it is not present, for example [here](https://github.com/canonical/charmed-etcd-operator/actions/runs/16868738111/job/47789211405#step:13:1112)

reneradoi:
@skourta I've found another case where the status is not consistent: When enabling/disabling peer TLS, the message `waiting to join cluster` is displayed, for example [here](https://github.com/canonical/charmed-etcd-operator/actions/runs/16868738111/job/47789211457#step:8:2956).

I've pushed a fix for it, which considers if there is a  peer TLS transition when checking cluster membership. Could you please reveiw again?

src/managers/cluster.py
```diff
@@ -396,7 +396,11 @@ def get_statuses(self, scope: Scope, recompute: bool = False) -> list[StatusObje
             if not self.state.cluster.cluster_state:
                 status_list.append(ClusterStatuses.CLUSTER_INITIALIZING.value)
 
-            if self.state.unit_server.member_endpoint not in self.state.cluster.cluster_members:
+            if (
+                self.state.unit_server.member_endpoint not in self.state.cluster.cluster_members
+                and self.state.unit_server.tls_peer_state
+                not in [TLSState.TO_TLS, TLSState.TO_NO_TLS]
```
skourta: I did consider this upon implementation. I think from the charm's point of view, a unit is in the cluster if its member endpoint is in the `cluster_members`. I do not consider the previous behaviour to be inconsistent, but a mere reflection of the charm's process of switching TLS.

reneradoi: I understand. If you want to keep it displayed, maybe we should change the message to `Waiting to join or update cluster membership`?

reneradoi: I've thought about it again and I think we should not display it anymore. Reason being: It is only displayed because the cluster membership information in the app databag has not been updated yet by the leader. The cluster membership itself is already up-to-date at the time, because of the `cluster_manager.broadcast_peer_url()` right before the peer TLS enabling/disabling restart.

skourta: What do you think about checking that the IP of the unit is in the cluster_members instead of deactivating the check entirely?

src/managers/cluster.py
```diff
@@ -384,13 +384,23 @@ def get_statuses(self, scope: Scope, recompute: bool = False) -> list[StatusObje
...
                 status_list.append(ClusterStatuses.AUTHENTICATION_NOT_ENABLED.value)
 
+            try:
+                if self.is_cluster_failed:
+                    status_list.append(ClusterStatuses.CLUSTER_FAILED.value)
+            except RequestException as e:
+                logger.warning(f"Could not determine if cluster failed: {e}")
+
         if not self.state.peer_relation:
             status_list.append(EtcdServiceStatuses.SERVICE_INSTALLING.value)
```
Mehdi-Bendriss: Shouldn't this be at the very beginning of this method and returning if that's the case? the prior checks assume the peer relation is already established.

src/managers/cluster.py
Mehdi-Bendriss: I would be inclined to perhaps have a tailored Maintenance (?) status such as "Refreshing cluster membership information / metadata.".  Implying that everything is good, just that the a routine maintenance task is ongoing. 
What do you think?

Mehdi-Bendriss (commented):
Thanks René! I have a couple of small comments

src/managers/cluster.py
reneradoi: Yes, that is a good compromise. I've added that and reworked the condition to reflect that.

src/managers/cluster.py
reneradoi: updated

Mehdi-Bendriss (approved):
Thanks René!

skourta (approved):

PR #139, opened 2025-08-11
[DPE-7735] Upgrades implementation

reneradoi:
This PR adds minor version upgrades without downtime to charmed etcd.

**What the PR includes:**
- the implementation of all functionalities as described in the [charm-refresh v3 docs](https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/)
- a new event handler `MachinesEtcdRefresh` that implements the etcd specifics for upgrades
- a new class `UpgradesManager` whose only responsibility is to bridge the gap between `charm-refresh v3` and `advanced statuses` regarding status handling
- adjustments to `EtcdWorkload` to be able to install a specific snap revision (not just the latest) and to be able to handle the `refresh_versions` file from `charm-refresh` instead of literals for snap revision
- adjustments to critical event handlers to block or defer while upgrade is in progress (as discussed in the [spec](https://docs.google.com/document/d/1o5UsdfA-fX_aOireeC-aCE4AVj8GzMWCBpVVouYFCpI/edit?usp=sharing))
- unit test coverage for the specific parts of implementation (not for the `charm-refresh` functionality itself)

**What is not part of this PR:**
- basic integration test coverage
- edge case integration test coverage

For proper tests, we need to be able to install / upgrade from a charm version that supports `charm-refresh v3`. Therefore I will create separate PR's for integration testing, which will then include newer snap versions as well.


carlcsaposs-canonical (changes requested):
discussed with Rene on mattermost—it looks like the UX requirements for the statuses (https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/status/) are not being met in the current implementation

config.yaml
```diff
@@ -40,3 +40,10 @@ options:
       The extra-sans are specified by comma-separated names to be added when requesting signed certificates.
       Use "{unit}" as a placeholder to be filled with the unit number, e.g. "worker-{unit}" will be translated as 
       "worker-0" for unit 0 and "worker-1" for unit 1 when requesting the certificate.
+
+  pause-after-unit-refresh:
```
skourta, line 44: We do not need a spec for this right? It should be part of the refresh spec.

skourta, line 44: We do not need a spec for this right? It should be part of the refresh spec.

src/events/refresh.py
```diff
@@ -0,0 +1,183 @@
...
+        snap_name: str,
+        snap_revision: str,
+        refresh: charm_refresh.Machines,
+    ) -> None:
+        """Refresh the snap for the etcd charm."""
+        self.charm.cluster_manager.move_leader_if_required()
+        self.charm.workload.stop()
+
+        revision_before_refresh = self.charm.workload.snap_revision()
+        assert snap_revision != revision_before_refresh
```
skourta: We should add an assertion message to explain

src/events/refresh.py
```diff
@@ -0,0 +1,183 @@
...
+        logger.info("Restarting workload")
+        # always apply the current charm revision's config -> no need to "migrate" configuration
+        # this charm revision's config is the one supported by the targeted workload version
+        self.charm.config_manager.set_config_properties()
+        self.charm.workload.start()
+        if self.charm.cluster_manager.is_healthy():
+            refresh.next_unit_allowed_to_refresh = True
+        else:
+            self.charm.status.set_running_status(
+                ClusterStatuses.HEALTH_CHECK_FAILED.value,
```
skourta, line 93: The status needs to be updated to be a running status.

src/charm.py
```diff
@@ -88,6 +115,37 @@ def __init__(self, *args):
             ],
         )
 
+        if self.refresh and not self.refresh.next_unit_allowed_to_refresh:
+            self._post_snap_refresh()
```
skourta, line 124: This is for when the health check at the end of `refresh_snap` fails, right?

src/managers/upgrades.py
```diff
@@ -0,0 +1,117 @@
...
+
+        Args:
+            ops_status (ops.StatusBase): the status to convert into an advanced status
+            critical (bool): whether the returned StatusObject should have the field
+                            `approved_critical_component` set to True or False
+        """
+        # this code may not be very concise, focus is on readability
+        if isinstance(ops_status, BlockedStatus):
+            return StatusObject(
+                status="blocked", message=ops_status.message, approved_critical_component=critical
```
skourta: We will need Mehdi's explicit approval for this based on the spec for advanced statuses.

src/managers/upgrades.py
```diff
@@ -0,0 +1,117 @@
...
+    ) -> StatusObject:
+        """Convert an ops status into an advanced statuses StatusObject.
+
+        Args:
+            ops_status (ops.StatusBase): the status to convert into an advanced status
+            critical (bool): whether the returned StatusObject should have the field
+                            `approved_critical_component` set to True or False
+        """
+        # this code may not be very concise, focus is on readability
+        if isinstance(ops_status, BlockedStatus):
```
skourta: nit: consider a match case

tests/unit/test_upgrades.py
```diff
@@ -0,0 +1,304 @@
...
+    # check fails because backup in progress
+    peer_relation = testing.PeerRelation(
+        id=1,
+        endpoint=PEER_RELATION,
+        local_app_data={"backup_id": "XYZ"},
+    )
+
+    state_in = testing.State(relations={peer_relation})
+
+    with ctx(ctx.on.relation_changed(relation=peer_relation), state_in) as manager:
```
skourta, line 55: we might be able to refactor all of these to use soemthing like [TableDrivenTests](https://go.dev/wiki/TableDrivenTests)

skourta (commented):
Thank you @reneradoi. I left some questions.

config.yaml
reneradoi: No need to do this per product, as UI/UX (actions, config options, etc.) have been approved in DA110 "In-place refreshes v3" for all our team.

src/charm.py
reneradoi: Correct. It is placed into the `__init__` to make sure this runs before every other event. For further information, see https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/next-unit-allowed-to-refresh/#machines.

src/managers/upgrades.py
reneradoi: For my understanding, this has been approved for `upgrade` statuses already as part of DA147 "UX for advanced statuses", explicit approval only required if other components than upgrades need that.

@Mehdi-Bendriss could you please confirm if upgrades has been "pre-approved" as critical component for overruling all other statuses?

src/charm.py
skourta: Yes, I've read that, thank you.

src/managers/upgrades.py
reneradoi: Thanks, good idea!

src/events/refresh.py
reneradoi: updated

src/events/refresh.py
reneradoi: updated

tests/unit/test_upgrades.py
reneradoi: Good idea, increases readability too!

src/managers/upgrades.py
Mehdi-Bendriss: Yes, so far, only the high priority refresh statuses are approved as "critical components"

src/charm.py
```diff
@@ -88,6 +115,37 @@ def __init__(self, *args):
...
+    def refresh_not_ready(self) -> bool:
+        """Check if charm-refresh is not available or currently in progress."""
+        return not self.refresh or self.refresh.in_progress
+
+    def _post_snap_refresh(self) -> None:
+        """Handle post-snap refresh health checks and set next_unit_allowed_to_refresh."""
+        if not self.refresh.in_progress:
+            self.refresh.next_unit_allowed_to_refresh = True
+            self.state.statuses.delete(
+                EtcdServiceStatuses.SERVICE_NOT_RUNNING.value,
```
Mehdi-Bendriss: Can you explain this? This status is not exactly set exclusively for refreshes no?

src/events/external_clients.py
```diff
@@ -60,6 +60,17 @@ def __init__(self, charm: "EtcdOperatorCharm"):
 
     def _on_mtls_cert_updated(self, event: MTLSCertUpdatedEvent) -> None:  # noqa: C901
         """Handle the ca chain updated event."""
+        if (
+            self.charm.state.cluster.is_restore_in_progress
+            or self.charm.state.cluster.rebuild_cluster_in_progress
```
Mehdi-Bendriss, line 65: nice

src/events/refresh.py
```diff
@@ -0,0 +1,185 @@
...
+            self.charm.state.unit_server.tls_client_state in tls_transition_states
+            or self.charm.state.unit_server.tls_peer_state in tls_transition_states
+        ):
+            raise charm_refresh.PrecheckFailed("TLS transition is in progress")
+
+        if not self.charm.cluster_manager.is_healthy():
+            raise charm_refresh.PrecheckFailed("Cluster is not healthy")
+
+
+def is_workload_compatible(
```
Mehdi-Bendriss: any reason why this is not a staticmethod as part of `MachinesEtcdRefresh`?

src/managers/upgrades.py
```diff
@@ -0,0 +1,122 @@
...
+        status at all. We achieve this by having `upgrades_manager` last in the component priority
+        order, but setting the field `approved_critical_component` to True if the refresh_status is
+        of higher priority than any other status.
+
+        This logic ignores all statuses set directly by the refresh lib, as of refresh v3.1.0.
+        For more information: see https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/status/
+        """
+        status_list: list[StatusObject] = []
+
+        if self.refresh is None:
```
Mehdi-Bendriss: nit, 
```suggestion
        if not self.refresh:
```

src/managers/upgrades.py
```diff
@@ -0,0 +1,122 @@
...
+            return [CharmStatuses.ACTIVE_IDLE.value]
+
+        if (refresh_app_status := self.refresh.app_status_higher_priority) is not None:
```
Mehdi-Bendriss: nit (and in subsequent conditions) 
```suggestion
        if refresh_app_status := self.refresh.app_status_higher_priority:
```

Mehdi-Bendriss (commented):
Thank you René, I don't have concerns only a few nits and a couple of questions

src/managers/upgrades.py
reneradoi: updated

src/managers/upgrades.py
reneradoi: updated

src/charm.py
reneradoi: Good find, Thank you! This has been a leftover. I had adjusted the running status set in `refresh_snap` from `EtcdServiceStatuses.SERVICE_NOT_RUNNING` to `ClusterStatuses.HEALTH_CHECK_FAILED` but did not update it here.

src/events/refresh.py
reneradoi: No specific reason, has been moved.

Mehdi-Bendriss (dismissed):
Thanks René!

skourta (dismissed):
Thanks!

actions.yaml
```diff
@@ -33,4 +33,52 @@ status-detail:
...
+      description: |
+        Potential of data loss and downtime
+        If `false`, force refresh if app is unhealthy or not ready to refresh (and unit status shows "Pre-refresh check failed")
+    check-workload-container:
+      type: boolean
+      default: true
+      description: | 
+        Potential of data loss and downtime during and after refresh
+        If `false`, allow refresh to etcd container version that has not been validated to work with the charm revision
+  additionalProperties: false
```
carlcsaposs-canonical:
```suggestion
force-refresh-start:
  description: |
    Potential of data loss and downtime

    Force refresh of first unit

    Must run with at least one of the parameters `=false`
  params:
    check-compatibility:
      type: boolean
      default: true
      description: | 
        Potential of data loss and downtime

        If `false`, force refresh if new version of etcd and/or charm is not compatible with previous version
    run-pre-refresh-checks:
      type: boolean
      default: true
      description: |
        Potential of data loss and downtime

        If `false`, force refresh if app is unhealthy or not ready to refresh (and unit status shows "Pre-refresh check failed")
    check-workload-container:
      type: boolean
      default: true
      description: | 
        Potential of data loss and downtime during and after refresh

        If `false`, allow refresh to etcd container version that has not been validated to work with the charm revision
  additionalProperties: false
```

actions.yaml
```diff
@@ -33,4 +33,52 @@ status-detail:
...
+  params:
+    check-health-of-refreshed-units:
+      type: boolean
+      default: true
+      description: |
+        Potential of data loss and downtime
+        If `false`, force refresh (of next unit) if 1 or more refreshed units are unhealthy
+        Warning: if first unit to refresh is unhealthy, consider running `force-refresh-start` action on that unit instead of using this parameter.
+        If first unit to refresh is unhealthy because compatibility checks, pre-refresh checks, or workload container checks are failing, this parameter is more destructive than the `force-refresh-start` action.
+  additionalProperties: false
```
carlcsaposs-canonical:
```suggestion
resume-refresh:
  description: |
    Refresh next unit(s) (after you have manually verified that refreshed units are healthy)

    If the `pause-after-unit-refresh` config is set to `all`, this action will refresh the next unit.

    If `pause-after-unit-refresh` is set to `first`, this action will refresh all remaining units.
    Exception: if automatic health checks fail after a unit has refreshed, the refresh will pause.

    If `pause-after-unit-refresh` is set to `none`, this action will have no effect unless it is called with `check-health-of-refreshed-units` as `false`.
  params:
    check-health-of-refreshed-units:
      type: boolean
      default: true
      description: |
        Potential of data loss and downtime

        If `false`, force refresh (of next unit) if 1 or more refreshed units are unhealthy

        Warning: if first unit to refresh is unhealthy, consider running `force-refresh-start` action on that unit instead of using this parameter.
        If first unit to refresh is unhealthy because compatibility checks, pre-refresh checks, or workload container checks are failing, this parameter is more destructive than the `force-refresh-start` action.
  additionalProperties: false
```

config.yaml
```diff
@@ -40,3 +40,10 @@ options:
...
+    type: string
+    default: first
+    description: |
+      Wait for manual confirmation to resume refresh after these units refresh
+      Allowed values: "all", "first", "none"
```
carlcsaposs-canonical:
```suggestion
options:
  pause-after-unit-refresh:
    description: |
      Wait for manual confirmation to resume refresh after these units refresh

      Allowed values: "all", "first", "none"
    type: string
    default: first
```

src/charm.py
```diff
@@ -55,13 +68,27 @@ def __init__(self, *args):
             self.state, self.workload, SUBSTRATE
         )
 
+        # --- UPGRADES ---
+        try:
+            self.refresh = charm_refresh.Machines(
+                MachinesEtcdRefresh(workload_name="etcd", charm_name="charmed-etcd", charm=self)
```
carlcsaposs-canonical, line 74:
```suggestion
                MachinesEtcdRefresh(workload_name="etcd", charm_name="charmed-etcd", _charm=self)
```
nit; to avoid potential clashes with future args in the public API

src/core/workload.py
```diff
@@ -109,6 +109,18 @@ def load_yaml_file(self, file: str) -> Dict[str, Any]:
...
+    def load_toml_file(self, file: str) -> Dict[str, Any]:
+        """Read toml content from a file.
+
+        Args:
+            file (str): Path to the file.
+
+        Returns:
+            The content of a TOML file as a dict.
+        """
+        pass
```
carlcsaposs-canonical, line 122: if possible, please use charm-refresh's API instead of reading from refresh_versions.toml
https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/remove-duplicate-hardcoded-snap/

src/events/etcd.py
```diff
@@ -232,9 +232,10 @@ def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
...
             self.charm.state.cluster.is_restore_in_progress
             or self.charm.state.cluster.rebuild_cluster_in_progress
+            or self.charm.refresh_not_ready
         ):
             logger.warning(
-                "Cannot update config while a restore or cluster-rebuild operation is in progress."
+                "Cannot update config while cluster is in vulnerable state because of restore, upgrade or cluster-rebuild"
             )
             event.defer()
             return
```
carlcsaposs-canonical: are IP address changes handled while a refresh is in progress?

src/events/backup.py
```diff
@@ -430,6 +430,9 @@ def _exists_preventing_reason(self) -> str:
         if not self.charm.unit.is_leader():
             return "Action must be performed on the leader unit."
 
+        if self.charm.refresh_not_ready:
+            return "Upgrade in progress, cannot perform action."
```
carlcsaposs-canonical: should "refresh" be used instead of "upgrade" for consistency with the UX and our other charms?

(this & all other instances)

src/events/external_clients.py
```diff
@@ -157,6 +168,17 @@ def _on_mtls_cert_updated(self, event: MTLSCertUpdatedEvent) -> None:  # noqa: C
...
+            or self.charm.refresh_not_ready
+        ):
+            logger.warning(
+                "Cannot remove client relation while cluster is in vulnerable state because of restore, upgrade or cluster-rebuild"
+            )
+            event.defer()
+            return
```
carlcsaposs-canonical, line 180: fyi relation broken can't be successfully deferred—the context in the relation broken event will be lost since Juju is not aware of defer

(e.g. any cleanup that relies on data in the local databags of the relation will not have access to that data)

src/events/external_clients.py
```diff
@@ -172,6 +194,17 @@ def _on_relation_broken(self, event: RelationBrokenEvent) -> None:
...
+        if (
+            self.charm.state.cluster.is_restore_in_progress
+            or self.charm.state.cluster.rebuild_cluster_in_progress
+            or self.charm.refresh_not_ready
+        ):
+            logger.warning(
+                "Cannot update certificates while cluster is in vulnerable state because of restore, upgrade or cluster-rebuild"
+            )
+            event.defer()
+            return
```
carlcsaposs-canonical: if the old certificate is about to expire while a refresh is in progress, is that handled?

refreshes may be in progress for several weeks at a time

src/events/refresh.py
```diff
@@ -0,0 +1,185 @@
...
+        try:
+            old_major, old_minor, old_patch, *_ = (
+                int(component) for component in old_workload_version.split(".")
+            )
+            new_major, new_minor, new_patch, *_ = (
+                int(component) for component in new_workload_version.split(".")
+            )
+        except ValueError:
+            # Not enough values to unpack or cannot convert
+            logger.info(
```
carlcsaposs-canonical:
```suggestion
            logger.error(
```
suggest error instead of info since this situation is not expected to happen (i.e. it's not an expected incompatible refresh—it's that checking if the refresh was compatible failed)

or letting exception raise without catching

src/events/refresh.py
```diff
@@ -0,0 +1,185 @@
...
+        if new_minor != old_minor:
+            # Minor version upgrades are treated like major version upgrades in etcd (separate track).
+            # Technically it is possible to allow minor version upgrades from a different track.
+            # Restriction: base operating system must be kept the same
+            # Recommendation from charm-refresh: only allow upgrades from a different track for a
+            # specific set of workload versions
+            # Once we move to a new track, this part should be adjusted to allow upgrade to this track.
+            # The condition should then be:
+            # if not (new_minor == old_minor)
+            # or (new_minor == old_minor + 1 and old_patch == specific_patch_version)
```
carlcsaposs-canonical: Would recommend checking specific versions on both ends for refreshes between tracks (instead of just on the version you're refreshing from)

src/events/refresh.py
```diff
@@ -0,0 +1,185 @@
...
+        self.charm.cluster_manager.move_leader_if_required()
+        self.charm.workload.stop()
+
+        revision_before_refresh = self.charm.workload.snap_revision()
+        assert snap_revision != revision_before_refresh, (
+            "current snap revision and target revision are equal"
+        )
+
+        logger.info("Updating snap installation")
+        if not self.charm.workload.install(revision=snap_revision):
```
carlcsaposs-canonical: snap refresh must not be retried: https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/refresh-snap/

src/literals.py
```diff
@@ -24,10 +23,12 @@
 DATABASE_DIR = "/var/snap/charmed-etcd/common/var/lib/etcd/member"
 BACKUP_FILE_NAME = "/var/snap/charmed-etcd/common/archive/charmed-etcd_snapshot.db"
 BACKUP_ID_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
+VERSIONS_FILE = "refresh_versions.toml"
```
carlcsaposs-canonical, line 26: if possible, please use charm-refresh's API instead of reading from refresh_versions.toml
https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/remove-duplicate-hardcoded-snap/

src/literals.py
```diff
@@ -24,10 +23,12 @@
...
 
 DATA_STORAGE = "data"
 PEER_RELATION = "etcd-peers"
 STATUS_PEERS_RELATION = "status-peers"
+REFRESH_PEERS_RELATION = "refresh-v-three"
```
carlcsaposs-canonical: the refresh peer relation is not part of the public interface & the charm code must never read/write from it https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/public-interface/

src/workload.py
```diff
@@ -40,16 +43,22 @@ def start(self) -> None:
...
+        Args:
+            revision (str | None): the snap revision to install. Will be loaded from the
+                `refresh_versions.toml` file if None.
+
         Returns:
             True if successfully installed, False if any error occurs.
         """
+        if not revision:
+            versions = self.load_toml_file(f"{WORKING_DIR}/../{VERSIONS_FILE}")
+            revision = versions["snap"]["revisions"][machine()]
```
carlcsaposs-canonical, line 66: if possible, please use charm-refresh's API instead of reading from refresh_versions.toml
https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/remove-duplicate-hardcoded-snap/

carlcsaposs-canonical (changes requested):
still have some concerns about the status implementation—will follow up tomorrow with details after I do some testing

wanted to share other feedback now so you have it quicker

src/events/refresh.py
reneradoi: retry mechanism was removed

@Mehdi-Bendriss: fyi

src/events/refresh.py
reneradoi: updated to error

src/charm.py
reneradoi: Understand, but will keep it for now.

src/events/refresh.py
carlcsaposs-canonical: fyi it's fine to keep for install case, just not okay to retry in `refresh_snap` call

actions.yaml
reneradoi: updated

actions.yaml
reneradoi: updated

config.yaml
reneradoi: updated

src/events/backup.py
reneradoi: Would prefer to keep "upgrade" here. Reasoning: The term "refresh" is very specific for the Juju, but outside of that context, "upgrade" is much more common. We refer to the feature as "upgrades without downtime", and we call it "upgrades" in our user docs, too. Wherever it's not specificly tied to the cli command `juju refresh`, I prefer the term "upgrades".

src/events/external_clients.py
reneradoi: Thanks for the hint! We will have to address this separately, as it would be out of the context of this PR. I will create an issue for this.

src/events/backup.py
carlcsaposs-canonical: fyi, we used "upgrade" in the UX initially & it was requested to change
see https://docs.google.com/document/d/1DYtedzWlWgknGE6nJnmKUf8tp1PBuQbCd5x-aH93VuQ/edit?tab=t.0#heading=h.v228s8hnqd53

I think the expectation (with that change to the refresh v3 UX) was that we'd be consistent in the UX—and to use "refresh" instead of "upgrade" in the UX

regarding the user docs, I believe we're moving towards using a combination of "refresh" and "upgrade", but mostly "refresh" to be consistent with the UX— @a-velasco might have better info here (at least in regards to postgresql)

src/events/external_clients.py
reneradoi: In this part of the code, only mTLS certs for external clients are handled. It does not impact the functionality of the charm if that cert expires. It would induct a rolling restore though, which we would like to avoid while upgrading.

It's up to the admin / user then to decide between upgrades and updating the client certificate.

carlcsaposs-canonical: > only mTLS certs for external clients are handled. It does not impact the functionality of the charm if that cert expires

to check if I'm understanding: if that cert expires, will read/write traffic for a client be interrupted?

reneradoi: Yes, but we do not manage these certs for the clients.

carlcsaposs-canonical: oh—they are manually managed by the user?

src/events/etcd.py
reneradoi: Good point, I missed that here. Have adjusted the condition here and moved the part for refreshes to after the ip changes.

src/core/workload.py
reneradoi: This would add a dependency from the workload to the charm here, which would be conflicting with the charm-structure-pattern we apply. Therefore I fell back to reading from the file where required.

src/workload.py
reneradoi: Same here: This would add a dependency from the workload to the charm here, which would be conflicting with the charm-structure-pattern we apply. Therefore I fell back to reading from the file where required.

We have two calls for `install`: the initial installation and the `refresh_snap` call. Only in the initial installation we read from this file.

src/literals.py
reneradoi: Same here: This would add a dependency from the workload to the charm here, which would be conflicting with the charm-structure-pattern we apply. Therefore I fell back to reading from the file where required.

src/literals.py
reneradoi: This was a leftover. I thought I'd need it for unit testing, but didn't. Have removed it now.

src/events/refresh.py
reneradoi: That is not even needed, as we raise anyway if install fails. The retry mechanism is outsourced to Juju.

src/events/external_clients.py
reneradoi: No, by the client application.

Edit: After reading again, I understand what you mean: Yes, if the client's mTLS cert expires, they cannot read/write in etcd anymore. I am aware that this could lead to issues when users have a long-running upgrade and the cert expires because they didn't update it early enough.

Updating the client's mTLS certs requires rolling restarts in etcd. This is a significant risk for cluster availability if initiated during an upgrade, and to mitigate this risk, I consider it better to rather block the cert-update until the upgrade is completed, at least for now. This will be one of the edge case scenarios covered in our testing phase, and implementation especially for certificate rotation might be updated then to allow the certs update. But this will only come after this PR was merged and we implement integration tests.

src/events/external_clients.py
reneradoi: I've checked here again: We don't access relation data for the removed relation, so deferring is not an issue for us here.

src/events/external_clients.py
carlcsaposs-canonical: > they cannot read/write in etcd anymore

I think this is an HA concern

> Maintain high availability while refresh is in progress (for up to multiple weeks)

https://canonical-charm-refresh.readthedocs-hosted.com/latest/product-requirements/

Ideally, if possible, the cert update could be blocked unless it's about to expire

I thought this was the approach planned in [DA193 - minor version upgrades in etcd](https://docs.google.com/document/d/1o5UsdfA-fX_aOireeC-aCE4AVj8GzMWCBpVVouYFCpI/edit)

> defer any events related to TLS or external client relations that are initiating rolling restarts, such as switching TLS mode or modifying CA certificates. Rotation of non-CA certificates can be processed if the certificate expires soon, because it does not initiate a restart.

either way, would check with engineering & product managers about this tradeoff—since, as I currently understand, this does not meet the product requirements for in-place upgrades

src/events/refresh.py
Mehdi-Bendriss: i would say we need to add back the retry for install, we've had multiple cases where install fails the first time due to a connection error or a temporary store issue, and an immediate retry solves it transparently to the user. 
Error'ing the hook from the first time is not very user friendly

src/charm.py
```diff
@@ -55,13 +68,27 @@ def __init__(self, *args):
...
+        # --- STATUS HANDLER ---
         self.status = StatusHandler(  # priority order
             self,
             self.cluster_manager,
             self.config_manager,
             self.tls_manager,
             self.external_clients_manager,
             self.backup_manager,
+            self.upgrades_manager,
         )
```
carlcsaposs-canonical, line 97: Refresh statuses need to be the highest priority critical status—i.e. if there are other critical statuses, refresh must be listed first so that its statuses do not get overridden

Please add a comment with this information in case another critical status is added in the future

Or maybe separate the upgrade statuses into two components—one for the higher priority & one for the lower priority, and list the higher priority one as the first component

src/managers/upgrades.py
```diff
@@ -0,0 +1,120 @@
...
+
+        if not self.refresh:
+            return [CharmStatuses.ACTIVE_IDLE.value]
+
+        if refresh_app_status := self.refresh.app_status_higher_priority:
+            try:
+                app_status = self._convert_ops_status_to_advanced_status(refresh_app_status)
+                status_list.append(app_status)
+            except ValueError as e:
+                logger.error(e)
```
carlcsaposs-canonical:
```suggestion
        if refresh_app_status := self.refresh.app_status_higher_priority:
            app_status = self._convert_ops_status_to_advanced_status(refresh_app_status)
            status_list.append(app_status)
```

In the current implementation, if the conversion fails, the higher priority refresh status will be overridden by another status

If the conversion fails, another status should not be set

IMO, it would be easiest to let the exception get raised without catching it. I believe it is safer (less likely to cause downtime and data loss) to let an uncaught exception get raised than it is to override a higher priority refresh status

Typically if there's an uncaught exception here I imagine it would get caught in testing before release, but even if this happens in production I believe it's safer to raise an uncaught exception than to override a higher priority refresh status

(same thing for `refresh.unit_status_higher_priority`)

src/charm.py
```diff
@@ -119,9 +119,14 @@ def __init__(self, *args):
...
-        """Check if charm-refresh is not available or currently in progress."""
-        return not self.refresh or self.refresh.in_progress
+    def refresh_in_progress(self) -> bool:
+        """Check if charm-refresh is currently in progress."""
+        try:
+            return self.refresh.in_progress
+        except AttributeError:
+            # if charm_refresh.UnitTearingDown or charm_refresh.PeerRelationNotReady, this raises
+            # we consider a refresh to NOT be in progress
+            return False
```
carlcsaposs-canonical: (not related to status)

this is a huge change
@Mehdi-Bendriss @skourta fyi when you're re-reviewing

I think this breaks scale up while refresh is in progress

> If PeerRelationNotReady is raised on a unit that is added during scale up, the charm code:
> - Must not install the workload
> - Must not start the workload
> - Must not run any operation that would be dangerous to run while a refresh is in progress

https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/peer-relation-not-ready/

From DA193, it sounds like you intend to support scale up while refresh is in progress so that recovery from some refresh failure situations is possible. It looks like the current implementation might break if there is scale up while a refresh is in progress

carlcsaposs-canonical (changes requested):
the higher priority refresh statuses:
- inform the user that a refresh is in progress
- instruct the user how to pause and/or rollback the refresh if necessary
- inform the user if automatic checks failed
- tell the user what action they should perform next

when a refresh is in progress, it is critical that the user follows the instructions in the higher priority refresh statues to decide what to do next.

if they perform other actions (because they don't know a refresh is in progress or because a status message is distracting them with something else that is not as urgent), that is likely to lead to downtime and/or data loss

more info about the statuses: https://canonical-charm-refresh.readthedocs-hosted.com/latest/user-experience/status/
and the requirements for displaying refresh statuses to avoid downtime and data loss: https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/status/

---

with that in mind, it is essential that the higher priority refresh statuses are never overridden

the current status handler lib appears to override higher priority refresh statues for "action running statuses": https://github.com/canonical/data-platform-helpers/blob/e224d04d67480b8c33336d3f82c5cc4f374dbced/data_platform_helpers/advanced_statuses/handler.py#L110-L112

This contradicts what we agreed on in [DA147 - UX of Statuses](https://docs.google.com/document/d/1SV11ct-flQkc5BOYOeXgmPeglL8bVs-mDVkGjG20K48/edit?tab=t.0), which states

> In some cases where there is a critical status i.e. upgrade related statuses. We permit the status to _overwrite the entire_ status and use the full 120 characters available.

(I recognize that charmed-etcd doesn't currently use action running statues; but it must never use them in the future or the lib needs to be changed to not override the higher priority refresh statuses)

src/events/refresh.py
reneradoi: Re-added back the retry. This is only retrying `install`, so it is transparent to `charm-refresh`, and `refresh_snap` itself is never retried.

src/charm.py
reneradoi: I've added an extensive comment ahead of the instantiation of the `StatusHandler`. The same comment is present in the `get_statuses` method of the upgrades manager.

src/managers/upgrades.py
reneradoi: I've updated this method and removed the error handling. If the status returned from `refresh.app_status_higher_priority | refresh.unit_status_higher_priority | refresh.unit_status_lower_priority` is not a valid ops status, this will raise.

If - for whatever reason - it doesn't raise a `ValueError` in `charm-refresh`, it will raise an `AttributeError` in `_convert_ops_status_to_advanced_status` when we want to access the `ops_status.message`. There is also unit test coverage for this case, to ensure we do not override with other statuses in error cases.

src/charm.py
reneradoi: I have experienced that deferring all the events from the installation phase (especially install, leader_elected, config_changed) until all the peer relations are ready significantly deteriorates the setup phase (up to 2x the time, depending on the machine's resources). This is very relevant to every single deployment or scale-up operation. I don't think it is reasonable to accept that, given that the setup phase is actually not scope of the refresh component, except for scale-up operations during refreshes.

This part will definitely be scope of testing, but I am heavily favoring keeping this for now. If it breaks scale-up while refresh is in progress, I will adjust the logic and find a solution for it.

src/events/backup.py
reneradoi: Confirmed with @a-velasco that it is ok to use "upgrade" in this context. This does not impact the actions or configuration of `charm-refresh` themselves, only the message that is displayed when users run other actions such as `create-backup` or `restore`.

src/events/refresh.py
carlcsaposs-canonical: am I missing something? it looks like `refresh_snap` calls the `install` method that the retry was added back to

carlcsaposs-canonical: to clarify, within the refresh_snap method, the snap install must not be retried

refresh_snap should never be called by the charm code or retried by the charm code, that is the responsibility of charm-refresh—but that was not what I was referring to with my initial comment; was referring to the snap refresh must not be retried within a single refresh_snap call

carlcsaposs-canonical: (reasoning for this is to prevent delay in rollback; if user starts a rollback we should respond immediately and stop trying to refresh other units immediately)

src/events/external_clients.py
reneradoi: This does not impact the HA availability of etcd itself. It is only relevant if users / client applications decide to rotate their certs while running an upgrade. I believe it is reasonable to expect from users to ensure themselves they don't upgrade while another critical operation is in progress and vice versa.

src/charm.py
carlcsaposs-canonical: nit: suggest adding to the comment that
- this approach needs to be changed if any other critical status is added
or
- any other critical status needs to be added behind refresh statuses

src/events/backup.py
carlcsaposs-canonical: regarding approval for UX, I think you would need to check with Jon/Mohamed—but would ask that you check with Mehdi/Mykola first since they have context about the change from "upgrade" to "refresh"

I only mentioned Andreia in regards to terminology used in the docs; not in regards to who decided the UX change of "refresh" instead of "upgrade"

src/events/external_clients.py
carlcsaposs-canonical: > It is only relevant if users / client applications decide to rotate their certs while running an upgrade

is it not relevant if the users took no action in regards to cert rotation and the certs were close (a few days to a week) from expiring? (i.e. if a user does nothing, they would still have an issue since their certs would expire, correct?)

I wouldn't characterize loss of read/write traffic as "not impacting HA"—just because the cluster is still functioning, if read/write is impacted I think that counts as impacting HA, at least in the context of my understanding of what is required/expected by the product requirements for refreshes

src/charm.py
carlcsaposs-canonical: > significantly deteriorates the setup phase

fyi, this is not an inherent limitation of charm-refresh but is caused by the architecture of the charm/reliance on specific events

there are a few options:
- declare scale up as not supported while refresh in progress (this will impact ability to recover from some disaster scenarios). See "Alternative: Declare scaling up as not supported" on https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/peer-relation-not-ready/
- handle PeerRelationNotReady differently in initial deployment case vs scale up case—from my experience with postgresql, I think this would add significant complexity and it may be easier to modify the charm to be less reliant on specific events instead, but it depends on the charm code and the tradeoff may be different for the etcd charm

src/events/refresh.py
reneradoi: Yes, referring to Mehdi's previous comment: "We need to add back the retry for install, we've had multiple cases where install fails the first time due to a connection error or a temporary store issue, and an immediate retry solves it transparently to the user." I consider this to be also valid for calling `install` from `refresh_snap`. Could you please confirm @Mehdi-Bendriss?

src/events/external_clients.py
reneradoi: It was already agreed on in the spec:

```
While upgrade in progress:
[...]
- defer any events related to TLS or external client relations that are initiating rolling restarts
```

src/charm.py
reneradoi: Thank you for these recommendations. I understand your concern, and if scale-up breaks when testing, this will be subject of change. I will probably go with differentiating cases for initial setup and scale-up then. Declaring it as not supported would only be the last option.

Btw. I checked in Postgresql, I don't see that the snap installation waits for `charm-refresh` to be ready ([link](https://github.com/canonical/postgresql-operator/blob/16/edge/src/charm.py#L2195-L2196)). There is a `todo`, but currently it just installs `if not snap_package.present`. Does scale-up while refresh is in progress break in Postgres?

src/charm.py
reneradoi: updated

src/events/refresh.py
carlcsaposs-canonical: The snap install should not be retried in refresh_snap—charm-refresh takes care of the retry

Again, this is to enable rollback to start quickly

src/events/external_clients.py
carlcsaposs-canonical: > such as switching TLS mode or modifying CA certificates. Rotation of non-CA certificates can be processed if the certificate expires soon, because it does not initiate a restart

is rotation of non-ca certs being handled? that was what I was initially trying to bring up

src/charm.py
carlcsaposs-canonical: > and if scale-up breaks when testing

if the charm does not handle `PeerRelationNotReady` during scale-up and follow these requirements
> - Must not install the workload
> - Must not start the workload
> - Must not run any operation that would be dangerous to run while a refresh is in progress

then scale up should be declared as not supported, regardless of what the result of your testing is—if these requirements are not met, there are situations that will break things if scale up is performed during refresh that you may not find in your testing, depending on how extensive your testing is

> postgresql

ah it looks like postgresql is not waiting to install the snap; we should fix that

postgresql is currently not starting the workload or running operations that would be dangerous during refresh if PeerRelationNotReady is raised during scale up

that will be temporarily disabled in this commit https://github.com/canonical/postgresql-operator/pull/1049/commits/7a8eb0a8f151ae6e75d7d15a91f4646cbe700199 to speed up postgresql initial install until the pg defers can be improved

src/events/refresh.py
reneradoi: I've adjusted it so that it's only retried on the initial installation, but not when called from charm-refresh.

src/events/external_clients.py
reneradoi: Yes, that is handled. I've fixed an issue there that I just found when going through the code again. It's living in the `tls_events.certificate_available()`.

src/events/backup.py
reneradoi: changed from "upgrade" to "refresh".

src/workload.py
```diff
@@ -40,20 +44,30 @@ def start(self) -> None:
             logger.exception(str(e))
 
     @retry(stop=stop_after_attempt(3), wait=wait_fixed(5), reraise=True)
```
carlcsaposs-canonical:
```suggestion
    @retry(stop=stop_after_attempt(3), wait=wait_fixed(5), reraise=True, retry=tenacity.retry_if_exception_type(EtcdServiceError))
```
so that install is not retried for other exceptions when called in refresh_snap

carlcsaposs-canonical (changes requested):

src/charm.py
reneradoi: After discussion with other teams, managers and Mykola we decided to declare scale-up during upgrades as not supported, for now. The main concern here is that it is too intrusive to delay all other events until the refresh peer relation is eventually available and would cause unacceptable impact on the setup phase for every single charm deployment or scaling operation.

This will be stated in the docs, and if another solution not relying on deferring all events is possible, might be changed at a later point to support scale-up during upgrades again.

With this decision, I believe we can resolve this conversation.

carlcsaposs-canonical: > another solution not relying on deferring all events is possible

fyi this is possible, mysql-router-operator implements it

but, at least for postgresql, implementing it might require significant changes to charm architecture so likely not worth implementing now

carlcsaposs-canonical (approved):
thank you!

will follow up about the concerns with action running statues with an issue on data-platform-helpers; not applicable here since charm is not currently using action running statuses and @reneradoi confirmed that

> as we don't use that [action running statuses] in etcd (and also will not for the foreseeable future)

reneradoi:
@skourta @Mehdi-Bendriss 

After all discussions have been settled, I want to ask again you for your review/approval of the changes to this PR after your previous approval.

The main changes are:
- We do no longer consider `UnitTearingDown` and `PeerRelationNotReady` from `charm-refresh` for events such as config changes, unit removal and other "dangerous" events (such as actions), to avoid long and "unnecessary" deferrals
- config changes are processed if they are updates of the IP address
- When converting statuses from charm-refresh to advanced statuses, we no longer handle errors but let them raise, to make sure important refresh-statuses never get overridden (see [comment](https://github.com/canonical/charmed-etcd-operator/pull/139#discussion_r2287627299)).
- The `install` method for the snap has been modified to be more flexible regarding retries: If called on the initial install, it will retry as before (I fixed a bug along the way that prevented retry...). If called from charm-refresh, it will not retry.
- TLS transition will be blocked (this was not the case until now), but TLS certificate rotation will be allowed
- We will officially not support scale-up while an upgrade is in progress (see [discussion thread](https://github.com/canonical/charmed-etcd-operator/pull/139#discussion_r2287647393) above for further details)

skourta (approved):

Mehdi-Bendriss (approved):

PR #140, opened 2025-08-12
[DPE-7916] Fix restart after vm reboot

reneradoi:
**Issue:**
If a unit is rebooted, the etcd service is not restarted correctly.

**Solution:**
Always restart the etcd service when processing the `on_start` event, no matter if it was a controlled shutdown (flag `unit_server.is_started` gets removed)  or not.

The PR also adds integration test coverage for rebooting the Raft leader and all units of a cluster.

### DIFF
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -196,13 +196,14 @@ class EtcdEvents(Object):
                         # if removing fails, we cannot start the workload or the member would crash
                         raise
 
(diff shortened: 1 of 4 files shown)


Mehdi-Bendriss (approved):
Thanks René!

skourta (approved):
Thank you @reneradoi. LGTM

PR #77, opened 2025-05-09
[DPE-7664] - POC: Enable using Vault as TLS provider

reneradoi:
This PR enables `charmed-etcd` to use [Vault](https://github.com/canonical/vault-k8s-operator) as provider for TLS certificates for both peer and client certificates.

## **Context**
When using Vault as intermediate CA, it currently is required to use a common name that is a subdomain of the common name configured in Vault. This is not ideal (see https://github.com/canonical/vault-k8s-operator/issues/599) and will be adjusted by the Vault team (see [TE173](https://docs.google.com/document/d/1N3as8iO1uldoDqF2kkCV0DhC_cK60jcRmLeZ4VpMByc/edit?usp=drivesdk)). Until then, we need a way to work around this issue by adding Vaults configured common name as a prefix to the common name in our certificates / signing requests.

**Functionality**
We add a new configuration option `common_name`. This can be used as a workaround for the issue mentioned above by setting this config to match the common name configured to Vault.

It can also be used to make the common name used in etcd's certificates match an organizations required naming pattern.

If set, the `common_name` value will be appended to the actual common name in each certificate / signing request when requesting TLS certificates. The following example shows how the common name in a certificate request would look like if `common_name` config option is set:

```
common_name=""
Common Name (CN): etcd-9-f81e1f7d-4418-49a3-88cd-e26c600de05c

common_name=mydomain.com
Common Name (CN): etcd-9-f81e1f7d-4418-49a3-88cd-e26c600de05c.mydomain.com
```

## **Implementation**
- add `common_name` config option to `config.yaml`
- read config value when instantiating the `TLSEvents` class
- append config value to `{unit.name}.{model.uuid}` for certificate requests of both peer and client certificates
- append config value to `sans_dns` for certificate requests
- replace `/` with `-` in `unit.name` for certificate requests (`/` not allowed by Vault)

### TLS Certificate Enhancements:

* **Configurable Common Name:** Added a new `common_name` option in `config.yaml` to allow customization of the Common Name used in certificate signing requests. This ensures compatibility with Vault as a TLS provider. (`config.yaml`, [config.yamlR22-R27](diffhunk://#diff-d8d0422389f03d783e32e627250fe29834bd09c6361640d1ff00661dd6820034R22-R27))
[...]

PR #144, opened 2025-08-25
[DPE-7917] Fix TLS attributes for Vault usage

reneradoi:
**Issue**

In order to use Vault as TLS provider, it is required to omit `/` in TLS names such as common name or DNS sans. Otherwise Vault will not generate TLS certificates for issued certificate requests.

**Solution**
Adjust the TLS common name and DNS sans to remove the `/`.

This PR also adds: 
- an integration test for using Vault as TLS provider (based on the draft from @skourta in #77)
- a disclaimer to the docs with explanations about required Vault version and configuration


skourta (approved):
Great work @reneradoi

Mehdi-Bendriss (approved):

PR #146, opened 2025-08-27
[DPE-7826] Upgrades basic integration tests

reneradoi:
This PR provides basic integration test coverage for upgrading charmed etcd.

**Functional changes**
- update etcd to version 3.6.2 (to test the workload upgrade)
- fix handling of upgrades relevant statuses:
  - finding from testing: upgrades have to be first in status priority order to avoid low-priority statuses being overridden, such as `etcd 3.6.1 running; Snap revision 15 (outdated); Charm version 3.6/1.0.0`
  - `health check failed` should not be set as running status, it should only be shown after the upgrades hook has completed (if failed)

**Test coverage**
- deploy from charmhub with the previous workload version of etcd (`3.6.1`, revisions and versions configurable)
- run an upgrade to `3.6.2`, make it fail after the first unit upgraded (because of cluster health) and roll back to `3.6.1`
- run the full upgrade to `3.6.2` for all units
- during upgrades, ensure cluster availability with continuous writes
- note: for now, upgrades will always be marked as incompatible if upgrading to a locally built charm (because of the charm version compatibility check) -> it is required to run the `force-refresh-start` action with `check-compatibility=false` to make the upgrade happen

### DIFF
--- a/tests/integration/ha/test_upgrade_and_rollback.py
+++ b/tests/integration/ha/test_upgrade_and_rollback.py
@@ -28,8 +28,8 @@ from .helpers import (
 logger = logging.getLogger(__name__)
 
 NUM_UNITS = 3
-PREVIOUS_REVISIONS = {"x86_64": 15, "aarch64": 16}
-WORKLOAD_VERSIONS = {"previous": "3.6.1", "target": "3.6.2"}
+CHARM_REVISIONS_TO_DEPLOY = {"x86_64": 89, "aarch64": 88}
+WORKLOAD_VERSION_TARGET = "3.6.2"
 
 
 async def test_deploy_previous_etcd_version(ops_test: OpsTest) -> None:
@@ -38,7 +38,7 @@ async def test_deploy_previous_etcd_version(ops_test: OpsTest) -> None:
         APP_NAME,
         num_units=NUM_UNITS,
         channel="3.6/edge",
-        revision=PREVIOUS_REVISIONS[machine()],
+        revision=CHARM_REVISIONS_TO_DEPLOY[machine()],
     )
 
     await wait_until(ops_test, apps=[APP_NAME], timeout=1000, wait_for_exact_units=NUM_UNITS)
@@ -113,7 +113,7 @@ async def test_upgrade_to_local(charm: str, ops_test: OpsTest) -> None:
         unit_endpoint = get_unit_endpoint(ops_test, unit_name=unit.name, app_name=APP_NAME)
         assert (
             get_etcd_version(unit_endpoint, user=INTERNAL_USER, password=password)
-            == WORKLOAD_VERSIONS["target"]
+            == WORKLOAD_VERSION_TARGET
         ), f"unit {unit.name} was not upgraded"
 
         assert any(unit.name.replace("/", "") == member["name"] for member in cluster_members), (
--- a/tests/integration/ha/test_upgrade_and_rollback.py
+++ b/tests/integration/ha/test_upgrade_and_rollback.py
@@ -61,18 +61,6 @@ async def test_fail_upgrade_and_rollback(charm: str, ops_test: OpsTest) -> None:
     # start writing data to the cluster
     start_continuous_writes(endpoints=endpoints, user=INTERNAL_USER, password=password)
 
-    # initiate the upgrade
-    logger.info(f"Refresh etcd to v{WORKLOAD_VERSION['target']}")
-    await etcd_application.refresh(path=charm)
-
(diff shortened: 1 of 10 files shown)


tests/integration/ha/test_upgrade_and_rollback.py
```diff
@@ -0,0 +1,223 @@
...
+    logger.info("Running `force-refresh-start` action with check-compatibility=false")
+    # we need to ignore pre-refresh-checks here because cluster health will be degraded immediately
+    await refresh_order[0].run_action(
+        "force-refresh-start", **{"check-compatibility": False, "run-pre-refresh-checks": False}
+    )
+
+    logger.info(f"Pause etcd service on unit {refresh_order[-1].name} to force upgrade to fail")
+    await disable_etcd_service(ops_test, unit_name=refresh_order[-1].name)
+
+    # no chance to know exactly which status messages to look for here, so we use `wait_for_idle`
```
skourta: Can you please explain?

skourta (dismissed):
Thank you René. I ran the upgrades manually and in integration tests. It looks good.
I left one question about statuses.

tests/integration/ha/test_upgrade_and_rollback.py
reneradoi: Thank you for reviewing @skourta!

During the upgrade, we will see statuses like this:
```
App           Version  Status   Scale  Charm         Channel  Rev  Exposed  Message
charmed-etcd  3.6.2    blocked      3  charmed-etcd             0  no       Refreshing. Check units >=2 are healthy & run `resume-refresh` on unit 1. To rollback, `juju refresh --revision 89`

Unit             Workload     Agent      Machine  Public address  Ports     Message
charmed-etcd/0*  waiting      idle       0        10.68.59.70     2379/tcp  etcd 3.6.1; Snap revision 15 (outdated); Charm version 3.6/1.0.0.post17.dev0+2c3d8be.dirty
charmed-etcd/1   active       idle       1        10.68.59.160    2379/tcp  etcd 3.6.1 running; Snap revision 15 (outdated); Charm version 3.6/1.0.0.post17.dev0+2c3d8be.dirty
charmed-etcd/2   maintenance  idle       2        10.68.59.151    2379/tcp  health check failed
```

From my point of view, there is no way of consistently checking this kind of combination of unit statuses with `wait_until`. Checking for the app status is not an option here because this could miss the events happening in case we need to run the `force-refresh-start` action due to incompatibility. So the only stable option here is to just `wait_for_idle` units instead of using `wait_until`.

I've revisited the usages of `wait_for_idle` and only kept those necessary. The test run should be fine now again (needed a couple of iterations to single them out...).

tests/integration/ha/test_upgrade_and_rollback.py
```diff
@@ -0,0 +1,220 @@
...
+    # initiate the upgrade
+    logger.info(f"Refresh etcd to v{WORKLOAD_VERSION['target']}")
+    await etcd_application.refresh(path=charm)
+
+    logger.info(f"Pause etcd service on unit {refresh_order[-1].name} to force upgrade to fail")
+    await disable_etcd_service(ops_test, unit_name=refresh_order[-1].name)
+
+    # versions will always be marked "incompatible" if refresh to a local version
+    # see: https://github.com/canonical/charm-refresh/blob/main/charm_refresh/_main.py#L182-L185
+    await ops_test.model.wait_for_idle(apps=[APP_NAME], wait_for_exact_units=NUM_UNITS)
```
Mehdi-Bendriss, line 80: We should rather use `wait_until` to stay consistent across all tests (applies for all the test)

Mehdi-Bendriss (commented):
Thanks René, looks good! Only a minor comment

tests/integration/ha/test_upgrade_and_rollback.py
reneradoi: Thanks for the review @Mehdi-Bendriss!

Using `wait_until` unfortunately is not sufficient in this case. I've tried it, but with all the different statuses it is not possible in all cases. I've tried to eliminate every single use of `wait_until`, but those cases remaining didn't work (also see [comment above](https://github.com/canonical/charmed-etcd-operator/pull/146#discussion_r2313375806)).

The reasons mostly are:
- lots of different statuses between the units
- statuses don't matter, but we want to wait for units to become idle
- the statuses we are waiting for are no `StatusObjects` because they are set directly by `charm-refresh`
- the statuses we are waiting for cannot be determined in advance because they depend on the charm version string created at build time

Mehdi-Bendriss (approved):

skourta (approved):

PR #150, opened 2025-09-02
[DPE-7907] Upgrades edge cases integration tests

reneradoi:
This PR adds integration tests for edge cases during upgrades. It also fixes/adjusts functionality based on the test findings.

It is based on #146 and includes this PR's changes until merged.

The corresponding Jira ticket is [DPE-7907](https://warthogs.atlassian.net/browse/DPE-7907).

## Solution:
**changes/fixes:**
- add safeguards that overwrite the `force-new-cluster` config on disk after successful completion of the respective workflow, to avoid a new cluster being forced in case of an uncoordinated restart (issue was found while testing upgrades after `rebuild-cluster`). This is relevant for these workflows:
  - rebuild-cluster
  - storage reuse
- update the `public_ip` value in the unit databag after ip address change (so far only the `private_ip` was updated)
- no longer block `update_status` if upgrade is in progress, to allow for detection of cluster failures and expiring certificates
- no longer block `rebuild-cluster` action if upgrade is in progress to allow disaster recovery in case of majority failure during upgrade

**test scenarios:**
- upgrade a 1-unit cluster
- upgrade a 2-unit cluster and run `rebuild-cluster` while upgrade is in progress
- process an updated ip address while upgrading
- rotate TLS certificates while upgrading
- scale up while upgrading
- scale down while upgrading

All of the tests ensure HA (via continuous writes) while running the operations, except for:
- the 1-unit cluster - no HA
- the 2-unit clusters - some writes might fail when the snap is refreshed, therefore we ignore missing revisions

- [X] Manually
- [X] Unit tests
- [X] Integration tests

All tests where run manually before automating them. Side effects from the functional changes (such as storage reuse workflow) have also been tested manually.

All tests where executed on a remote VM with Juju on LXD.

[DPE-7907]: https://warthogs.atlassian.net/browse/DPE-7907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -152,6 +152,8 @@ class EtcdEvents(Object):
                         self.charm.state.unit_server.peer_url
                     )
                     self.charm.state.cluster.update({"authentication": "enabled"})
+                    # overwrite the `force-new-cluster` config after cluster has been initialized
+                    self.charm.config_manager.set_config_properties()
                 except ValueError:
                     logger.error("Failed to update member configuration")
                     event.defer()
@@ -245,7 +247,7 @@ class EtcdEvents(Object):
         ip_address = self.charm.workload.get_host_mapping().get("private_ip")
         if ip_address and ip_address != self.charm.state.unit_server.ip:
             logger.info(f"New ip address: {ip_address}")
-            self.charm.state.unit_server.update({"private_ip": ip_address})
+            self.charm.state.unit_server.update(self.charm.workload.get_host_mapping())
 
             # we need to update the client-urls by restarting etcd
             self.charm.config_manager.set_config_properties()
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -143,6 +143,8 @@ class EtcdEvents(Object):
                 statuses_state=self.charm.state.statuses,
             )
             self.charm.cluster_manager.start_member()
+            # overwrite the `force-new-cluster` config after cluster has been initialized
+            self.charm.config_manager.set_config_properties()
 
             if storage_reuse:
                 # this is a new application but storage is reused
@@ -152,8 +154,6 @@ class EtcdEvents(Object):
                         self.charm.state.unit_server.peer_url
                     )
                     self.charm.state.cluster.update({"authentication": "enabled"})
-                    # overwrite the `force-new-cluster` config after cluster has been initialized
-                    self.charm.config_manager.set_config_properties()
                 except ValueError:
                     logger.error("Failed to update member configuration")
                     event.defer()
@@ -621,6 +621,8 @@ class EtcdEvents(Object):
                 logger.info("Enabling and starting etcd again.")
                 self.charm.workload.enable_service()
                 self.charm.cluster_manager.start_member()
+                # overwrite the `force-new-cluster` config after cluster has been rebuilt
+                self.charm.config_manager.set_config_properties()
                 self.charm.state.unit_server.update({"rebuild_completed": "True"})
             elif (
                 all(unit.rebuild_completed for unit in self.charm.state.servers)
@@ -629,8 +631,6 @@ class EtcdEvents(Object):
             ):
                 logger.info("All units started again - cluster rebuild completed.")
                 self.charm.state.cluster.update({"rebuild_cluster": ""})
-                # overwrite the `force-new-cluster` config after cluster has been rebuilt
-                self.charm.config_manager.set_config_properties()
 
             if self.charm.state.cluster.learning_member:
                 self.charm.cluster_manager.promote_learning_member()
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -315,7 +315,8 @@ class EtcdEvents(Object):
                     event.defer()
                     return
 
-            # reflect membership updates in the cluster state, e.g. ip change or tls switchover
+            # reflect membership updates in the cluster state and config file in case of restarts
+            # e.g. ip change or tls switchover
             self.charm.cluster_manager.update_cluster_member_state()
 
             self._update_client_relations()
@@ -626,6 +627,8 @@ class EtcdEvents(Object):
             ):
                 logger.info("All units started again - cluster rebuild completed.")
                 self.charm.state.cluster.update({"rebuild_cluster": ""})
+                # overwrite the `force-new-cluster` config after cluster has been rebuilt
+                self.charm.config_manager.set_config_properties()
 
             if self.charm.state.cluster.learning_member:
                 self.charm.cluster_manager.promote_learning_member()
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -401,11 +401,10 @@ class EtcdEvents(Object):
             not self.charm.state.cluster.cluster_state
             or self.charm.state.cluster.is_restore_in_progress
             or self.charm.state.cluster.rebuild_cluster_in_progress
-            or self.charm.refresh_in_progress
         ):
             return
 
-        if not self.charm.workload.alive():
+        if not self.charm.workload.alive() and not self.charm.refresh_in_progress:
             if not self.charm.cluster_manager.restart_member():
                 self.charm.status.set_running_status(
                     EtcdServiceStatuses.SERVICE_NOT_RUNNING.value,
@@ -422,7 +421,7 @@ class EtcdEvents(Object):
             # if anything fails with the metrics request, we don't want to panic
             pass
 
-        if self.charm.unit.is_leader():
+        if self.charm.unit.is_leader() and not self.charm.refresh_in_progress:
             try:
                 self.charm.cluster_manager.clean_users()
                 self.charm.cluster_manager.remove_inconsistent_members_if_required()
@@ -682,9 +681,6 @@ class EtcdEvents(Object):
         """
         if not self.charm.unit.is_leader():
             return "Action must be performed on the leader unit."
-
-        if self.charm.refresh_in_progress:
-            return "Refresh in progress, cannot perform action."
 
         if self.charm.state.cluster.is_backup_in_progress:
             return "Backup in progress, cannot perform action."
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -143,8 +143,6 @@ class EtcdEvents(Object):
                 statuses_state=self.charm.state.statuses,
             )
             self.charm.cluster_manager.start_member()
-            # overwrite the `force-new-cluster` config after cluster has been initialized
-            self.charm.config_manager.set_config_properties()
 
             if storage_reuse:
                 # this is a new application but storage is reused
@@ -154,6 +152,8 @@ class EtcdEvents(Object):
                         self.charm.state.unit_server.peer_url
                     )
                     self.charm.state.cluster.update({"authentication": "enabled"})
+                    # overwrite the `force-new-cluster` config after cluster has been initialized
+                    self.charm.config_manager.set_config_properties()
                 except ValueError:
                     logger.error("Failed to update member configuration")
                     event.defer()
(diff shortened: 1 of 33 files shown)


src/events/etcd.py
```diff
@@ -152,6 +152,8 @@ def _on_start(self, event: ops.StartEvent) -> None:  # noqa: C901
                         self.charm.state.unit_server.peer_url
                     )
                     self.charm.state.cluster.update({"authentication": "enabled"})
+                    # overwrite the `force-new-cluster` config after cluster has been initialized
+                    self.charm.config_manager.set_config_properties()
```
Mehdi-Bendriss: Can you explain this? we're already resetting with `force-new-cluster` the config earlier in the method (line 134) in the event of a reuse, given https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/src/managers/config.py#L75-L79

src/events/etcd.py
```diff
@@ -245,7 +247,7 @@ def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
         ip_address = self.charm.workload.get_host_mapping().get("private_ip")
         if ip_address and ip_address != self.charm.state.unit_server.ip:
             logger.info(f"New ip address: {ip_address}")
-            self.charm.state.unit_server.update({"private_ip": ip_address})
+            self.charm.state.unit_server.update(self.charm.workload.get_host_mapping())
```
Mehdi-Bendriss, line 250: good catch

src/events/etcd.py
```diff
@@ -627,6 +629,8 @@ def _rebuild_cluster(self) -> None:  # noqa: C901
             ):
                 logger.info("All units started again - cluster rebuild completed.")
                 self.charm.state.cluster.update({"rebuild_cluster": ""})
+                # overwrite the `force-new-cluster` config after cluster has been rebuilt
+                self.charm.config_manager.set_config_properties()
```
Mehdi-Bendriss: Similar comment: 
1. this would have been called in an earlier stage of rebuilding no? 
2. Why do it _post_ start

src/events/refresh.py
```diff
@@ -88,13 +87,6 @@ def refresh_snap(
         self.charm.workload.start()
         if self.charm.cluster_manager.is_healthy():
             refresh.next_unit_allowed_to_refresh = True
-        else:
-            self.charm.status.set_running_status(
```
Mehdi-Bendriss: I saw this in an earlier PR, and lost context, can you explain the reasoning behind why the HEALTH_CHECK_FAILED is no longer a running status?

src/managers/upgrades.py
```diff
@@ -40,21 +40,23 @@ def get_statuses(self, scope: Scope, recompute: bool = False) -> list[StatusObje
...
-        This logic ignores all statuses set directly by the refresh lib, as of refresh v3.1.0.
         For more information: see https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/status/
         """
         status_list: list[StatusObject] = []
 
         if not self.refresh:
             return [CharmStatuses.ACTIVE_IDLE.value]
 
-        if refresh_app_status := self.refresh.app_status_higher_priority:
+        if scope == "app" and (refresh_app_status := self.refresh.app_status_higher_priority):
```
Mehdi-Bendriss: heh, nice

tests/integration/ha/helpers.py
```diff
@@ -167,3 +167,21 @@ async def reboot_unit(ops_test: OpsTest, unit_name: str) -> None:
     reboot_cmd = f"exec --unit {unit_name} -- sudo reboot"
     await ops_test.juju(*reboot_cmd.split(), check=True)
     logger.info(f"Rebooted unit {unit_name}.")
+
+
+async def disable_etcd_service(ops_test: OpsTest, unit_name: str) -> None:
```
Mehdi-Bendriss: This was already part of previous PR, no? I wonder why it's still showing the diff

Mehdi-Bendriss (commented):
Thanks René, left a few questions. I will review the tests once the PR is rebased to 3.6/edge

tests/integration/ha/helpers.py
reneradoi: Yes that is correct, it was only shown because this PR was not rebased yet.

src/events/refresh.py
reneradoi: Mostly a design decision. I want to avoid running statuses as much as possible, because they are scattered all over the code and have to be reset in all places where they could become obsolete. This is not very clean and reduces code readability. And in this case it's fine (from my point of view) if the failing health check is reported at the end of the hook.

src/events/etcd.py
reneradoi: Yes that is true. But if an uncoordinated restart happens (without being triggered by our charm), such as service restart by `systemd`, the current config is taken from the file on disk. The value `force-new-cluster` is still there if we don't actively overwrite the file itself, and this can potentially lead to loosing a cluster if not prevented.

I've seen that when manually testing `cluster-rebuild` while upgrading. For that I stopped the etcd service on one unit, but didn't disable it. After (!) the cluster was rebuild, I stopped the service again manually to set how upgrades behave if the cluster is unhealthy. I ended up with a cluster of one unit after the unit being restarted by `systemd`, because `force-new-cluster` was still there and no workflow controlled adding the other unit to the forced new cluster.

I consider the `force-new-cluster` to be very dangerous if we leave it there. That's why I think this safeguard one-liner is acceptable, even though it seems redundant.

src/events/etcd.py
reneradoi: Same comment as above: Safeguard to overwrite the config on disk, to avoid data / cluster loss.

src/events/etcd.py
Mehdi-Bendriss: If I understand correctly, it looks like we need to reset this flag to `False` on the config file on disk after a start_service, meaning: 
- when we need it, we explicitly set it to True
- As soon as we start, we set it to False on disk, so any subsequent restart will not consider it a new bootstrap until we explicitely state so

reneradoi: To be explicit and not cause any unwanted side effects, I would do it as part of the workflows that require the flag set to `True`. But you are right, it should happen right after the start. I will move it further ahead and test the workflows.

reneradoi: Turns out we can't overwrite it immediately, because the actual startup is "too slow" and the config is already overwritten on disk before the etcd service has even loaded the config and started. This results in the deployment to fail because etcd doesn't immediately start and auth cannot be enabled. Confirmed with a manual deployment:

```
unit-charmed-etcd-0: 14:35:48 DEBUG root:Emitting Juju event start.
unit-charmed-etcd-0: 14:35:48 DEBUG managers.config:Writing configuration
[...]
unit-charmed-etcd-0: 14:35:49 DEBUG managers.config:Writing configuration
[...]
unit-charmed-etcd-0: 14:35:52 ERROR common.client:etcdctl user command failed: returncode: 2, error: [...] connection refused
```

Journal log:
```
$ journalctl -u snap.charmed-etcd.etcd
Sep 03 14:35:48 juju-324bd0-0 systemd[1]: Started snap.charmed-etcd.etcd.service - Service for snap application charmed-etcd.etcd.
Sep 03 14:35:50 juju-324bd0-0 systemd[1]: snap.charmed-etcd.etcd.service: Main process exited, code=exited, status=1/FAILURE
Sep 03 14:35:50 juju-324bd0-0 systemd[1]: snap.charmed-etcd.etcd.service: Failed with result 'exit-code'.
Sep 03 14:36:10 juju-324bd0-0 systemd[1]: snap.charmed-etcd.etcd.service: Scheduled restart job, restart counter is at 1.
Sep 03 14:36:10 juju-324bd0-0 systemd[1]: Started snap.charmed-etcd.etcd.service - Service for snap application charmed-etcd.etcd.
```

In the etcd log:
```
{"level":"info","ts":"2025-09-03T14:35:50.056943Z","msg":"loaded server configuration, other configuration command line flags and environment variables will be ignored if provided","path":"/var/snap/charmed-etcd/17/etcd.conf.yml"}
{"level":"info","ts":"2025-09-03T14:35:50.057436Z","msg":"starting an etcd server","initial-cluster-state":"existing"
{"level":"error","ts":"2025-09-03T14:35:50.074763Z","caller":"etcdserver/server.go:309","msg":"bootstrap failed"[...]}
```

So we have to keep it at the end of the respective workflows.

reneradoi: I've now added a check to the cluster manager's `start_member()` method to check if the service is reachable after starting (incl. a reasonable retry mechanism). This method is used in the `on_start` and `rebuild-cluster` workflows, and it will raise if it fails, in order to re-run the hook if not successful.

Furthermore I've moved the overwriting of the config `force-new-cluster` to immediately after the restart.

Mehdi-Bendriss (dismissed):

tests/integration/ha/upgrades/test_upgrade_edge_cases.py
```diff
@@ -0,0 +1,372 @@
...
+
+    logger.info(f"Force new ip address for {refresh_order[-1].name}")
+    ip_renewal_hostname = await hostname_from_unit(ops_test, unit_name=refresh_order[-1].name)
+    old_unit_ip = await ip_address_from_unit(ops_test, unit_name=refresh_order[-1].name)
+    old_unit_endpoint = get_unit_endpoint(ops_test, unit_name=refresh_order[-1].name)
+    cut_network_from_unit_with_ip_change(ip_renewal_hostname)
+
+    # make sure the unit is not reachable from the controller
+    controller_hostname = await get_controller_hostname(ops_test)
+    assert not is_unit_reachable(controller_hostname, ip_renewal_hostname)
```
skourta: Add message in case of failure

skourta:
```suggestion
    assert not is_unit_reachable(controller_hostname, ip_renewal_hostname), f"unit {refresh_order[-1].name} is still reachable from controller"
```

tests/integration/ha/upgrades/test_upgrade_edge_cases.py
```diff
@@ -0,0 +1,372 @@
...
+        ops_test,
+        apps=[APP_NAME],
+        apps_statuses=["blocked"],
+        units_statuses=["active"],
+        wait_for_exact_units=NUM_UNITS,
+    )
+
+    # ensure the unit is up again
+    new_unit_ip = await ip_address_from_unit(ops_test, unit_name=refresh_order[-1].name)
+    logger.info(f"{refresh_order[-1].name} is available again with new ip {new_unit_ip}")
```
skourta, line 223: Maybe add a check on the new IP and whether it is reachable again?

src/events/etcd.py
skourta: +1

tests/integration/ha/upgrades/test_upgrade_cluster_size_related_tests.py
```diff
@@ -0,0 +1,280 @@
...
+        logger.info("Upgrade is blocked due to incompatibility")
+
+        logger.info("Running `force-refresh-start` action with check-compatibility=false")
+        await etcd_application.units[0].run_action(
+            "force-refresh-start", **{"check-compatibility": False}
+        )
+
+    # wait for upgrade to complete
+    await wait_until(ops_test, apps=[APP_NAME])
+    assert_continuous_writes_increasing(endpoints=endpoints, user=INTERNAL_USER, password=password)
```
skourta, line 75: We mostly use `assert_continuous_writes_increasing` to check cluster health, correct?

skourta (commented):
Minor suggestions and some comments. Great tests @reneradoi

tests/integration/ha/upgrades/test_upgrade_cluster_size_related_tests.py
reneradoi: Yes, could say so. With the continuous writes, we ensure the cluster is available not only for reading but also for writing.

tests/integration/ha/upgrades/test_upgrade_edge_cases.py
reneradoi: That implicitly happens on the previous line, where we connect into the unit and run the `hostname -i` command.

tests/integration/ha/upgrades/test_upgrade_edge_cases.py
reneradoi: Thanks for the hint!

skourta (approved):

delgod (approved):

PR #152, opened 2025-09-11
adjust upgrades integration test

reneradoi:
The integration test for upgrade rollback fails outside of a PR with the current setup, see [scheduled test runs](https://github.com/canonical/charmed-etcd-operator/actions/runs/17631153177)

The issue for this is that rolling back from locally built version to published one is only possible to the latest revision. But if the local built version is the same as the latest published, the rollback will not do anything, reporting: `Charm is up-to-date.` This is true, as the versions are the same, but means the rollback to the published version was not completed.

## Solution:
To roll back to the initially deployed version, we need to issue another rollback command.

- [X] Manually
- [ ] Unit tests
- [x] Integration tests

Manual testing was executed by:
- deploying revision 94 from charmhub
- checking out the repository and packing a local version
- refreshing to that local version
- after one unit was refreshed, rolling back with `--switch charmed-etcd`
- issuing another rollback command with `juju refresh charmed-etcd --revision 94`


smethnani (approved):
Looks good to me!

skourta (approved):

PR #153, opened 2025-09-25
pin juju agent-version to ' 3.6.9' because of issue #20714

reneradoi:
Juju release `3.6.10` introduced a regression: New IP addresses are not picked up after an IP address change, see [#20714](https://github.com/juju/juju/issues/20714).

## Solution:
Fix the Juju agent-version used in integration tests to the previous version `3.6.9`, to allow for proper testing and unblock releases.

Therefore we temporarily disable bootstrapping a Juju lxd controller with concierge and instead do that ourselves in the spread script.

- [ ] Manually
- [ ] Unit tests
- [X] Integration tests

### DIFF
--- a/concierge.yaml
+++ b/concierge.yaml
@@ -4,7 +4,7 @@ juju:
 providers:
   lxd:
     enable: true
-    bootstrap: true
+    bootstrap: false
 host:
   snaps:
     jhack:
--- a/concierge.yaml
+++ b/concierge.yaml
@@ -1,7 +1,7 @@
 juju:
   model-defaults:
     logging-config: <root>=INFO; unit=DEBUG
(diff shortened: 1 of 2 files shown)


Mehdi-Bendriss (approved):

concierge.yaml
```diff
@@ -4,7 +4,7 @@ juju:
 providers:
   lxd:
     enable: true
-    bootstrap: true
```
skourta, line 7: Can you please leave a TODO to revert back when everything settles? We do not want to diverge from the rest of the charm.

reneradoi: I'm watching the issue, as soon as it get's fixed I'll revert the change.

skourta (approved):

PR #155, opened 2025-10-01
[DPE-8007] Add config option for certificate domain

reneradoi:
When using Vault as TLS provider for etcd, we currently only support it with the configuration `pki_allow_any_name=true`. In order to allow using Vault (or another TLS provider) with a more restrictive configuration, we need to provide a specific domain as part of the common name and DNS SANs in the certificate requests.

[DPE-8007](https://warthogs.atlassian.net/browse/DPE-8007)

## Solution:
The PR adds configuration options for `client-certificate-domain` and `peer-certificate-domain`. The values set to these configurations will be appended to the common name and all DNS SANs of the respective certificate requests, if valid. If invalid, a blocked status will be displayed to the user.

The existing code structure can mostly be used for this, the foundation has already been laid when implementing the `certificate-extra-sans` configuration. Now, only two new methods `certificate_domain_config_is_valid()` and `build_common_name()` have been added to the TLS manager.

**Important note:**
Etcd's communication is strictly IP-based, both for peer-to-peer as well as client-to-server communication. It is not possible to disable IP SANs in the certificate requests. For more information see: [breaking change in v3.2](https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.2.md#breaking-changes) and [etcd issue 6336](https://github.com/etcd-io/etcd/issues/6336)

The PR also adds user documentation for the new configuration options and a note regarding IP SANs to the existing TLS docs.

- [X] Manually
- [X] Unit tests
- [X] Integration tests

Manual testing was performed with both self-signed-certificates for regression testing as well as Vault for ensuring that we are now able to use Vault as TLS provider with restricted domain names (also covered with integration test).

[DPE-8007]: https://warthogs.atlassian.net/browse/DPE-8007?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/how-to/tls/enable-tls.md
+++ b/docs/how-to/tls/enable-tls.md
@@ -30,6 +30,8 @@ This guide will use the [Self-signed Certificates](https://charmhub.io/self-sign
 Check [this guide](https://charmhub.io/topics/security-with-x-509-certificates) for an overview of all the TLS certificates charms available. 
 
 If [Vault](https://charmhub.io/vault-k8s?channel=1.18/edge) is used as TLS provider, it is required in version 1.18 at least. 
+See [](../tune-settings.md/#certificate-sans-configuration) for information on how to configure the certificate domain 
+in accordance to Vault's configuration.
 ```
 
 Deploy the `self-signed-certificates` charm. etcd uses `v4` of the [{spellexception}`tls-certificates` library](https://charmhub.io/tls-certificates-interface/libraries/tls_certificates), which is currently only supported in the `edge` channel.
--- a/docs/how-to/tls/enable-tls.md
+++ b/docs/how-to/tls/enable-tls.md
@@ -14,6 +14,10 @@ etcd also supports mutual TLS authentication, which provides an additional layer
 
 ## Deploy a TLS provider
 
+```{attention}
+Etcd's communication is strictly IP-based, both for peer-to-peer as well as client-to-server communication. Make sure to choose a TLS provider that supports IP SANs.
+```
+
 Charmed etcd provides the option of using different CA certificates for client-server and peer-to-peer communication. This allows you to have different levels of trust for the two types of communication. You can also use the same CA certificate for both types of communication.
 
 You can enable peer-to-peer encryption alone, client-to-server encryption alone, or both at the same time.
@@ -25,8 +29,7 @@ This guide will use the [Self-signed Certificates](https://charmhub.io/self-sign
 
 Check [this guide](https://charmhub.io/topics/security-with-x-509-certificates) for an overview of all the TLS certificates charms available. 
 
-If [Vault](https://charmhub.io/vault-k8s?channel=1.18/edge) is used as TLS provider, it is required in version 1.18 at least. Currently Vault 
-is only supported with the configuration option `pki_allow_any_name=true`, allowing for any domain name in requested certificates. 
+If [Vault](https://charmhub.io/vault-k8s?channel=1.18/edge) is used as TLS provider, it is required in version 1.18 at least. 
 ```
 
 Deploy the `self-signed-certificates` charm. etcd uses `v4` of the [{spellexception}`tls-certificates` library](https://charmhub.io/tls-certificates-interface/libraries/tls_certificates), which is currently only supported in the `edge` channel.
--- a/src/managers/tls.py
+++ b/src/managers/tls.py
@@ -288,12 +288,6 @@ class TLSManager(ManagerStatusProtocol):
         """
         sans_ip = set()
 
-        if tls_type == TLSType.CLIENT and not self.state.config.get(
-            "client-certificate-include-ip-sans"
-        ):
-            # return early without any IP's because IP SANs are disabled
-            return frozenset(sans_ip)
-
         if self.extra_sans_config_is_valid() and (
             extra_sans_config := self.state.config.get("certificate-extra-sans")
         ):
--- a/src/managers/tls.py
+++ b/src/managers/tls.py
@@ -294,6 +294,7 @@ class TLSManager(ManagerStatusProtocol):
             or tls_type == TLSType.PEER
             and not self.state.config.get("peer-certificate-include-ip-sans")
         ):
+            # return early without any IP's because IP SANs are disabled
             return frozenset(sans_ip)
 
         if self.extra_sans_config_is_valid() and (
@@ -335,6 +336,7 @@ class TLSManager(ManagerStatusProtocol):
         sans_dns.add(self.state.unit_server.unit_name.replace("/", ""))
         sans_dns.add(self.workload.get_host_mapping()["hostname"])
 
+        # add desired domain to every DNS name if configured
         if self.certificate_domain_config_is_valid(tls_type) and (
             certificate_domain_config := self.state.config.get(f"{tls_type}-certificate-domain")
         ):
--- a/src/managers/tls.py
+++ b/src/managers/tls.py
@@ -548,4 +548,10 @@ class TLSManager(ManagerStatusProtocol):
         if not self.extra_sans_config_is_valid():
             status_list.append(TLSStatuses.SANS_CONFIG_INVALID.value)
 
+        if not self.certificate_domain_config_is_valid(TLSType.CLIENT):
+            status_list.append(TLSStatuses.CLIENT_DOMAIN_CONFIG_INVALID.value)
+
+        if not self.certificate_domain_config_is_valid(TLSType.PEER):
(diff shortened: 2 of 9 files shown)


docs/how-to/tls/enable-tls.md
```diff
@@ -25,8 +29,7 @@ This guide will use the [Self-signed Certificates](https://charmhub.io/self-sign
 
 Check [this guide](https://charmhub.io/topics/security-with-x-509-certificates) for an overview of all the TLS certificates charms available. 
 
-If [Vault](https://charmhub.io/vault-k8s?channel=1.18/edge) is used as TLS provider, it is required in version 1.18 at least. Currently Vault 
-is only supported with the configuration option `pki_allow_any_name=true`, allowing for any domain name in requested certificates. 
+If [Vault](https://charmhub.io/vault-k8s?channel=1.18/edge) is used as TLS provider, it is required in version 1.18 at least.
```
skourta, line 32: Should we add some info about configuring the domain? To make it know that the domain needs to be configured on both  etcd and vault

src/managers/tls.py
```diff
@@ -325,8 +326,30 @@ def build_sans_dns(self) -> frozenset[str]:
...
+            updated_sans_dns = set()
+            for san in sans_dns:
+                updated_sans_dns.add(san + "." + certificate_domain_config)
+            return frozenset(updated_sans_dns)
+
         return frozenset(sans_dns)
 
+    def build_common_name(self, common_name: str, tls_type: TLSType) -> str:
+        """Build the Common Name for the TLS certificate."""
+        if not (
```
skourta: You can simplify the code by doing it the same as `build_sans_dns`

src/managers/tls.py
```diff
@@ -348,6 +371,19 @@ def extra_sans_config_is_valid(self) -> bool:
 
         return True
 
+    def certificate_domain_config_is_valid(self, tls_type: TLSType) -> bool:
+        """Validate configuration value for {peer|client}-certificate-domain option."""
+        if not (
```
skourta: Same you can simplify the code by testing the presence of `certificate_domain_config` and that it is not valid

src/statuses.py
```diff
@@ -120,9 +120,17 @@ class TLSStatuses(Enum):
     )
     SANS_CONFIG_INVALID = StatusObject(
         status="blocked",
-        message="Invalid value set for the config options 'certificate-extra-sans'",
+        message="Invalid value set for the config option 'certificate-extra-sans'",
         short_message="Invalid value `certificate-extra-sans`",
     )
+    CLIENT_DOMAIN_CONFIG_INVALID = StatusObject(
```
skourta, line 126: Let's normalize our statuses and add `for config option` in these as well

tests/integration/tls/test_vault_intermediate_ca.py
```diff
@@ -220,3 +221,55 @@ async def test_tls_enabled(ops_test: OpsTest) -> None:
...
+    await wait_until(ops_test, apps=[APP_NAME], wait_for_exact_units=NUM_UNITS)
+
+    await download_client_certificate_from_unit(ops_test, APP_NAME)
+    client_cert_subject = subprocess.getoutput("openssl x509 -noout -subject -in client.pem ")
+    assert etcd_domain_config_value in client_cert_subject, (
+        f"expected domain name {etcd_domain_config_value} not found in certificate subject {client_cert_subject}"
+    )
+    logger.info("Certificates in etcd updated with new domain")
+
+    logger.info("Set config in etcd to invalid value")
```
skourta, line 258: I would rather have this in a separate test instead of the vault test

skourta (commented):
Thank you @reneradoi. I left some minor comments.

src/managers/tls.py
reneradoi: Make sense - updated.

src/managers/tls.py
reneradoi: updated

src/statuses.py
reneradoi: updated

docs/how-to/tls/enable-tls.md
reneradoi: Good idea.

tests/integration/tls/test_vault_intermediate_ca.py
reneradoi: updated

docs/how-to/tls/enable-tls.md
```diff
@@ -30,6 +30,8 @@ This guide will use the [Self-signed Certificates](https://charmhub.io/self-sign
 Check [this guide](https://charmhub.io/topics/security-with-x-509-certificates) for an overview of all the TLS certificates charms available. 
 
 If [Vault](https://charmhub.io/vault-k8s?channel=1.18/edge) is used as TLS provider, it is required in version 1.18 at least. 
+See [](../tune-settings.md/#certificate-sans-configuration) for information on how to configure the certificate domain
```
skourta, line 33: @a-velasco is it ok to have relative links in the docs?

skourta (dismissed):
Thank you, @reneradoi !

docs/how-to/tls/enable-tls.md
reneradoi: We do already have, see [here](https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/docs/how-to/backup-and-restore/migrate-etcd.md?plain=1#L20).

a-velasco: It is not the best practice (I only discovered this myself recently) - we should instead use targets and references.

It would look like this: 

`tune-settings.md`
```diff
+(certificate-sans-configuration)=
## Certificate SANs configuration
```

`enable-tls.md`
```
See {ref}`certificate-sans-configuration`
```

Same goes for referencing full documents; you just add `(target-name)=` before the title heading.

I would not consider this a blocker for this PR, but we should definitely start doing this moving forward. I'm planning a bigger PR where I'll replace all references in the docs.

reneradoi: Thanks for the insights Andreia. I suggest I don't mess around with it now and instead wait for a general rework of this.

a-velasco: Agreed!

a-velasco (dismissed):
LGTM! Looks like the failed link checks were server-side issues with Charmhub links that are resolved now, so no actions needed there.

src/managers/tls.py
```diff
@@ -325,8 +326,27 @@ def build_sans_dns(self) -> frozenset[str]:
...
+            return frozenset(updated_sans_dns)
+
         return frozenset(sans_dns)
 
+    def build_common_name(self, common_name: str, tls_type: TLSType) -> str:
+        """Build the Common Name for the TLS certificate."""
+        if self.certificate_domain_config_is_valid(tls_type) and (
+            certificate_domain_config := self.state.config.get(f"{tls_type}-certificate-domain")
+        ):
+            return common_name + "." + certificate_domain_config
```
Mehdi-Bendriss: nit:

```suggestion
            return f"{common_name}.{certificate_domain_config}"
```

src/managers/tls.py
```diff
@@ -348,6 +368,16 @@ def extra_sans_config_is_valid(self) -> bool:
 
         return True
 
+    def certificate_domain_config_is_valid(self, tls_type: TLSType) -> bool:
+        """Validate configuration value for {peer|client}-certificate-domain option."""
+        if (
+            certificate_domain_config := self.state.config.get(f"{tls_type}-certificate-domain")
+        ) and not self._is_hostname(certificate_domain_config):
```
Mehdi-Bendriss, line 375: nit:

```suggestion
            certificate_domain_config := self.state.config.get(f"{tls_type}-certificate-domain")
            and not self._is_hostname(certificate_domain_config)
        ):
```

Mehdi-Bendriss (commented):
Thanks René! I have a couple of nits, but looks good

src/managers/tls.py
reneradoi: updated

src/managers/tls.py
reneradoi: That will not work.

Mehdi-Bendriss: right 🤦🏻‍♂️

Mehdi-Bendriss (approved):

skourta (approved):

PR #156, opened 2025-10-02
[DPE-8403] Improve error handling for missing permissions on user secrets

reneradoi:
Users can configure a password for the root user with a user secret. If this secret is not granted to the etcd application, the `secret_changed` / `config_changed` event for processing the secret will fail and a status message gets displayed.

But after updating the secret permissions, no new event is triggered and the desired password can not be applied, until users manually update the config or secret.

[DPE-8403](https://warthogs.atlassian.net/browse/DPE-8403)

## Solution:
In order to work around this gap, the failed events with permission errors should be deferred and retried later. This allows for processing the password, once users have granted permissions for the secret to the etcd application.

## Unrelated changes:
- fix an issue with status computation while TLS switchover is in progress (the previous fix in #138) was not sufficient
- observe `storage_attached` events for all defined storages (following MongoDB approach as discussed with @patriciareinoso)
- fix a pretty critical issue with IP address changes when client TLS is enabled (discovered while testing, see [test run](https://github.com/canonical/charmed-etcd-operator/actions/runs/18227789817/job/51913188907#step:13:4956))
  - the `config_changed` event crashes when the operator runs the health check after restarting
  - this happens because the client TLS cert used by the operator is not valid anymore (was issued for the old IP address)
  - if the failed health check would be ignored, all subsequent actions such as updating the member information with the new peer url will also fail
  - the only way out is to request new client certs before doing anything else on the cluster, and until then, defer the IP change (`config_changed` event)

- [X] Manually
- [x] Unit tests
- [x] Integration tests

All manual tests where executed on a remote VM with LXD:
- The status computation fix was tested manually by enabling and disabling peer TLS.
- The renewal of TLS certs was tested with a manual network interruption + IP address change for one unit, both with peer and client TLS enabled; an integration test has been added for IP address change with client certificates to increase test coverage
- The `storage_attached` events where tested manually by deploying a cluster with storage constraints, removing units and the application and re-attaching the detached storage.
- The secret permission issue was tested manually by deploying an application with configured secret, but without secret access given to the charm. After failing to process this, the access was granted and the next `update_status` event was checked for the password update. Integration test coverage was also added for this.

[DPE-8403]: https://warthogs.atlassian.net/browse/DPE-8403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -244,7 +244,7 @@ class EtcdEvents(Object):
                 statuses_state=self.charm.state.statuses,
             )
 
-    def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
+    def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:  # noqa: C901
         """Handle config_changed event."""
         if (
             self.charm.state.cluster.is_restore_in_progress
@@ -260,6 +260,15 @@ class EtcdEvents(Object):
         ip_address = self.charm.workload.get_host_mapping().get("private_ip")
         if ip_address and ip_address != self.charm.state.unit_server.ip:
             logger.info(f"New ip address: {ip_address}")
+            # before we do any cluster operation, we must update client certificates
+            # otherwise the certificate will be invalid and the cluster operation will fail
+            if self.charm.tls_manager.certificate_sans_require_update(TLSType.CLIENT):
+                logger.info("Updating TLS certificates because of new IP address")
+                self.charm.tls_events.refresh_tls_certificates_event.emit()
+                event.defer()
+                return
+
+            # after TLS certificates are renewed, we come back here and update the unit state
             self.charm.state.unit_server.update(self.charm.workload.get_host_mapping())
 
             # we need to update the client-urls by restarting etcd
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -47,7 +47,7 @@ from literals import (
     TLSState,
     TLSType,
 )
-from statuses import CharmStatuses, ClusterStatuses, EtcdServiceStatuses
+from statuses import CharmStatuses, ClusterStatuses, EtcdServiceStatuses, TLSStatuses
 
 if TYPE_CHECKING:
     from charm import EtcdOperatorCharm
@@ -262,12 +262,23 @@ class EtcdEvents(Object):
                 and self.charm.tls_manager.certificate_sans_require_update(TLSType.CLIENT)
             ):
                 logger.info("Updating TLS certificates because of new IP address")
+                self.charm.status.set_running_status(
+                    TLSStatuses.CERT_REFRESH_IP_CHANGE.value,
+                    scope="unit",
+                    component_name=self.charm.tls_manager.name,
+                    statuses_state=self.charm.state.statuses,
+                )
                 self.charm.tls_events.refresh_tls_certificates_event.emit()
                 event.defer()
                 return
 
             # after TLS certificates are renewed, we come back here and update the unit state
             self.charm.state.unit_server.update(self.charm.workload.get_host_mapping())
+            self.charm.state.statuses.delete(
+                TLSStatuses.CERT_REFRESH_IP_CHANGE.value,
+                scope="unit",
+                component=self.charm.tls_manager.name,
+            )
 
             # we need to update the client-urls by restarting etcd
             self.charm.config_manager.set_config_properties()
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -30,11 +30,13 @@ from common.exceptions import (
     RaftLeaderNotFoundError,
 )
 from literals import (
+    ARCHIVE_STORAGE,
     CLIENT_PORT,
     DATA_STORAGE,
     DATABASE_DIR,
     INTERNAL_USER,
     INTERNAL_USER_PASSWORD_CONFIG,
+    LOG_STORAGE,
     PEER_RELATION,
     SNAP_ARCHIVE_PATH,
     SNAP_DATA_PATH,
@@ -87,15 +89,25 @@ class EtcdEvents(Object):
             self.charm.on[DATA_STORAGE].storage_attached, self._on_storage_attached
         )
         self.framework.observe(
+            self.charm.on[ARCHIVE_STORAGE].storage_attached, self._on_storage_attached
+        )
+        self.framework.observe(
+            self.charm.on[LOG_STORAGE].storage_attached, self._on_storage_attached
+        )
+        self.framework.observe(
             self.charm.on.rebuild_cluster_action, self._on_rebuild_cluster_action
         )
 
     def _on_storage_attached(self, event: ops.StorageAttachedEvent) -> None:
         """Handle storage attachment."""
-        # fix the permissions of the data dir if re-attaching existing storage
         for path in [SNAP_DATA_PATH, SNAP_LOG_PATH, SNAP_ARCHIVE_PATH]:
-            self.charm.workload.exec(["chmod", "-R", "750", path])
-            self.charm.workload.exec(["chown", "-R", f"{SNAP_USER}:{SNAP_GROUP}", path])
+            try:
+                # fix the permissions of the directory if re-attaching existing storage
+                self.charm.workload.exec(["chmod", "-R", "750", path])
+                self.charm.workload.exec(["chown", "-R", f"{SNAP_USER}:{SNAP_GROUP}", path])
+            except CalledProcessError:
+                # gracefully continue if the path is not there yet
+                logger.warning("Could not adjust directory permissions")
 
     def _on_install(self, event: ops.InstallEvent) -> None:
         """Handle install event."""
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -85,15 +85,10 @@ class EtcdEvents(Object):
         self.framework.observe(
             self.charm.on[DATA_STORAGE].storage_detaching, self._on_storage_detaching
         )
-        self.framework.observe(
-            self.charm.on[DATA_STORAGE].storage_attached, self._on_storage_attached
-        )
-        self.framework.observe(
-            self.charm.on[ARCHIVE_STORAGE].storage_attached, self._on_storage_attached
-        )
-        self.framework.observe(
-            self.charm.on[LOG_STORAGE].storage_attached, self._on_storage_attached
-        )
+        for storage in [ARCHIVE_STORAGE, DATA_STORAGE, LOG_STORAGE]:
+            self.framework.observe(
+                self.charm.on[storage].storage_attached, self._on_storage_attached
+            )
         self.framework.observe(
             self.charm.on.rebuild_cluster_action, self._on_rebuild_cluster_action
         )
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -45,7 +45,7 @@ from literals import (
     TLSState,
     TLSType,
 )
-from statuses import ClusterStatuses, EtcdServiceStatuses
+from statuses import CharmStatuses, ClusterStatuses, EtcdServiceStatuses
 
 if TYPE_CHECKING:
     from charm import EtcdOperatorCharm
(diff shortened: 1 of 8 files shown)


src/managers/cluster.py
```diff
@@ -400,11 +400,8 @@ def get_statuses(self, scope: Scope, recompute: bool = False) -> list[StatusObje
         if not self.state.cluster.cluster_state:
             status_list.append(ClusterStatuses.CLUSTER_INITIALIZING.value)
 
-        if self.state.unit_server.member_endpoint not in self.state.cluster.cluster_members:
-            if self.state.unit_server.tls_peer_state in [TLSState.TO_TLS, TLSState.TO_NO_TLS]:
```
reneradoi, line 404: I've noticed that this does not work:
- the unit will immediately update it`s `TLSState` once it has finished switching over
- the `cluster_members` in the app databag will only be updated by the leader unit after `peer_relation_changed`

In between, the status will display `Waiting to join cluster`, which is misleading.

I've decided to revert the earlier change in instead just check whether the `member_name` of a unit is in the cluster members. For status computation, this should be fine.

src/events/etcd.py
```diff
@@ -261,13 +273,6 @@ def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
             if self.charm.unit.is_leader():
                 self.charm.cluster_manager.update_cluster_member_state()
 
-            # update tls certificates with new ip address
-            if self.charm.state.unit_server.tls_client_state in (
-                TLSState.TO_TLS,
-                TLSState.TLS,
-            ) or self.charm.state.unit_server.tls_peer_state in (TLSState.TO_TLS, TLSState.TLS):
-                self.charm.tls_events.refresh_tls_certificates_event.emit()
```
reneradoi, line 269: This logic had to be moved to an earlier point because if the client certificate is not updated before cluster operation (such as health check, update of peer url), it is considered invalid.

src/events/etcd.py
```diff
@@ -86,16 +88,26 @@ def __init__(self, charm: "EtcdOperatorCharm"):
...
-        # fix the permissions of the data dir if re-attaching existing storage
         for path in [SNAP_DATA_PATH, SNAP_LOG_PATH, SNAP_ARCHIVE_PATH]:
-            self.charm.workload.exec(["chmod", "-R", "750", path])
-            self.charm.workload.exec(["chown", "-R", f"{SNAP_USER}:{SNAP_GROUP}", path])
+            try:
+                # fix the permissions of the directory if re-attaching existing storage
+                self.charm.workload.exec(["chmod", "-R", "750", path])
+                self.charm.workload.exec(["chown", "-R", f"{SNAP_USER}:{SNAP_GROUP}", path])
+            except CalledProcessError:
+                # gracefully continue if the path is not there yet
```
Mehdi-Bendriss, line 104: I'm not sure i understand this one, I have a couple of questions here:
1. when would that happen? 
2. how would the charm be notified so it can reapply these permissions/ownerships

src/events/etcd.py
Mehdi-Bendriss: The moved logic is omitting the handling of `TLSState.TO_TLS` for both certificate types

src/managers/cluster.py
Mehdi-Bendriss: So you're saying that `CLUSTER_MEMBER_RECONFIGURATION` will never be set in this path, correct?

tests/integration/ha/test_storage_reuse.py
```diff
@@ -79,10 +81,15 @@ async def test_attach_storage_after_scale_down(ops_test: OpsTest) -> None:
     )
 
     # add unit with previous storage attached
-    add_unit_cmd = f"add-unit {app} --model={ops_test.model.info.name} --attach-storage={data_storage_id} --attach-storage={archive_storage_id}"
+    add_unit_cmd = f"""add-unit {app} \
+                    --model={ops_test.model.info.name} \
+                    --attach-storage={data_storage_id} \
+                    --attach-storage={archive_storage_id} \
+                    --attach-storage={log_storage_id}
```
Mehdi-Bendriss, line 88: Could you add a test where some storage volumes are added _post_ deployment? The `data` and `logs` ones ought to be interesting

tests/integration/test_charm.py
```diff
@@ -92,3 +93,41 @@ async def test_update_admin_password(ops_test: OpsTest) -> None:
...
+
+    logger.info("Updating configuration with the new secret - but without access")
+    await ops_test.model.applications[APP_NAME].set_config(
+        {INTERNAL_USER_PASSWORD_CONFIG: secret_id}
+    )
+
+    await wait_until(
+        ops_test,
+        apps=[APP_NAME],
+        apps_full_statuses={APP_NAME: [CharmStatuses.SECRET_ACCESS_ERROR.value]},
```
Mehdi-Bendriss, line 118: nice

Mehdi-Bendriss (commented):
Thanks René! I have a few questions

src/events/etcd.py
reneradoi: There's some background story here: When I initially developed the multiple storages feature, I observed all volume mounts being present on the first received `storage_attached` hook. So I simplified everything and only observed the hook for the `data` volume. 

Then in a discussion with Patricia, she rightfully made the point that there is a risk of granting permissions on paths that do not exist yet, potentially crashing the hooks. I can see a slight risk that this could happen (if not all mounts are available at the time of the data-`storage_attached` hook being processed), and I don't want to make assumptions on Juju lifecycle things we can't control. 

Therefore I decided to add this minimal safeguard logic here that allows for both: all mounts present at the first hook or each mount only present at the time of its hook. This logic is simply yet robust, because if there is any issue, the last of the three `storage_attached` hooks will still apply all of the permissions we need. If, at that time, the mounts are not present, something outside of our control went wrong and we should raise/fail, which will happen at install time then.

src/managers/cluster.py
reneradoi: Yes, correct, it was never displayed.

src/events/etcd.py
reneradoi: Yes, and that is on purpose: If we are in `TLSState.TO_TLS`, that means that etcd has not been restarted yet and client TLS is not active yet. In this case, we don't have to defer the `config_changed` event because cluster operations will work without a valid client cert.

If there was an IP change while client or peer TLS is in the process of being enabled, our current strategy is to raise ([here](https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/src/charm.py#L179-L182) and [here](https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/src/charm.py#L208-L211)) if the switchover fails, to make the user aware that enabling TLS didn't work. Requesting new certs because of an IP change while in the progress of enabling TLS doesn't even have any effect currently, because the operator would already be in error state at the time of receiving the "updated" certs. If we want to change that, we should sync and discuss separately I'd say.

tests/integration/ha/test_storage_reuse.py
reneradoi: I'm not sure if I understand correctly what you mean; add post deployment as in `juju add-storage`?

This would not work, because the storage mounts are created by default and we don't support multiple volumes per storage. Example:
```
$ juju create-storage-pool etcd-storage lxd volume-type=standard
$ juju deploy charmed-etcd --channel 3.6/edge --storage data=etcd-storage,1G,1

$ juju storage
Unit            Storage ID  Type        Pool          Size     Status    Message
charmed-etcd/0  archive/0   filesystem  rootfs        915 GiB  attached  
charmed-etcd/0  data/1      filesystem  etcd-storage  1.0 GiB  attached  
charmed-etcd/0  logs/2      filesystem  rootfs        915 GiB  attached  

$ juju add-storage charmed-etcd/0 logs=etcd-storage,2G,1
failed to add storage "logs" to charmed-etcd/0: adding "logs" storage to charmed-etcd/0: cannot attach, storage is singular
```

Or do you mean something else?

src/events/etcd.py
Mehdi-Bendriss: Let's consider this one as an edge case for now, could you create an issue for it so we don't forget?

src/events/etcd.py
```diff
@@ -86,16 +88,26 @@ def __init__(self, charm: "EtcdOperatorCharm"):
         self.framework.observe(
             self.charm.on[DATA_STORAGE].storage_attached, self._on_storage_attached
         )
+        self.framework.observe(
```
skourta: nit: You can use a for loop to group them all and shorten the code

src/events/etcd.py
```diff
@@ -248,6 +260,18 @@ def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
...
             logger.info(f"New ip address: {ip_address}")
+            # before we do any cluster operation, we must update client certificates
+            # otherwise the certificate will be invalid and the cluster operation will fail
+            if (
+                self.charm.state.unit_server.tls_client_state == TLSState.TLS
+                and self.charm.tls_manager.certificate_sans_require_update(TLSType.CLIENT)
+            ):
+                logger.info("Updating TLS certificates because of new IP address")
+                self.charm.tls_events.refresh_tls_certificates_event.emit()
+                event.defer()
```
skourta, line 272: I think we should add a status to notify the user

tests/integration/ha/test_storage_reuse.py
skourta: @reneradoi do you mean juju will automatically attach default volumes?

src/events/etcd.py
skourta: Just for clarification, In case the paths do not exist, yet the assumption is that will get another storage attached event?

src/events/etcd.py
skourta: The certificates should be reissued as soon as the config_changed event that is triggered bythe change of the IP is fired (This can come late based on when juju discovers the change). Meaning the situation will self-heal then.

skourta (commented):
Thank you @reneradoi I left some questions to understand better. It looks good already.

src/events/etcd.py
reneradoi: This is true, but it will not come into effect: As soon as the operator processes the first `certificate_available` (with the previous IP address) and restarts, the health check after the restart will fail because the client certificate will be invalid (wrong IP address). This will raise and the operator will be in error state and can't get out of it on its own.

src/events/etcd.py
reneradoi: Yes.

tests/integration/ha/test_storage_reuse.py
reneradoi: Yes it does: `When you use storage mounts with Juju, they will be automatically mounted into the charm container`, see [Ops: How to manage storage](https://documentation.ubuntu.com/ops/latest/howto/manage-storage/#declare-the-storage).

src/events/etcd.py
reneradoi: Yeah we can do, good idea.

I'm not a big fan of adding running statuses all over the place, but in this case it should be one. I don't want to add 2x `certificate_sans_require_update` to status computation on all hooks, would be too expensive IMO.

src/events/etcd.py
reneradoi: Updated, thx. Wanted to explicitly name the storages here, but with a list we loop over that's even more explicit.

src/events/etcd.py
skourta: I see. I agree with @Mehdi-Bendriss let's create an issue to track it and handle it in a future PR.

skourta (approved):
Thank you René. LGTM

src/events/etcd.py
Mehdi-Bendriss: As discussed, to clarify on the above, we should aim for BlockedStatus in the above case where exceptions are currently raised and propagated.

reneradoi: Issue is here: https://github.com/canonical/charmed-etcd-operator/issues/159

Mehdi-Bendriss (approved):

Issue #102, opened 2025-06-25
Add an integration test for metrics export


skourta:
Add an integration test for metrics export. [OpenSearch integration test](https://github.com/canonical/opensearch-operator/blob/1699feaa7e507c81a5b92e40e1fdf979c87c7632/tests/integration/plugins/test_plugins.py#L398).

reneradoi:
Integration test coverage has been added with #162.

PR #157, opened 2025-10-03
test: adds integration tests for metrics export; introduces jubilant

Soundarya03:
- Adds integration tests for metrics (#102 )
- Begins integration test rewrites for [Jubliant](https://github.com/canonical/jubilant). Rewrites test_charm.py to use Jubilant.

## Solution:
- Introduces two tests for metrics integration to test_charm.py: `test_etcd_metrics_endpoint` and `test_etcd_metrics_cos_relation`, to verify direct metrics scrape and integration with grafana-agent respectively.
- Rewrites existing tests in test_charm.py, as well as the helpers functions they invoke, to use Jubliant.

- [x] Manually (to verify jubliant models created and destroyed)
- [ ] Unit tests
- [x] Integration tests


tests/integration/helpers.py
```diff
@@ -147,6 +152,18 @@ def get_cluster_endpoints(
     )
 
 
+def get_cluster_endpoints_jubilant(
```
skourta, line 160: Eventually, we should remove all suffixes of jubilant

tests/integration/test_charm.py
```diff
@@ -27,19 +33,25 @@
 TEST_VALUE = "42"
 
 
+@pytest.fixture(scope="module")
+def juju():
```
skourta: This should go in conftest

pyproject.toml
```diff
@@ -66,6 +66,7 @@ pydantic = ">= 1.0"
 boto3 = "^1.40.11"
 mypy-boto3-s3 = "^1.40.0"
 data-platform-helpers = ">=0.1.7"
+jubilant = "^1.4.0"
```
skourta, line 69: We shouldn't forget to remove `juju`, `pytest-operator`, and `pytest-asyncio` when we are fully done migrating all integration tests

tests/integration/helpers.py
```diff
@@ -493,3 +550,109 @@ def get_etcd_version(
...
+    unit_name: str,
+    target_unit_name: str,
+    relation_name: str,
+    key: str,
+) -> Optional[str]:
+    """Get relation data for an application.
+
+    Args:
+        juju: An instance of Jubilant's Juju class on which to run Juju commands
+        unit_name: The name of app 1's unit
```
skourta: We should use provider/requirer instead of 1/2

tests/integration/helpers.py
```diff
@@ -493,3 +550,109 @@ def get_etcd_version(
...
+    raise RuntimeError(f"No leader unit found for app {app}")
+
+
+def get_unit_relation_data(
+    juju: jubilant.Juju,
+    unit_name: str,
+    target_unit_name: str,
+    relation_name: str,
+    key: str,
+) -> Optional[str]:
```
skourta: We use python 3.12 we do not need `Optional` and `Union` you can use `or None` and `|`

tests/integration/helpers.py
```diff
@@ -493,3 +550,109 @@ def get_etcd_version(
...
+        RuntimeError: if no leader unit is found.
+    """
+    for unit in juju.status().get_units(app).values():
+        if unit.leader:
+            return unit.public_address
+
+    raise RuntimeError(f"No leader unit found for app {app}")
+
+
+def get_unit_relation_data(
```
skourta: We should also consider getting all the relation data and letting the user filter by target unit. (Depends on how we use the helper)

tests/integration/test_charm.py
```diff
@@ -60,35 +72,76 @@ async def test_build_and_deploy(charm: str, ops_test: OpsTest) -> None:
...
+    endpoints = get_cluster_endpoints_jubilant(juju, APP_NAME)
 
     # create a user secret and grant it to the application
     new_password = "some-password"
-    await set_password(ops_test, new_password)
-    await wait_until(ops_test, apps=[APP_NAME])
+    set_password_jubilant(juju, new_password)
+
+    # wait for config-changed hook to finish executing
+    juju.wait(lambda status: jubilant.all_agents_idle(status, APP_NAME), timeout=1200)
```
skourta: We should check if we still need `wait_unitl` to conver the complicated wait conditions we have in some tests. If that is the case we can remove `wait_until` otherwise we should migrate `wait_until` to use jubilant inside. It is ok for this case to use juju wait because it is simple enough.

tests/integration/test_charm.py
```diff
@@ -60,35 +72,76 @@ async def test_build_and_deploy(charm: str, ops_test: OpsTest) -> None:
...
         get_key(endpoints, user=INTERNAL_USER, password=new_password, key=TEST_KEY) == TEST_VALUE
     )
+
+
+@pytest.mark.abort_on_fail
+async def test_etcd_metrics_endpoint(juju: jubilant.Juju):
+    # direct metrics scrape
+    leader_unit_ip = get_leader_unit_ip(juju, app=APP_NAME)
+    endpoint = f"http://{leader_unit_ip}:{METRICS_PORT}/metrics"
+    resp = await http_request("get", endpoint)
```
skourta: I don't think we require a helper if we do not use it a lot. It looks a bit redundant to me

tests/integration/test_charm.py
```diff
@@ -60,35 +72,76 @@ async def test_build_and_deploy(charm: str, ops_test: OpsTest) -> None:
...
+        juju, f"{COS_APP_NAME}/{cos_leader_id}", leader_name, COS_RELATION_NAME, "config"
+    )
+    if not isinstance(relation_data, dict):
+        import json
+
+        relation_data = json.loads(relation_data)
+
+    # assert that right targets are set
+    scrape_job = relation_data["metrics_scrape_jobs"][0]
+    assert scrape_job["static_configs"][0]["targets"][0].endswith(f":{METRICS_PORT}")
```
skourta, line 182: I think we should also explore the possibility of querying grafana agent to see the metrics exported

skourta (commented):
Thank you @Soundarya03. I left some comments.

tests/integration/helpers.py
Soundarya03: Yes, I plan to remove it once all helpers are ported

tests/integration/test_charm.py
Soundarya03: Shall I move it once all modules make use of jubilant? This was my thought process..

pyproject.toml
Soundarya03: Yes. Perhaps I should add a TODO at all these points (the "jubilant" suffix too) so we don't miss anything while cleaning up?

tests/integration/test_charm.py
Soundarya03: Done, simply defined it as a function in test_charm.py

tests/integration/helpers.py
Soundarya03: This helper is only used in our newly introduced metrics test, as of now. In fact, this realization has now prompted me to move it to test_charm.py rather than keep it as a helper.

tests/integration/test_charm.py
Soundarya03: That's a good point. I do have one concern though: wouldn't this mean I have to deploy the COS-lite bundle's k8s charms on another model, to get grafana-agent out of blocked state and actually functioning. Would this be too many resources deployed for an integration test for etcd?

skourta: That it ok, in my opinion we already have tests on OS that do that.

pyproject.toml
skourta: Yes, I like that idea.

tests/integration/test_charm.py
skourta: Yes, we should just keep track

tests/integration/test_charm.py
Soundarya03: Done, will work on this and update the PR

tests/integration/test_charm.py
```diff
@@ -60,35 +74,148 @@ async def test_build_and_deploy(charm: str, ops_test: OpsTest) -> None:
...
+        if target_unit_name in relation_data[idx]["related-units"]:
+            break
+    else:
+        return None
+    return (
+        relation_data[idx]["related-units"].get(target_unit_name, {}).get("data", {}).get(key, {})
+    )
+
+
+async def http_request(method: str, endpoint: str):
```
reneradoi: I think we can get along without having a wrapper for requests and remove this. It could just be a one-liner like this:

```
endpoint = f"http://{leader_unit_ip}:{METRICS_PORT}/metrics"
result = requests.get(endpoint)
```

tests/integration/test_charm.py
```diff
@@ -60,35 +74,148 @@ async def test_build_and_deploy(charm: str, ops_test: OpsTest) -> None:
...
+        import json
+
+        relation_data = json.loads(relation_data)
+
+    # assert that right targets are set
+    scrape_job = relation_data["metrics_scrape_jobs"][0]
+    assert scrape_job["static_configs"][0]["targets"][0].endswith(f":{METRICS_PORT}")
+
+
+def get_unit_relation_data(
```
reneradoi: We already have a similar helper `get_application_relation_data()` in [helpers.py](https://github.com/Soundarya03/charmed-etcd-operator/blob/test/metrics-integration-test/tests/integration/helpers.py#L259), and this one here should be moved there as well, to be reusable for multiple tests.

tests/integration/helpers.py
```diff
@@ -314,6 +332,15 @@ async def get_secret_by_label(ops_test: OpsTest, label: str) -> Dict[str, str]:
     raise SecretNotFoundError(f"Secret with label {label} not found")
 
 
+def get_secret_by_label_jubilant(juju: jubilant.Juju, label: str) -> Dict[str, str]:
+    for secret in juju.secrets():
+        if label == secret.label:
+            revealed_secret = juju.show_secret(secret.uri, reveal=True)
+            return revealed_secret.content
+
+    raise SecretNotFoundError(f"Secret with label {label} not found")
```
reneradoi, line 401: Nice, this is significantly shorter than the previous code!

tests/integration/helpers.py
```diff
@@ -493,3 +551,27 @@ def get_etcd_version(
...
+
+def get_leader_unit_ip(juju: jubilant.Juju, app: str = APP_NAME) -> str:
+    """Retrieve the leader unit's public address.
+
+    Raises:
+        RuntimeError: if no leader unit is found.
+    """
+    for unit in juju.status().get_units(app).values():
+        if unit.leader:
+            return unit.public_address
```
reneradoi, line 624: Just a heads-up: Doing it this way is ok for the simple tests here. For all HA-related tests involving network interruptions or else, we should use the `ip_address_from_unit()` method in [helpers_network.py](https://github.com/Soundarya03/charmed-etcd-operator/blob/test/metrics-integration-test/tests/integration/ha/helpers_network.py#L36C11-L36C31), which covers for issues when the IP address displayed in `juju status` is not updated in time after a new IP address has been assigned to a unit.

tests/integration/test_charm.py
```diff
@@ -60,35 +74,148 @@ async def test_build_and_deploy(charm: str, ops_test: OpsTest) -> None:
...
+    juju.deploy(COS_APP_NAME, channel=COS_CHANNEL)
+    juju.integrate(APP_NAME, COS_APP_NAME)
+    # wait for units active (or blocked, in case of grafana agent)
+    juju.wait(lambda status: jubilant.all_active(status, APP_NAME), timeout=1000)
+    juju.wait(lambda status: jubilant.all_blocked(status, COS_APP_NAME), timeout=1000)
+
+    # get relation data sent to grafana-agent
+    cos_leader_id = get_leader_unit_id(juju, COS_APP_NAME)
+    leader_id = get_leader_unit_id(juju, APP_NAME)
+    leader_name = f"{APP_NAME}/{leader_id}"
```
reneradoi: You could also use the existing helper `get_juju_leader_unit_name`, if you only need it for this.

reneradoi (commented):
Hey @Soundarya03 nice work already!

I've added a couple of hints and comments, could you please have a look at them?

Furthermore, I've got a few general things:
- Could you please aim your PR at a feature branch? I've created `feature/jubilant-integration`, where we can collect all jubilant-related changes and then merge them on one go. This will help us to keep things separate from another and not mess with the changes/PR's we have in the meantime, which could end in chaos.
- Do you have a strong desire to add the "Codecoverage" bot? I don't really like the noisy bots from Github, and if you don't mind, I'd rather get away without having it enabled.

tests/integration/helpers.py
Soundarya03: Ah, okay. Thank you

tests/integration/test_charm.py
Soundarya03: Hi, @skourta , I've enhanced the test now. It should be comprehensive. Please let me know if this looks fine.

Soundarya03:
> Hey @Soundarya03 nice work already!
> 
> I've added a couple of hints and comments, could you please have a look at them?
> 
> Furthermore, I've got a few general things:
> 
> * Could you please aim your PR at a feature branch? I've created `feature/jubilant-integration`, where we can collect all jubilant-related changes and then merge them on one go. This will help us to keep things separate from another and not mess with the changes/PR's we have in the meantime, which could end in chaos.
> * Do you have a strong desire to add the "Codecoverage" bot? I don't really like the noisy bots from Github, and if you don't mind, I'd rather get away without having it enabled.

I'll decline this PR and open a new one targeted at `feature/jubilant-integration` once existing conversations are resolved and the changes are more or less approved.
Also, I'm not quite sure how codecov sneaked in...

tests/integration/conftest.py
```diff
@@ -1,17 +1,38 @@
...
+
+
+@pytest.fixture(scope="package")
+def arch() -> str:
+    """Fixture to provide the platform architecture for testing."""
+    return platforms.get(machine(), "amd64")
 
 
 @pytest.fixture
 def platform() -> str:
```
skourta: What is the difference with arch?

tests/integration/conftest.py
```diff
@@ -22,3 +43,121 @@ def charm(platform: str) -> str:
...
+            ["sudo", "microk8s", "enable", "metallb:10.64.140.43-10.64.140.49"],
+            check=True,
+        )
+
+        # Configure kubectl now
+        subprocess.run(["mkdir", "-p", str(pathlib.Path.home() / ".kube")], check=True)
+        kubeconfig = subprocess.check_output(["sudo", "microk8s", "config"])
+        with open(str(pathlib.Path.home() / ".kube" / "config"), "w") as f:
+            f.write(kubeconfig.decode())
+        for attempt in Retrying(stop=stop_after_delay(150), wait=wait_fixed(15)):
```
skourta, line 96: Why the long delays?

tests/integration/conftest.py
```diff
@@ -22,3 +43,121 @@ def charm(platform: str) -> str:
...
+                            "kubectl get po -A  --field-selector=status.phase!=Running",
+                            shell=True,
+                            stderr=subprocess.DEVNULL,
+                        ).decode()
+                    )
+                    != 0
+                ):  # We got sth different from "No resources found." in stderr
+                    raise Exception()
+
+        # Add microk8s to the kubeconfig
```
skourta: wrong place for the comment I believe.

tests/integration/conftest.py
```diff
@@ -22,3 +43,121 @@ def charm(platform: str) -> str:
...
+        config = kubeconfig.decode()
+        juju.cli("add-k8s", MICROK8S_CLOUD_NAME, "--client", stdin=config, include_model=False)
+        juju.bootstrap(MICROK8S_CLOUD_NAME, MICROK8S_CONTROLLER_NAME)
```
skourta: Any reason not to use the same controller?

tests/integration/conftest.py
```diff
@@ -22,3 +43,121 @@ def charm(platform: str) -> str:
...
+        MICROK8S_CONTROLLER_NAME,
+        include_model=False,
+    )
+    juju.cli(
+        "remove-cloud",
+        "--client",
+        MICROK8S_CLOUD_NAME,
+        include_model=False,
+    )
+    subprocess.run(["sudo", "snap", "remove", "--purge", "microk8s"], check=True)
```
skourta, line 142: We should have an option to avoid installing/removing microk8s when testing locally

tests/integration/conftest.py
```diff
@@ -22,3 +43,121 @@ def charm(platform: str) -> str:
...
+    subprocess.run(["sudo", "snap", "remove", "--purge", "kubectl"], check=True)
+
+
+@pytest.fixture(scope="module")
+def k8s_controller(k8s_cloud: str, juju: Juju):
```
skourta: Missing return type

tests/integration/conftest.py
```diff
@@ -22,3 +43,121 @@ def charm(platform: str) -> str:
...
+@pytest.fixture(scope="module")
+def juju(arch: str):
+    with jubilant.temp_model() as juju:
+        juju.wait_timeout = 1000
+        juju.cli("set-model-constraints", f"arch={arch}")
+        yield juju
+
+
+@pytest.fixture(scope="module")
+async def k8s_cloud(juju: Juju):
```
skourta: Missing return type

tests/integration/conftest.py
```diff
@@ -22,3 +43,121 @@ def charm(platform: str) -> str:
     # juju bundle files expect local charms to begin with `./` or `/` to distinguish them from
     # Charmhub charms.
     return f"./charmed-etcd_ubuntu@24.04-{platform}.charm"
+
+
+@pytest.fixture(scope="module")
+def juju(arch: str):
```
skourta, line 40: Missing return type

tests/integration/conftest.py
```diff
@@ -22,3 +43,121 @@ def charm(platform: str) -> str:
...
+def k8s_controller(k8s_cloud: str, juju: Juju):
+    controllers = json.loads(juju.cli("controllers", "--format", "json", include_model=False))
+    for controller, details in controllers.get("controllers").items():
+        if k8s_cloud == details.get("cloud"):
+            logger.info(f"Identified K8s controller: {controller}")
+            yield controller
+
+
+@pytest.fixture(scope="module")
+def lxd_cloud(juju: Juju):
```
skourta: Missing return type and in all subsequent functions

tests/integration/conftest.py
```diff
@@ -22,3 +43,121 @@ def charm(platform: str) -> str:
...
+    clouds = json.loads(juju.cli("clouds", "--format", "json", include_model=False))
+    for cloud, details in clouds.items():
+        if "lxd" == details.get("type"):
```
skourta: I like the yoda style conditions

tests/integration/conftest.py
```diff
@@ -22,3 +43,121 @@ def charm(platform: str) -> str:
...
+def lxd_controller(lxd_cloud: str, juju: Juju):
+    controllers = json.loads(juju.cli("controllers", "--format", "json", include_model=False))
+    for controller, details in controllers.get("controllers").items():
+        if lxd_cloud == details.get("cloud"):
+            logger.info(f"Identified LXD controller: {controller}")
+            yield controller
+
+
+@pytest.fixture(scope="module")
+def juju_lxd(arch: str, lxd_cloud: str, lxd_controller):
```
skourta: The name is a bit ambiguous. Maybe add `model` in the name?

tests/integration/test_charm.py
```diff
@@ -96,38 +111,146 @@ async def test_update_admin_password(ops_test: OpsTest) -> None:
...
-    await ops_test.model.applications[APP_NAME].set_config(
-        {INTERNAL_USER_PASSWORD_CONFIG: secret_id}
-    )
+    juju_lxd.config(app=APP_NAME, values={INTERNAL_USER_PASSWORD_CONFIG: secret_id})
 
-    await wait_until(
-        ops_test,
-        apps=[APP_NAME],
-        apps_full_statuses={APP_NAME: [CharmStatuses.SECRET_ACCESS_ERROR.value]},
+    def secret_access_error(status: jubilant.Status, app: str) -> bool:
```
skourta: I would prefer we get this out of the function or use a lambda function if it is not reused.

tests/integration/test_charm.py
```diff
@@ -96,38 +111,146 @@ async def test_update_admin_password(ops_test: OpsTest) -> None:
...
 
     logger.info("Secret access will be granted now - wait for updated password")
     # deferred `config_changed` event will be retried before `update_status`
-    async with ops_test.fast_forward("10s"):
-        await ops_test.model.grant_secret(secret_name=secret_name, application=APP_NAME)
+    with fast_forward(juju_lxd):
+        juju_lxd.grant_secret(identifier=secret_name, app=APP_NAME)
 
-    await wait_until(ops_test, apps=[APP_NAME])
+    juju_lxd.wait(
```
skourta: This is ok for simple tests but we will have to migrate `wait_until`.

tests/integration/test_charm.py
```diff
@@ -96,38 +111,146 @@ async def test_update_admin_password(ops_test: OpsTest) -> None:
...
+    if not isinstance(relation_data, dict):
+        relation_data = json.loads(relation_data)
+
+    # assert that right targets are set in grafana-agent
+    scrape_job = relation_data["metrics_scrape_jobs"][0]
+    assert scrape_job["static_configs"][0]["targets"][0].endswith(f":{METRICS_PORT}")
+
+    # further, verify integration with cos-lite bundle.
+    # Some charms in this bundle are only supported by amd64
+    if "amd64" == arch:
```
skourta: Maybe separate the tests so we don't have to set up the k8s env if we are running arm.

skourta (commented):
Great work @Soundarya03! I left some comments

tests/integration/test_charm.py
Soundarya03: Yes, I actually began the exercise, then realized it's better if I take it up as part of the next task as this one kept growing.

tests/integration/test_charm.py
Soundarya03: Makes sense

tests/integration/conftest.py
Soundarya03: We install microk8s snap (and subsequently remove it) only when no existing k8s cloud is detected

tests/integration/conftest.py
Soundarya03: Thanks, Smail!

tests/integration/conftest.py
Soundarya03: I referred to [this](https://github.com/canonical/opensearch-operator/blob/2/edge/tests/integration/relations/conftest.py#L27) implementation in OIDC; I edited other portions to suit charmed-etcd and our adoption of jubilant, but felt that these values were likely setup for a reason

tests/integration/conftest.py
Soundarya03: Thanks for pointing this out. Indeed, `platform` and `arch` fixtures are simply duplicates now. I'll clean this up.

tests/integration/test_charm.py
Soundarya03: That's a good thought

tests/integration/conftest.py
Soundarya03: Hmm, good question. I followed what was mentioned in [this how-to](https://canonical-charmed-etcd.readthedocs-hosted.com/how-to/enable-monitoring/#deploy-cos), and since it mentions "a K8s controller" I just followed that. I ran some tests on local now though to use a single controller , and it seems to be working just fine. I'll be pushing a change to do so in these tests as well.

reneradoi (approved):
Great work @Soundarya03!

Soundarya03:
Opened fresh PR (#162) to merge changes to `feature/jubilant-integration`, thus closing this one.

PR #161, opened 2025-10-15
[DPE-8425]: Integrate data interfaces v1

skourta:

This pull request introduces a refactor to the etcd charm's internal data interfaces updating them to use the new `v1` data platform libraries and pydantic-based models. 

**Data Interface Refactor and Model Updates:**

* Migrated from `charms.data_platform_libs.v0.data_interfaces` to `v1`, replacing legacy data classes with new repository interfaces and pydantic models (`PeerAppModel`, `PeerUnitModel`) throughout `src/core/cluster.py` and `src/core/models.py`. This affects how relation and unit data is accessed, stored, and updated. [[1]](diffhunk://#diff-728670c1b94eb83ab287f8abf7b8c88c7c80b237e1ea8a8191a942976dcb42dcL10-L30) [[2]](diffhunk://#diff-e46c88bb687d77b94764a732b3e08a0a7bd544a64e606429c0b787c13e1c226aL10-R20)
* Replaced direct dictionary-based relation data access (`relation_data`) with model-based access via repository interfaces, updating all properties and methods in `EtcdServer` and `EtcdCluster` to use the new models. [[1]](diffhunk://#diff-e46c88bb687d77b94764a732b3e08a0a7bd544a64e606429c0b787c13e1c226aR35-R99) [[2]](diffhunk://#diff-e46c88bb687d77b94764a732b3e08a0a7bd544a64e606429c0b787c13e1c226aL97-R198) [[3]](diffhunk://#diff-e46c88bb687d77b94764a732b3e08a0a7bd544a64e606429c0b787c13e1c226aL165-R345)
* Updated the `update` method in `RelationState` to write data using model setters and repository interfaces, ensuring proper type handling and validation.

**Interface and API Changes:**

* Updated cluster and unit interface construction to use new repository interfaces (`OpsPeerRepositoryInterface`, `OpsPeerUnitRepositoryInterface`, etc.), replacing previous `DataPeerData` and related classes. [[1]](diffhunk://#diff-728670c1b94eb83ab287f8abf7b8c88c7c80b237e1ea8a8191a942976dcb42dcL48-L51) [[2]](diffhunk://#diff-728670c1b94eb83ab287f8abf7b8c88c7c80b237e1ea8a8191a942976dcb42dcL72-R88)
* Added new helper methods for building requirer/provider models for external relations, supporting more robust contract handling. [[1]](diffhunk://#diff-728670c1b94eb83ab287f8abf7b8c88c7c80b237e1ea8a8191a942976dcb42dcL129-R143) [[2]](diffhunk://#diff-728670c1b94eb83ab287f8abf7b8c88c7c80b237e1ea8a8191a942976dcb42dcR202-R218)

- [x] Manually
- [x] Unit tests
- [x] Integration tests


src/core/models.py
```diff
@@ -94,50 +149,53 @@ def member_name(self) -> str:
...
+        return self.data_interface.build_model(self.relation.id) if self.relation else None
 
     @property
     def peer_url(self) -> str:
         """The peer connection endpoint for the etcd server."""
         scheme = "https" if self.tls_peer_state in [TLSState.TLS, TLSState.TO_NO_TLS] else "http"
         # `peer_url` MUST be IP address, etcd does not support using DNS names for the listeners
         # see https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.2.md#breaking-changes
-        return f"{scheme}://{self.ip}:{PEER_PORT}"
+        return f"{scheme}://{self.model.private_ip}:{PEER_PORT}" if self.model else ""
```
reneradoi, line 165: I know that it won't work without, but it is very annoying to have `if self.model else ...` all over the place :(

src/events/etcd.py
```diff
@@ -362,6 +366,20 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent) -> None:
...
+            and self.charm.state.unit_server.tls_peer_ca_rotation_state
+            == TLSCARotationState.NO_ROTATION
+            and self.charm.state.unit_server.tls_client_ca_rotation_state
+            == TLSCARotationState.NO_ROTATION
+            and (all_cas := self.charm.tls_manager.collect_client_cas())
+            != self.charm.tls_manager.load_trusted_ca(TLSType.CLIENT)
+        ):
+            logger.debug("CAs have changed, updating client truststore")
+            self.charm.tls_manager.update_cas(all_cas, TLSType.CLIENT)
+            self.charm.rolling_restart()
```
reneradoi, line 384: Do I understand correctly that this is the required change because only the Juju leader will receive the `mtls_cert_updated` event? And the `peer_relation_changed` to all non-leader units is triggered by the leader updating the `managed_user` in the app databag?

src/events/external_clients.py
```diff
@@ -58,8 +63,13 @@ def __init__(self, charm: "EtcdOperatorCharm"):
             self.charm.on[EXTERNAL_CLIENTS_RELATION].relation_broken, self._on_relation_broken
         )
 
-    def _on_mtls_cert_updated(self, event: MTLSCertUpdatedEvent) -> None:  # noqa: C901
+    def _on_mtls_cert_updated(self, event: MtlsCertUpdatedEvent[RequirerCommonModel]) -> None:  # noqa: C901
```
reneradoi, line 66: If I understand the changes correctly, then we should be able to simplify the logic here, especially the part after `if self.charm.unit.is_leader()` in line 137 (because we don't get the event for non-leader units anymore).

src/core/models.py
```diff
@@ -28,23 +32,71 @@
...
         substrate: SUBSTRATES,
     ):
         self.relation = relation
         self.data_interface = data_interface
         self.component = component
         self.substrate = substrate
-        self.relation_data = self.data_interface.as_dict(self.relation.id) if self.relation else {}
 
-    def update(self, items: dict[str, str]) -> None:
+    def update(self, items: dict[str, Any]) -> None:
```
reneradoi, line 102: What's the reason for this change?

src/core/models.py
```diff
@@ -55,27 +107,30 @@ def update(self, items: dict[str, str]) -> None:
...
 
         for field in delete_fields:
-            # use del instead of pop here because of error with dataplatform-libs
-            try:
-                del self.relation_data[field]
-            except KeyError:
-                pass
+            setattr(model, field.replace("-", "_"), None)
+
+        self.data_interface.write_model(self.relation.id, model)
```
reneradoi, line 120: That is really convenient for us, thank you for adding it!

reneradoi (commented):
That's a lot of work, thank you Smail! I don't see major blockers from my side, you have done a great job of maintaining at least some degree of readability for our code.

I've done a bit of manual testing, and I think that looks good, too. The comments are mostly for my understanding, and maybe we can simplify the client relations handler a little bit.

src/events/etcd.py
skourta: Partially, yes, correct. This brought several edge cases, hence the lengthy condition.

src/events/external_clients.py
skourta: We do get the `MtlsCertUpdatedEvent` on non-leader units for `SecretChanged`. It is only on the first `RelationChanged` when the mtls_cert is first added that we don't get `MtlsCertUpdatedEvent` on the non-leader units.

src/core/models.py
skourta: Setting managed users for example is now a dict because pydantic parses it.

src/core/models.py
skourta: There have been some discussion as well about this with Mehdi and Neha. There are a few options, but none of them are optimal.

Mehdi-Bendriss: Yeah this is a major point we need to improve, no good solution so far, unless we create proxy properties but that would defeat the point...

src/core/models.py
```diff
@@ -94,50 +149,53 @@ def member_name(self) -> str:
...
+            TLSState(self.model.tls_peer_state or TLSState.NO_TLS.value)
+            if self.model
+            else TLSState.NO_TLS
+        )
 
     @property
     def peer_cert_ready(self) -> bool:
         """Check if the peer certificate is ready."""
-        return self.relation_data.get("peer_cert_ready", "") == "True"
+        return self.model.peer_cert_ready == "True" if self.model else False
```
Mehdi-Bendriss, line 196: so we are now always implicitely checking on the existence of the peer rel

src/events/external_clients.py
```diff
@@ -41,7 +44,9 @@ def __init__(self, charm: "EtcdOperatorCharm"):
         super().__init__(charm, key="etcd_events")
         self.charm = charm
 
-        self.etcd_provides = EtcdProvides(self.charm, EXTERNAL_CLIENTS_RELATION)
+        self.etcd_provides = ResourceProviderEventHandler(
+            self.charm, EXTERNAL_CLIENTS_RELATION, RequirerCommonModel, mtls_enabled=True
```
Mehdi-Bendriss, line 48: this is fine for the time being, but we'll need to then add support for BulkEvents instead, so we only get a single request for all user requests of a client (meaning 1 restart)

src/managers/cluster.py
```diff
@@ -376,13 +376,17 @@ def get_statuses(self, scope: Scope, recompute: bool = False) -> list[StatusObje
         ):
             return [EtcdServiceStatuses.SERVICE_NOT_INSTALLED.value]
 
-        if not self.state.peer_relation:
+        if (
+            not self.state.peer_relation
```
Mehdi-Bendriss: I don't think this needed anymore, since it is implicitely tested on in `self.model`

Mehdi-Bendriss (commented):
Thank you Smail, looks straightforward. Great work! 
The main issues I see, are what we discussed on the `.model` part which I find redundant and not very pretty. And support for bulk mTLS (in a future PR)

src/events/external_clients.py
skourta: Yes, I considered adding it at this stage, but it would bring significant logic change, so I decided to postpone it after stable.

src/managers/cluster.py
skourta: Yes, that is true.

tests/integration/client_relations/requirer-charm/src/charm.py
```diff
@@ -97,10 +94,10 @@ def common_name(self) -> str:
         if not self.etcd_relation:
             return "requirer-charm"
         request = self.etcd_relation_local_model.requests[0]
-        if not request.mtls_cert or request.mtls_cert.get_secret_value() == "":
+        if not request.mtls_cert or request.mtls_cert == "":
```
Mehdi-Bendriss:
```suggestion
        if not request.mtls_cert:
```

src/events/external_clients.py
```diff
@@ -163,9 +174,29 @@ def _on_mtls_cert_updated(self, event: MTLSCertUpdatedEvent) -> None:  # noqa: C
...
+                            )
+                            if res.request_id == event.request.request_id
+                        ),
+                        None,
+                    ) or ResourceProviderModel(
+                        username=common_name,
+                        request_id=event.request.request_id,
+                        resource=event.request.resource,
+                        salt=event.request.salt,
+                    )
```
Mehdi-Bendriss: nit: This should be better off in a separate method, this method is already quite big

tests/integration/ha/upgrades/test_upgrade_edge_cases.py
```diff
@@ -95,7 +95,12 @@ async def test_disaster_recovery_during_upgrade(charm: str, ops_test: OpsTest) -
     rebuild_response = await rebuild_action.wait()
     assert rebuild_response.results.get("return-code") == 0, "rebuild failed"
 
-    await ops_test.model.wait_for_idle(apps=[APP_NAME], wait_for_exact_units=2)
+    # TODO remove once we have a start upgrade tests with a charm revision with v1
+    # data interfaces v1 uses - instead of _ for relation data keys
+    # this breaks disaster recovery during upgrades if the upgraded unit is not the leader
+    await ops_test.model.wait_for_idle(
+        apps=[APP_NAME], wait_for_exact_units=2, raise_on_error=False
```
Mehdi-Bendriss, line 102: Please create a followup PR for this, we will forget it

Mehdi-Bendriss (commented):
Thanks Smail! Looks good. I only have a couple of small comments.

tests/integration/ha/upgrades/test_upgrade_edge_cases.py
skourta: I have created a [jira ticket](https://warthogs.atlassian.net/browse/DPE-8716?atlOrigin=eyJpIjoiZTk0NGU5MmMwNmM1NGY4ZGJkNzdiMWQ3MmZiNDcwYzMiLCJwIjoiaiJ9) for it

reneradoi (approved):
I think we are good. Let's wait for the integration tests to pass and then we can proceed with the merge.

Awesome work Smail!

Mehdi-Bendriss (approved):

PR #165, opened 2025-10-17
[DPE-7836]: Add backend quota backend bytes config option

skourta:
[Spec](https://docs.google.com/document/d/1oTPbR65NSQxWvG5Fjkwo18bGCkArYNvmSS1mZBBsZ1A/edit?tab=t.0)

This pull request introduces a new configuration option for etcd, `quota-backend-bytes`, which sets a limit on the backend database size for each etcd member. The implementation ensures this value is validated, respects system memory constraints, and is enforced with a minimum threshold. 
## Solution:

**Etcd storage quota configuration and validation:**

* Added a new `quota-backend-bytes` option to `config.yaml` for specifying the backend database size limit, with documentation and default value set to 8 GiB.
* Introduced `MIN_QUOTA_BACKEND_BYTES` constant (100MB) and added `QUOTA_BACKEND_BYTES_CONFIG` to `TuningOptions` for consistent referencing. [[1]](diffhunk://#diff-959ba637043ab4ac88c6f9f9109022f78a7df1b9a2ab3a0407bff4f124c2db64R56-R57) [[2]](diffhunk://#diff-959ba637043ab4ac88c6f9f9109022f78a7df1b9a2ab3a0407bff4f124c2db64R118)
* Updated config validation in `ConfigManager` to check that `quota-backend-bytes` is an integer and not less than 100MB and that the database file size does not exceed this limit. [[1]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cR188-R197) [[2]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cR199-R204)
* Implemented `_get_tuning_config_value` to dynamically enforce that `quota-backend-bytes` does not exceed 90% of available memory and 90% of data storage size. [[1]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cR245-R267)

**Workload introspection:**

* Added an abstract `memory_size()` method to `Workload` and implemented it in `EtcdWorkload` to read system memory from `/proc/meminfo`. [[1]](diffhunk://#diff-4d1ed63ca48f694e77680b1434284f1d8d896cffe1842fb6b8b8e616c192b696R264-R272) [[2]](diffhunk://#diff-38281d02638d07eddd511b40cb2aa3cc1877e72f7ce4caacfe2a2a0c34cfd75fR196-R213)
* Added an abstract `get_db_file_size()` method to `Workload` and implemented it in `EtcdWorkload` to retrieve the size of the etcd database file. [[1]](diffhunk://#diff-4d1ed63ca48f694e77680b1434284f1d8d896cffe1842fb6b8b8e616c192b696R283-R290) [[2]](diffhunk://#diff-38281d02638d07eddd511b40cb2aa3cc1877e72f7ce4caacfe2a2a0c34cfd75fR224-R234)

**Config management and status reporting:**

* Updated config property generation and restart logic to use the new value resolution method for tuning options. [[1]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cL96-R98)
* Improved status messaging to include `quota-backend-bytes` in error reports for invalid tuning configs.

**Event handling:**

* Ensured configuration changes only trigger restarts if the unit server is started.
- [x] Manually
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/src/core/workload.py
+++ b/src/core/workload.py
@@ -288,3 +288,12 @@ class WorkloadBase(ABC):
             float: The size of the etcd database file in Bytes.
         """
         pass
+
+    @abstractmethod
+    def is_lxd_cloud(self) -> bool:
+        """Check if the workload is running in an LXD cloud environment.
+
+        Returns:
+            bool: True if running in LXD cloud, False otherwise.
+        """
+        pass
--- a/src/core/workload.py
+++ b/src/core/workload.py
@@ -263,29 +263,29 @@ class WorkloadBase(ABC):
         return {"hostname": hostname, "private_ip": private_ip, "public_ip": public_ip}
 
     @abstractmethod
-    def memory_size(self) -> float:
+    def memory_size(self) -> int:
         """Get the total memory size of the system in Bytes.
 
         Returns:
-            float: The total memory size in Bytes.
+            int: The total memory size in Bytes.
         """
         pass
 
     @abstractmethod
-    def data_storage_size(self) -> float:
+    def data_storage_size(self) -> int:
         """Get the size of the data storage in Bytes.
 
         Returns:
-            float: The size of the data storage in Bytes.
+            int: The size of the data storage in Bytes.
         """
         pass
 
     @abstractmethod
-    def get_db_file_size(self) -> float:
+    def get_db_file_size(self) -> int:
         """Get the size of the etcd database file in Bytes.
 
         Returns:
-            float: The size of the etcd database file in Bytes.
+            int: The size of the etcd database file in Bytes.
         """
         pass
 
--- a/src/core/workload.py
+++ b/src/core/workload.py
@@ -261,3 +261,12 @@ class WorkloadBase(ABC):
             raise ValueError("Could not get private IP address of the unit.")
 
         return {"hostname": hostname, "private_ip": private_ip, "public_ip": public_ip}
+
+    @abstractmethod
+    def memory_size(self) -> float:
+        """Get the total memory size of the system in Bytes.
+
+        Returns:
+            float: The total memory size in Bytes.
+        """
+        pass
--- a/src/core/workload.py
+++ b/src/core/workload.py
@@ -279,3 +279,12 @@ class WorkloadBase(ABC):
             float: The size of the data storage in Bytes.
         """
         pass
+
+    @abstractmethod
+    def get_db_file_size(self) -> float:
+        """Get the size of the etcd database file in Bytes.
+
+        Returns:
+            float: The size of the etcd database file in Bytes.
+        """
+        pass
--- a/src/core/workload.py
+++ b/src/core/workload.py
@@ -297,3 +297,12 @@ class WorkloadBase(ABC):
             bool: True if running in LXD cloud, False otherwise.
         """
         pass
+
+    @abstractmethod
+    def data_storage_attached(self) -> bool:
+        """Check if the data storage is attached.
+
+        Returns:
+            bool: True if data storage is attached, False otherwise.
+        """
+        pass
--- a/src/core/workload.py
+++ b/src/core/workload.py
@@ -270,3 +270,12 @@ class WorkloadBase(ABC):
             float: The total memory size in Bytes.
         """
         pass
+
+    @abstractmethod
+    def data_storage_size(self) -> float:
+        """Get the size of the data storage in Bytes.
+
+        Returns:
+            float: The size of the data storage in Bytes.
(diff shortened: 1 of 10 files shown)


src/managers/config.py
```diff
@@ -174,12 +177,51 @@ def are_tuning_parameters_valid(self) -> bool:
...
             if election_timeout < heartbeat_interval * 10 or election_timeout > 50000:
+                logger.error(
+                    "Election timeout %s is invalid. It must be at least 10x the heartbeat interval (%s) and no more than 50000ms.",
+                    election_timeout,
+                    heartbeat_interval,
+                )
                 return False
 
+        # avoid recalculating quota_backend_bytes multiple times
+        quota_backend_bytes = self.quota_backend_bytes
```
reneradoi, line 196: We need to make this fail-safe. It will not work if the storage volume has already been detached, and because `are_tuning_parameters_valid` is called on status computation, this must never fail.

Can be seen in storage-reuse integration tests: https://github.com/canonical/charmed-etcd-operator/actions/runs/18740479615/job/53470303389?pr=165#step:14:2579

src/managers/config.py
```diff
@@ -218,3 +260,43 @@ def _get_cluster_endpoints(self) -> str:
...
+        """Get the value of a tuning configuration option.
+
+        Args:
+            option (TuningOptions): The tuning option to get the value for.
+
+        Returns:
+            int: The value of the tuning option.
+        """
+        if option == TuningOptions.QUOTA_BACKEND_BYTES_CONFIG:
+            return self.quota_backend_bytes
```
reneradoi, line 274: Same here: We need to add error handling to avoid crashing if the call fails.

src/workload.py
```diff
@@ -191,3 +192,53 @@ def enable_database(self) -> None:
...
+        return 0
+
+    @override
+    def is_lxd_cloud(self) -> bool:
+        """Check if the workload is running in an LXD cloud environment.
+
+        Returns:
+            bool: True if running in LXD cloud, False otherwise.
+        """
+        # LXD sets up a Unix socket at /dev/lxd/sock inside the container
```
reneradoi, line 243: Can you please add a link to documentation if you have one? Just to be able to look it up if it ever changes.

reneradoi (commented):
Thanks Smail! My main concern is error handling, which we need to add in a few places to make sure this doesn't crash the charm.

src/managers/config.py
skourta: Yes, I'm working on it. Thank you @reneradoi

src/workload.py
```diff
@@ -191,3 +192,63 @@ def enable_database(self) -> None:
...
+        """
+        with open("/proc/meminfo") as f:
+            meminfo = f.read().split("\n")
+            meminfo = [line.split() for line in meminfo if line.strip()]
+
+        memory_sizes = {line[0][:-1]: float(line[1]) for line in meminfo}
+        return int(memory_sizes["MemTotal"] * 1024)  # convert from kB to Bytes
+
+    @override
+    def data_storage_size(self) -> float:
```
reneradoi: If I'm not mistaken, `shutil.disk_usage.total` should return an integer.

src/core/workload.py
```diff
@@ -261,3 +261,48 @@ def get_host_mapping(self) -> dict[str, str]:
...
+    def memory_size(self) -> float:
+        """Get the total memory size of the system in Bytes.
+
+        Returns:
+            float: The total memory size in Bytes.
+        """
+        pass
+
+    @abstractmethod
+    def data_storage_size(self) -> float:
```
reneradoi: Return types here should be `int`, applies for the other abstract methods, too.

src/managers/config.py
reneradoi: Looks good now.

Mehdi-Bendriss (dismissed):
Thanks Smail!

reneradoi (commented):
Just minor things to adjust regarding return types, rest is good.

src/workload.py
skourta: Thank you I updated all the types

Mehdi-Bendriss (approved):

reneradoi (approved):
Thank you Smail!

Issue #164, opened 2025-10-16
Charm can crash on invalid private key


Gu1nness:
## Steps to reproduce

1. juju add-secret tls-peer-private-key private-key="b'invalid'" # gives secret-id
2. juju grant-secret tls-peer-private-key etcd
3. juju config etcd tls-peer-private-key=<secret-id>

## Expected behavior

The charm displays a message for an invalid private key

## Actual behavior

Charm is in error on config-changed

## Versions

Operating system: 

Juju CLI: 

Juju agent: 

Charm revision: latest

## Additional context

[This](https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/src/managers/tls.py#L486-L490) logic will raise a UnicodeDecodeError on at least some invalid strings.
The logic here assumes that either the key is in plain text or is a valid base64 string, which might not be the case.

Prevent this with a try/catch for UnicodeDecodeError and binascii.Error so all possible errors are caught as invalid private key

reneradoi:
Has been fixed with #166.

PR #166, opened 2025-10-17
[DPE-8487] Improve error handling

reneradoi:
The following issues for stability and error handling have been addressed:
1) do not raise an uncaught error if the health check after restarting a unit fails
2) do not raise if getting the member information for updating the peer url fails
3) disaster recovery with `cluster-rebuild` action fails if peer TLS is enabled
4) charm can crash on invalid private key (#164)
5) do not raise if an object storage bucket/container is not accessible on backup/restore actions
6) do not raise if adding a managed user to the etcd cluster fails
7) do not raise uncaught errors if an external client provides an invalid mTLS certificate

## Solution:
These changes have been implemented:
1) Instead of raising, we set a `maintenance` status to display a message `unhealthy after restarting` to users. This applies to all restarts, for example TLS switchover, IP address update, CA rotation or config changes.
2) The error handling for `broadcast_peer_url` only catches the error, status display is handled by 1) mentioned above (which always follows).
3) Only replace `http` with `https` in config property `initial-cluster` if the config property actually exists (in cluster rebuild it does not).
4) Errors when decoding a provided private key for peer or client TLS are handled and no longer raise.
5) Handle the `EtcdBackupError` returned from the `BackupManager` class and fail the event
6) Handle the `EtcdUserManagementError` returned from the `ClusterManager` class and set a blocked status
7) Handle the `ValueError` returned from `is_leaf_certificate_valid` and the `TLSCertificatesError` in `get_common_name_from_chain` for invalid mTLS certificates

- [X] Manually
- [X] Unit tests
- [X] Integration tests

All described issues above have been tested manually, as well as the proposed fixes.

### DIFF
--- a/src/common/certificates.py
+++ b/src/common/certificates.py
@@ -34,8 +34,13 @@ def is_leaf_certificate_valid(mtls_cert: str) -> bool:
     Returns:
         (bool): True if the certificate is not a CA.
     """
-    leaf_cert = leaf_certificate(mtls_cert)
-    certificate = x509.load_pem_x509_certificate(data=leaf_cert.encode())
+    try:
+        leaf_cert = leaf_certificate(mtls_cert)
+        certificate = x509.load_pem_x509_certificate(data=leaf_cert.encode())
+    except ValueError as e:
+        logger.error(e)
+        return False
+
     # check if the certificate is a CA
     try:
         basic_constraints = certificate.extensions.get_extension_for_class(
@@ -43,6 +48,7 @@ def is_leaf_certificate_valid(mtls_cert: str) -> bool:
         ).value
     except x509.ExtensionNotFound:
         return False
+
     # check if the certificate can sign other certificates
     try:
         key_usage = certificate.extensions.get_extension_for_class(x509.KeyUsage).value
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -289,6 +289,12 @@ class EtcdEvents(Object):
                     component_name=self.charm.cluster_manager.name,
                     statuses_state=self.charm.state.statuses,
                 )
+            else:
+                self.charm.state.statuses.delete(
+                    ClusterStatuses.RESTART_FAILED.value,
+                    scope="unit",
+                    component=self.charm.cluster_manager.name,
+                )
 
             # update cluster configuration
             self.charm.cluster_manager.broadcast_peer_url(self.charm.state.unit_server.peer_url)
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -5,7 +5,7 @@
 """Etcd related and core event handlers."""
 
 import logging
-from subprocess import CalledProcessError
+from subprocess import CalledProcessError, TimeoutExpired
 from typing import TYPE_CHECKING
 
 import ops
@@ -99,7 +99,7 @@ class EtcdEvents(Object):
                 # fix the permissions of the directory if re-attaching existing storage
                 self.charm.workload.exec(["chmod", "-R", "750", path])
                 self.charm.workload.exec(["chown", "-R", f"{SNAP_USER}:{SNAP_GROUP}", path])
-            except CalledProcessError:
+            except (CalledProcessError, TimeoutExpired):
                 # gracefully continue if the path is not there yet
                 logger.warning("Could not adjust directory permissions")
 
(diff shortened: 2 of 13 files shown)


src/common/certificates.py
```diff
@@ -34,15 +34,21 @@ def is_leaf_certificate_valid(mtls_cert: str) -> bool:
     Returns:
         (bool): True if the certificate is not a CA.
     """
-    leaf_cert = leaf_certificate(mtls_cert)
-    certificate = x509.load_pem_x509_certificate(data=leaf_cert.encode())
+    try:
+        leaf_cert = leaf_certificate(mtls_cert)
+        certificate = x509.load_pem_x509_certificate(data=leaf_cert.encode())
+    except ValueError as e:
```
Mehdi-Bendriss, line 40: nice

src/managers/cluster.py
```diff
@@ -162,7 +161,10 @@ def broadcast_peer_url(self, peer_urls: str) -> None:
             password=self.admin_password,
             client_url=self.state.unit_server.client_url,
         )
-        client.broadcast_peer_url(self.member.id, peer_urls)
+        try:
+            client.broadcast_peer_url(self.member.id, peer_urls)
+        except ValueError as e:
```
Mehdi-Bendriss, line 166: when would this happen?

src/managers/external_clients.py
```diff
@@ -94,7 +94,11 @@ def get_common_name_from_chain(self, mtls_cert: str) -> str:
         raw_cas = mtls_cert.split("-----END CERTIFICATE-----")
         # add the marker back to the certificate
         cert = raw_cas[0].strip() + "\n-----END CERTIFICATE-----"
-        return Certificate.from_string(cert).common_name
+        try:
+            return Certificate.from_string(cert).common_name
+        except TLSCertificatesError as e:
+            logger.error(e)
+            return ""
```
Mehdi-Bendriss, line 101: Do we handle this in the caller side?

src/managers/tls.py
```diff
@@ -483,11 +484,16 @@ def read_and_validate_private_key(self, private_key_secret_id: str) -> PrivateKe
...
-        )
+        try:
+            private_key = (
+                secret_content
+                if re.match(r"(-+(BEGIN|END) [A-Z ]+-+)", secret_content)
+                else base64.b64decode(secret_content).decode("utf-8").strip()
+            )
+        except (UnicodeDecodeError, binascii.Error) as e:
+            logger.error(e)
+            return None
```
Mehdi-Bendriss, line 495: Similar question, do we handle the None on the caller side?

Mehdi-Bendriss (commented):
Thanks René! I have a couple of questions, but really good at scanning through the code for those

src/managers/cluster.py
reneradoi: It happens if the `member list` command does not return a result (see [here](https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/src/managers/cluster.py#L138-L139)). This can be because the client certificate is considered invalid if it contains the wrong IP address, for example (TLS switchover while IP address updated).

src/managers/external_clients.py
reneradoi: This logic is in all occurrences always preceded or directly followed by `if not is_leaf_certificate_valid()` and a return. That means, if `""` is returned from `get_common_name_from_chain()`, this has no impact on functionality, we only need to make sure it doesn't fail. I've checked these occurrences one by one.

src/managers/tls.py
reneradoi: The behavior is the same as in case of errors earlier in this function, for example if the secret can't be accessed or such. And yes, the `None` is handled in the places where it is called.

Mehdi-Bendriss (approved):

src/events/etcd.py
```diff
@@ -109,15 +108,7 @@ def _on_install(self, event: ops.InstallEvent) -> None:
         try:
             self.charm.workload.install()
         except EtcdServiceError:
-            self.charm.status.set_running_status(
```
skourta, line 112: Can you please clarify why you removed the status?

skourta (commented):
Thank you @reneradoi. I have one question

src/events/etcd.py
reneradoi: Sure: Right after, we raise. Any status set by us will not be displayed, because the unit is in `ErrorStatus` anyway. So no sense in adding a status here.

skourta (approved):

PR #167, opened 2025-10-20
test: rewrites tls tests to use jubilant

Soundarya03:
Addresses [this](https://warthogs.atlassian.net/issues/?jql=assignee%20%3D%20currentUser%28%29%20AND%20project%3D%27DPE%27%20AND%20statusCategory%20%21%3D%203&selectedIssue=DPE-7875) jira ticket to rewrite all TLS tests to use Jubilant. 

## Solution:
This PR rewrites all 4 files under the TLS integration tests folder to use Jubilant. It also introduces/rewrites several helper functions as necessary.

- [ ] Manually
- [ ] Unit tests
- [x] Integration tests


tests/integration/helpers.py
```diff
@@ -294,7 +295,7 @@ async def get_application_relation_data(
 
 
 def get_unit_relation_data(
-    juju: jubilant.Juju,
+    juju: Juju,
```
reneradoi, line 297: Let's decide on a way of doing this and then do it the same way consistently. In the previous PR we added `juju: jubilant.Juju` in a couple of helpers, we should either continue to do that or adjust the previously added ones to also use `juju: Juju`.

tests/integration/tls/test_ca_rotation.py
```diff
@@ -292,35 +279,27 @@ async def test_ca_rotation_by_expiration(ops_test: OpsTest) -> None:
...
-            APP_NAME: [
-                TLSStatuses.TLS_PEER_CERTS_EXPIRING.value,
-            ],
-            TLS_NAME: [
-                CharmStatuses.ACTIVE_IDLE.value,
-            ],
-        },
-    )
+
+    juju_lxd_model.wait(tls_certs_expiring)
```
reneradoi: Nice!

tests/integration/helpers_deployment.py
```diff
@@ -294,6 +299,38 @@ def _is_every_condition_on_units_met(
     return True
 
 
+def apps_active_and_agents_idle(
+    status: jubilant.Status,
+    *apps: str,
+    idle_period: int = 0,
+    unit_count: Union[int, Dict[str, int]] | None = None,
+) -> bool:
+    return (
```
reneradoi, line 324: Please don't forget about doc strings and comments.

reneradoi (commented):
Hey Soundarya! I've taken a high-level look at your approach, and I really like it. Looks like we can get away with not refactoring `wait_until`, that would be a great win and we could get rid of quite a few helpers.

I've left a few general comments, please always take care of them to ensure quality.

Please continue on that way, it seems really promising. Great work!

Soundarya03:
> Hey Soundarya! I've taken a high-level look at your approach, and I really like it. Looks like we can get away with not refactoring `wait_until`, that would be a great win and we could get rid of quite a few helpers.
> 
> I've left a few general comments, please always take care of them to ensure quality.
> 
> Please continue on that way, it seems really promising. Great work!

Thank you for giving this a look, Rene! 
Understood, have cleaned up the PR now. I'll make sure to keep these in mind going forward.

I'll ping y'all for the final review once all the tests are running fine and it is ready to merge from my end.

tests/integration/tls/test_ca_rotation.py
```diff
@@ -208,30 +209,27 @@ async def test_ca_rotation_by_config_change(ops_test: OpsTest) -> None:
     ), "Failed to read new key"
 
 
-async def _prepare_units_for_ca_expiration_test(ops_test: OpsTest) -> None:
+def _prepare_units_for_ca_expiration_test(juju: Juju) -> None:
     """Prepare the units for the CA expiration test."""
-    etcd_app: Application = ops_test.model.applications[APP_NAME]  # type: ignore
-    for unit in etcd_app.units:
-        logger.info("Updating renewal relative time to 0.6 for unit %s", unit.name)
+    for unit_name in juju.status().get_units(APP_NAME).keys():
```
skourta: You don't have to use `.keys()`

tests/integration/tls/test_ca_rotation.py
```diff
@@ -34,30 +32,33 @@
 
 
 @pytest.mark.abort_on_fail
-async def test_build_and_deploy_with_tls(charm: str, ops_test: OpsTest) -> None:
+async def test_build_and_deploy_with_tls(charm: str, juju_lxd_model: Juju) -> None:
```
skourta: Why are we using `juju_lxd_model` specifically instead of `juju`?

tests/integration/tls/test_ca_rotation.py
```diff
@@ -208,30 +209,27 @@ async def test_ca_rotation_by_config_change(ops_test: OpsTest) -> None:
     ), "Failed to read new key"
 
 
-async def _prepare_units_for_ca_expiration_test(ops_test: OpsTest) -> None:
+def _prepare_units_for_ca_expiration_test(juju: Juju) -> None:
```
skourta, line 216: Why do we have 2 models?

tests/integration/tls/test_private_key_rotation.py
```diff
@@ -43,38 +42,38 @@
...
 
 @pytest.mark.abort_on_fail
-async def test_build_and_deploy_with_tls(charm: str, ops_test: OpsTest) -> None:
+def test_build_and_deploy_with_tls(charm: str, juju_lxd_model: Juju) -> None:
     """Build the charm-under-test and deploy it with three units.
 
     The initial cluster should be formed and accessible.
     """
-    assert ops_test.model is not None, "Model is not set"
+    assert juju_lxd_model.model is not None, "Model is not set"
```
skourta: We can delete these.

tests/integration/helpers_deployment.py
```diff
@@ -294,6 +298,59 @@ def _is_every_condition_on_units_met(
...
+        jubilant.all_active(status, *apps)
+        and jubilant.all_agents_idle(status, *apps)
+        and all(
+            parse(unit.juju_status.since, ignoretz=True) + timedelta(seconds=idle_period)
+            < datetime.now()
+            for app in apps
+            for unit in status.get_units(app).values()
+        )
+        and (True if not unit_count else verify_unit_count(status, *apps, unit_count=unit_count))
```
skourta: nit. I usually prefer readability over oneliners. Personal opinion, I would refactor this into some variables to have it easily read.

tests/integration/helpers_deployment.py
```diff
@@ -294,6 +298,59 @@ def _is_every_condition_on_units_met(
...
+    )
+
+
+def verify_unit_count(status: jubilant.Status, *apps: str, unit_count: Union[int, Dict[str, int]]):
```
skourta: We use python 3.12 no need for the type helpers you can just use types directly. For Union you can use `|`

tests/integration/helpers_deployment.py
skourta: This goes for all other occurrences.

tests/integration/helpers_deployment.py
```diff
@@ -294,6 +298,59 @@ def _is_every_condition_on_units_met(
...
+        status: represents the jubilant model's current status
+        apps: A list of applications whose statuses to test against
+        unit_count: The desired number of units to wait for, can be >= to -1
+            if set as int, this value is expected for all apps but if more granularity is needed,
+            pass a dictionary such as: {"app1": 2, "app2": 1, ...}, if set to -1, the check
+            only happens at the application level.
+    """
+    if isinstance(unit_count, int):
+        unit_count = dict.fromkeys(apps, unit_count)
+    elif not unit_count:
```
skourta, line 361: Be careful of `unit_count = 0` case. We should be more specific.

tests/integration/tls/test_ca_rotation.py
```diff
@@ -292,35 +279,27 @@ async def test_ca_rotation_by_expiration(ops_test: OpsTest) -> None:
...
-            APP_NAME: [
-                TLSStatuses.TLS_PEER_CERTS_EXPIRING.value,
-            ],
-            TLS_NAME: [
-                CharmStatuses.ACTIVE_IDLE.value,
-            ],
-        },
-    )
+
+    juju_lxd_model.wait(tls_peer_certs_expiring)
```
skourta: This is where the previous `wait_unitil` shone. There will be multiple cases like this, I believe.

skourta (commented):
Thank you for the great work @Soundarya03.
I left some comments.

tests/integration/tls/test_ca_rotation.py
Soundarya03: `juju_lxd_model` is a pytest fixture that sets up a jubilant Juju model on LXD. It's named this way because we've got another fixture, `juju_k8s_model` defined for a K8s model (which was introduced because we needed a K8s model on which to deploy observability charms for the metrics int tests introduced in the previous PR).

tests/integration/tls/test_ca_rotation.py
Soundarya03: This is only a helper function, so it doesn't actually make use of the fixture. It's invoked by a test, with the actual model in use being passed to it. So it's merely an arg name here

tests/integration/helpers_deployment.py
Soundarya03: Thanks for pointing this out!

tests/integration/tls/test_ca_rotation.py
Soundarya03: Yes, there are multiple cases; but in fact I've extracted this to a method, so there should be less repeated code now

image
image

image

tests/integration/tls/test_ca_rotation.py
skourta: We also have a `juju` model that is not specific for any cloud. See [here](https://github.com/canonical/charmed-etcd-operator/blob/0558d4782015da5e7754abfc9d4fa7e046317609/tests/integration/conftest.py#L40).

reneradoi (approved):
I'm fine with the changes, thank you @Soundarya03!

tests/integration/tls/test_ca_rotation.py
skourta: Alright, so we should follow this pattern to replace all `wait_until` usage.
WDYT @Mehdi-Bendriss ?

skourta (approved):
Thank you @Soundarya03

reneradoi:
This PR can be merged into the feature branch, there is no issue that has to be fixed, only minor comments that can be reworked later.

PR #168, opened 2025-10-20
[DPE-8606] - Add profile config option

skourta:

This pull request introduces a new `profile` configuration option to tune etcd deployment for either `production` or `testing` environments. It enforces profile validity, adjusts configuration parameters based on the selected profile, and ensures that changes to the profile trigger appropriate configuration updates and restarts. The tests have been updated to cover these new behaviors.

## Solution:

**Profile configuration and validation:**
* Added a `profile` option to `config.yaml`, allowing users to select either "production" or "testing" deployment profiles.
* Introduced `Profile` enum and validation logic to ensure only valid profile values are accepted; invalid values block the unit and surface a clear status message. [[1]](diffhunk://#diff-959ba637043ab4ac88c6f9f9109022f78a7df1b9a2ab3a0407bff4f124c2db64R118-R124) [[2]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cL164-R192) [[3]](diffhunk://#diff-074beea58df119261178a25e474cd0c240fa20f1c6fb6990867d357676e96cefR95-R98)

**Configuration tuning based on profile:**
* In production profile, automatically sets `quota-backend-bytes` to 8GiB; in testing profile, removes this parameter from the config. [[1]](diffhunk://#diff-959ba637043ab4ac88c6f9f9109022f78a7df1b9a2ab3a0407bff4f124c2db64R55-R56) [[2]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cR101-R108)
* Ensures that changes to the profile or related config parameters trigger a rolling restart when necessary. [[1]](diffhunk://#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cR210-R226) [[2]](diffhunk://#diff-4d4b45129d72bf79a08e70009a59f4ebe0a016b4727117155249c864ae4a1b49R310)

**Event handling and startup logic:**
* Defers start events if tuning parameters or profile are invalid, logging errors for easier troubleshooting.

**Testing improvements:**
* Added and updated unit tests to verify correct behavior for profile changes, including transitions between profiles and handling of invalid profiles. [[1]](diffhunk://#diff-0580ab9aef31b44580478eb270ec3422c5e6fef35df585fecb3a68317e246809L615-R619) [[2]](diffhunk://#diff-0580ab9aef31b44580478eb270ec3422c5e6fef35df585fecb3a68317e246809L634-R639) [[3]](diffhunk://#diff-0580ab9aef31b44580478eb270ec3422c5e6fef35df585fecb3a68317e246809L648-R657) [[4]](diffhunk://#diff-0580ab9aef31b44580478eb270ec3422c5e6fef35df585fecb3a68317e246809L667-R680) [[5]](diffhunk://#diff-0580ab9aef31b44580478eb270ec3422c5e6fef35df585fecb3a68317e246809L686-R703) [[6]](diffhunk://#diff-0580ab9aef31b44580478eb270ec3422c5e6fef35df585fecb3a68317e246809L705-R726) [[7]](diffhunk://#diff-0580ab9aef31b44580478eb270ec3422c5e6fef35df585fecb3a68317e246809L728-R753) [[8]](diffhunk://#diff-0580ab9aef31b44580478eb270ec3422c5e6fef35df585fecb3a68317e246809R767-R847)

- [x] Manually
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/src/events/etcd.py
+++ b/src/events/etcd.py
@@ -139,6 +139,11 @@ class EtcdEvents(Object):
             event.defer()
             return
 
+        if not self.charm.config_manager.are_tuning_parameters_valid():
+            logger.error("Deferring start because of invalid tuning parameters.")
+            event.defer()
+            return
+
         self.charm.config_manager.set_config_properties()
 
         if not self.charm.state.cluster.cluster_state and self.charm.unit.is_leader():
@@ -302,6 +307,7 @@ class EtcdEvents(Object):
         if (
             self.charm.config_manager.are_tuning_parameters_valid()
             and self.charm.config_manager.requires_restart()
+            and self.charm.state.unit_server.is_started
         ):
             # apply config and initiate restart
             self.charm.rolling_restart()
--- a/src/managers/config.py
+++ b/src/managers/config.py
@@ -19,6 +19,9 @@ from literals import (
     CONFIG_FILE,
     DATABASE_DIR,
     METRICS_PORT,
+    PRODUCTION_QUOTA_BACKEND_BYTES,
+    QUOTA_BACKEND_BYTES,
+    Profile,
     RestoreStep,
     TLSState,
     TuningOptions,
@@ -95,6 +98,14 @@ class ConfigManager(ManagerStatusProtocol):
                 # take over the config value set by users
                 config_properties[option.value] = self.config.get(option.value)
 
+            config_profile = Profile(self.config.get("profile"))
+            if config_profile == Profile.PRODUCTION:
+                config_properties[QUOTA_BACKEND_BYTES] = PRODUCTION_QUOTA_BACKEND_BYTES
+            else:
+                # remove quota-backend-bytes from config for testing profile
+                if QUOTA_BACKEND_BYTES in config_properties:
+                    del config_properties[QUOTA_BACKEND_BYTES]
+
         if self.state.unit_server.tls_client_state in [TLSState.TO_TLS, TLSState.TLS]:
             # replace http with https in listen-client-urls and advertise-client-urls
             config_properties["listen-client-urls"] = self.state.unit_server.client_url.replace(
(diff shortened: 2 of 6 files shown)


src/events/etcd.py
```diff
@@ -139,6 +139,11 @@ def _on_start(self, event: ops.StartEvent) -> None:  # noqa: C901
             event.defer()
             return
 
+        if not self.charm.config_manager.are_tuning_parameters_valid():
+            logger.error("Deferring start because of invalid tuning parameters.")
+            event.defer()
+            return
+
         self.charm.config_manager.set_config_properties()
```
reneradoi, line 147: Let's not do that here, it's not required to. None of the current tuning options prevents us from starting. We could instead just start regularly and then set status to `blocked` (as already happening).

With deferring the start up of a cluster, this also leads to deferring all `peer_relation_changed` events which are used to initially assemble the cluster members. This causes a long list of unnecessarily deferred events.

reneradoi (commented):
Thank you Smail! The PR looks good, I'd only like to get rid of the deferred start. Other than that, no issues from my side.

src/managers/config.py
```diff
@@ -95,6 +98,14 @@ def config_properties(self) -> str:
                 # take over the config value set by users
                 config_properties[option.value] = self.config.get(option.value)
 
+            config_profile = Profile(self.config.get("profile"))
+            if config_profile == Profile.PRODUCTION:
+                config_properties[QUOTA_BACKEND_BYTES] = PRODUCTION_QUOTA_BACKEND_BYTES
+            else:
+                # remove quota-backend-bytes from config for testing profile
+                if QUOTA_BACKEND_BYTES in config_properties:
+                    del config_properties[QUOTA_BACKEND_BYTES]
```
Mehdi-Bendriss: You can avoid this by using `config_properties.pop(QUOTA_BACKEND_BYTES, None)`

src/statuses.py
```diff
@@ -92,6 +92,10 @@ class ConfigStatuses(Enum):
         status="blocked",
         message="Invalid values set for the config options: 'election-timeout', 'heartbeat-interval'",
     )
+    PROFILE_INVALID = StatusObject(
+        status="blocked",
+        message="Invalid profile configuration option. Only “production” and “testing” values are allowed",
```
Mehdi-Bendriss: for consistency:
```suggestion
        message="Invalid profile configuration option. Only 'production' and 'testing' values are allowed",
```

src/events/etcd.py
Mehdi-Bendriss: i'm a bit torn here, as if the user from the get go set the `profuction` (typo) profile, and we're unable to fulfill it, why start the service at all with the default value that the user clearly doesn't want?

Mehdi-Bendriss (commented):
Thanks Smail! Left a couple of comments

src/events/etcd.py
reneradoi: There's only one reason from my side: To avoid the deferral loop.

skourta: So in the case where the user deploys with typos we do not add at all the tuning parameters and just use etcd's defaults? I think that simplifies the charm start procedure and avoids the deferral loop but might start etcd with an unwanted configuration. The impact is minimal now because we do not set a lot of config options but as soon as we blow up the profile implementation we will have to add this back.

I think we can remove it for now, but we must add it back when we update the profiles implementation. WDYT?

reneradoi: Agree, once we introduce a configuration for the profiles which requires us not to start the units, then we should bring it back. Until then, we can follow YAGNI and keep it simple.

reneradoi:
I will close this PR as it is being replaced by https://github.com/canonical/charmed-etcd-operator/pull/165.

PR #170, opened 2025-10-23
unpin juju agent

reneradoi:
Because of a regression introduced in Juju release 3.6.10 (see https://github.com/juju/juju/issues/20714). we had pinned the juju-agent version to 3.6.9 when bootstrapping LXD clouds for integration tests. Now that Juju 3.6.11 has been released, we can revert that workaround.

## Solution:
unpin bootstrap agent

- [ ] Manually
- [ ] Unit tests
- [X] Integration tests

### DIFF
--- a/tests/integration/backup/test_s3_backup_restore.py
+++ b/tests/integration/backup/test_s3_backup_restore.py
@@ -141,6 +141,8 @@ async def test_restore_backup_on_different_cluster(charm: str, ops_test: OpsTest
     """Restore a backup and check if data is recovered."""
     logger.info("Remove existing etcd cluster and deploy a new one.")
     await ops_test.model.remove_application(APP_NAME, block_until_done=True)
+    await ops_test.model.remove_secret(secret_name="system_users_secret")
+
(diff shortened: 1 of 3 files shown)


tests/integration/backup/test_s3_backup_restore.py
```diff
@@ -141,6 +141,8 @@ async def test_restore_backup_on_different_cluster(charm: str, ops_test: OpsTest
     """Restore a backup and check if data is recovered."""
     logger.info("Remove existing etcd cluster and deploy a new one.")
     await ops_test.model.remove_application(APP_NAME, block_until_done=True)
+    await ops_test.model.remove_secret(secret_name="system_users_secret")
+
```
reneradoi, line 145: This test was failing both on CI as well as locally, probably because of a change to the secrets engine in Juju. The output of `list secrets` seems different now, and in order to avoid "accidental behavior", I decided to explicitly clean up the previous test setup.

This does not have real-world impact, the functionality itself works correctly.

Mehdi-Bendriss (approved):

skourta (approved):

PR #171, opened 2025-10-23
[DPE-8678] Duplicate private key as fallback

reneradoi:
If a user-provided private key has been set for TLS client or peer certificates and that private key is updated to an invalid key, the next time the certificates are refreshed, it will be done with an auto-generated key (generated by the TLS library). This is not the desired behavior, instead the previously valid private key should be kept and continue to be used.

## Solution:
The PR changes the implementation to duplicate user-provided private keys into the internal application databag. If an the private key gets updated to an invalid one, the previous key will be kept and certificates are not updated. Only if no previously valid private key is available, an auto-generated one will be used.

- [X] Manually
- [ ] Unit tests
- [X] Integration tests

Manual tests:
1)
- enable peer TLS with auto-generated key
- provide a valid private key and check updated certs
- update to an invalid private key and check certs have not been updated
- update to a valid private key again and check certs have been updated

2)
- configure an invalid private key for client TLS
- enable client TLS and ensure an auto-generated key has been used
- update the user-provided key to a valid private key and check certs have been updated

### DIFF
--- a/src/core/models.py
+++ b/src/core/models.py
@@ -13,6 +13,7 @@ from charms.data_platform_libs.v0.data_interfaces import (
     DataPeerData,
     DataPeerUnitData,
 )
+from charms.tls_certificates_interface.v4.tls_certificates import PrivateKey
 from ops.model import Application, Relation, Unit
 
 from literals import (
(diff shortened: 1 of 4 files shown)


src/events/tls.py
```diff
@@ -82,34 +82,16 @@ def __init__(self, charm: "EtcdOperatorCharm"):
                     peer_private_key_id
                 )
             ) is None:
-                self.charm.state.statuses.add(
-                    TLSStatuses.TLS_INVALID_PRIVATE_KEY.value,
-                    scope="unit",
-                    component=self.charm.tls_manager.name,
-                )
```
Mehdi-Bendriss, line 89: we still need to add a blocked status stating that we did not proceed with the private key change due to it being invald

Mehdi-Bendriss (commented):
Thanks René! I have a comment

src/events/tls.py
reneradoi: Yes we do. It has already been there in the TLS manager, see [here](https://github.com/canonical/charmed-etcd-operator/blob/duplicate-private-key/src/managers/tls.py#L532-L539). It was a duplicate where I removed it, from my point of view.

Mehdi-Bendriss (dismissed):

src/core/models.py
```diff
@@ -220,6 +221,24 @@ def internal_user_credentials(self) -> dict[str, str]:
 
         return {}
 
+    @property
+    def tls_client_private_key(self) -> dict[str, PrivateKey]:
+        """Retrieve the private key for client TLS."""
+        if private_key := self.relation_data.get("tls-client-private-key"):
+            private_key = PrivateKey(raw=private_key)
+            return {"key": private_key}
```
skourta: Why are you using a dict?

reneradoi: Good find! The original idea was to make use of `None` as default value with `get` on the dict, but this can of course be achieved otherwise as well.

Mehdi-Bendriss (approved):

skourta (approved):
Thank you @reneradoi

PR #174, opened 2025-10-24
[DPE-8405] Improve integration test stability

reneradoi:
With the recent intensity of integration testing, we have experienced some failures on a few tests because of the test flow.

## Solution:
- disaster recovery test: the `update_status` event needs to happen to remove cluster members that are gone
- upgrades tests: no need to wait for application removal at the end of the tests

- [ ] Manually
- [ ] Unit tests
- [X] Integration tests

### DIFF
--- a/tests/integration/ha/test_disaster_recovery.py
+++ b/tests/integration/ha/test_disaster_recovery.py
@@ -54,15 +54,15 @@ async def test_membership_reconfiguration_after_unit_loss(ops_test: OpsTest) -> 
     destroy_unit_cmd = f"remove-unit {unit_to_remove.name} --model={ops_test.model.info.name} --force --no-wait --no-prompt"
     return_code, _, _ = await ops_test.juju(*destroy_unit_cmd.split())
     assert return_code == 0, "Failed to remove unit"
-    await wait_until(ops_test, apps=[APP_NAME], wait_for_exact_units=NUM_UNITS - 1)
+
+    # wait for the next `update_status` for the cluster membership to be updated
+    async with ops_test.fast_forward("15s"):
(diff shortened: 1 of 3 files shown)


tests/integration/ha/test_disaster_recovery.py
```diff
@@ -54,17 +54,15 @@ async def test_membership_reconfiguration_after_unit_loss(ops_test: OpsTest) ->
     destroy_unit_cmd = f"remove-unit {unit_to_remove.name} --model={ops_test.model.info.name} --force --no-wait --no-prompt"
     return_code, _, _ = await ops_test.juju(*destroy_unit_cmd.split())
     assert return_code == 0, "Failed to remove unit"
-    await wait_until(ops_test, apps=[APP_NAME], wait_for_exact_units=NUM_UNITS - 1)
+
+    # wait for the next `update_status` for the cluster membership to be updated
+    async with ops_test.fast_forward("15s"):
+        await wait_until(ops_test, apps=[APP_NAME], wait_for_exact_units=NUM_UNITS - 1)
```
skourta, line 60: Is there anyway we can avoid waiting for `update_status` as the hook can be deactivated by users or set for a very long duration.

reneradoi: Not with the current implementation.  Cleaning up cluster members happens in `remove_inconsistent_members_if_required`, which is currently run only on `update_status` (same as `clean_users`).

skourta: Yes, that is aligned with my understanding. We should consider this in the future.
Thank you @reneradoi

skourta (approved):

Mehdi-Bendriss (approved):

PR #182, opened 2025-11-13
[DPE-8611]: bulk event handling

skourta:

This pull request introduces significant improvements to how external client users are managed and handled in the codebase, with a focus on supporting bulk resource requests and more granular tracking of managed users. The changes refactor the user management logic to use composite keys (relation ID and request ID), add bulk event handling, and improve the robustness and clarity of user lifecycle management. Additionally, several integration test action names have been updated for consistency.

## Solution:

**User Management Refactoring:**

* Changed the `managed_users` field in `PeerAppModel` from `dict[int, str]` to `dict[str, str]`, allowing managed users to be tracked using composite keys of relation ID and request ID for more granular management.
* Refactored methods in `ExternalClientsManager` (`add_managed_user`, `remove_managed_user`, `get_relation_managed_user`) to use composite keys, supporting multiple requests per relation and enabling precise user tracking and removal.

**Bulk Resource Request Handling:**

* Added support for bulk resource requests via the new `BulkResourcesRequestedEvent`, updating event handler registration and implementing `_on_bulk_resources_requested` to handle multiple requests efficiently and safely. [[1]](diffhunk://#diff-41d3b0e68fba8bc44a83df27841cc482ba1df8a5cfe8e476397765673f46f446R16) [[2]](diffhunk://#diff-41d3b0e68fba8bc44a83df27841cc482ba1df8a5cfe8e476397765673f46f446L48-R53) [[3]](diffhunk://#diff-41d3b0e68fba8bc44a83df27841cc482ba1df8a5cfe8e476397765673f46f446R67-R149)

**User Lifecycle and Robustness:**

* Centralized user creation and removal logic into `_add_user` and `_remove_user` helper methods, improving error handling and ensuring proper cleanup of old users during updates and relation breaks. [[1]](diffhunk://#diff-41d3b0e68fba8bc44a83df27841cc482ba1df8a5cfe8e476397765673f46f446L139-R179) [[2]](diffhunk://#diff-41d3b0e68fba8bc44a83df27841cc482ba1df8a5cfe8e476397765673f46f446L159-R231) [[3]](diffhunk://#diff-41d3b0e68fba8bc44a83df27841cc482ba1df8a5cfe8e476397765673f46f446L258-R301) [[4]](diffhunk://#diff-41d3b0e68fba8bc44a83df27841cc482ba1df8a5cfe8e476397765673f46f446R359-R434)
* Improved resource response generation to consistently include all necessary metadata (username, endpoints, URIs, CA, version) for both single and bulk requests.

**Integration Test Improvements:**

* Renamed integration test actions for clarity and consistency (e.g., `update-mtls-cert` → `update-mtls-certs`, `get-certificate` → `get-certificates`). [[1]](diffhunk://#diff-e68021f0fa03c0dce06764f557169d5fb85fa747715facf18ada1d4fbecc8a5bL39-R40) [[2]](diffhunk://#diff-e68021f0fa03c0dce06764f557169d5fb85fa747715facf18ada1d4fbecc8a5bL66-R63)

- [x] Manually
- [x] Unit tests
- [x] Integration tests


src/events/external_clients.py
```diff
@@ -322,3 +360,79 @@ def _update_client_truststore(self) -> None:
             logger.debug("CAs have changed, updating client truststore")
             self.charm.tls_manager.update_cas(all_cas, TLSType.CLIENT)
             self.charm.rolling_restart()
+
+    def _exists_preventing_reason(self) -> tuple[bool, bool]:
```
reneradoi: Seems like it always returns `True, True` if there is any preventing reason -> we can get rid of the `should_defer` part if it is redundant.

src/events/external_clients.py
```diff
@@ -59,58 +64,93 @@ def __init__(self, charm: "EtcdOperatorCharm"):
...
+                continue
 
-        if (
-            self.charm.state.unit_server.tls_client_ca_rotation_state
-            != TLSCARotationState.NO_ROTATION
-        ):
-            logger.debug("CA rotation is in progress")
-            event.defer()
+            if self.charm.cluster_manager.get_user(common_name) is not None:
+                logger.error("User already exists in database for another request")
```
reneradoi, line 111: Should we add it to the invalid requests in this case? It could be that the user has been requested by a different client relation, and thus this client relation cannot use it.

src/events/external_clients.py
```diff
@@ -156,56 +191,48 @@ def _on_mtls_cert_updated(self, event: MtlsCertUpdatedEvent[RequirerCommonModel]
...
-                        )
-                    except EtcdUserManagementError as e:
-                        self.charm.status.set_running_status(
-                            ExternalClientsStatuses.EC_USER_MANAGEMENT_ERROR.value,
-                            scope="app",
-                            component_name=self.charm.external_clients_manager.name,
-                            statuses_state=self.charm.state.statuses,
-                        )
-                        logger.error(e)
+                    if not self._add_user(common_name, event.request, event.relation.id):
```
reneradoi, line 192: Same here: Should we add it to the invalid requests if the user could not be added?

src/managers/external_clients.py
```diff
@@ -42,47 +42,64 @@ def __init__(
         self.workload = workload
         self.substrate = substrate
 
-    def remove_managed_user(self, relation_id: int) -> None:
+    def _generate_key(self, relation_id: int, request_id: str | None) -> str:
```
reneradoi: I'd recommend to rename for easier understanding, as this method does not actually generate anything. Maybe better `_resolve_key` or `_compile_key`?

src/events/external_clients.py
```diff
@@ -322,3 +360,79 @@ def _update_client_truststore(self) -> None:
...
+            logger.debug("CA rotation is in progress")
+            return True, True
+
+        if not self.charm.state.cluster.auth_enabled:
+            logger.error("Cluster authentication is not enabled")
+            return True, True
+
+        return False, False
+
+    def _add_user(
```
reneradoi, line 410: I like that you have extracted this and the `_exists_preventing_reason` into separate methods!

reneradoi (commented):
Thank you Smail for adding this implementation!

Besides the few comments on the code itself, I do have two questions related to testing:
1) If I understand correctly, we do not have integration test coverage anymore for the case where a client doesn't send a bulk request? How can we make sure we don't break that? I assume this will be the majority of use cases and bulk requests will be rather rare.

2) Did you manually test compatibility, as in: create a client relation with the previous implementation, then upgrade the charm code to this implementation and check whether the managed users can still be processed correctly, even after rolling back?

src/events/external_clients.py
skourta: I thought about it and I agree.

src/events/external_clients.py
skourta: We can do that when we have error propagation. We only have one request at a time for the `mtls_cert_updated` event.

skourta (commented):
@reneradoi 
The answers to the questions:
1. Bulk is on the etcd side and not the requerier side. But we do have v0 integration tests where bulk events are not fired and the previous mtls_cert_changed event is fired instead. That is why I did not remove the user creation logic from there.
2. Yes, I did. I testing with simple pydantic models then with refreshing the charm to local from stable and back. As long as no multiple users on the same relation you can downgrade without issue.

reneradoi:
Regarding failures of the S3-related tests on arm: I've created a PR contributing arm-support to the object-storage-integrators. Progress can be tracked [here](https://github.com/canonical/object-storage-integrators/pull/71). As soon as they have been published, our tests should pass.

src/events/external_clients.py
```diff
@@ -59,58 +64,92 @@ def __init__(self, charm: "EtcdOperatorCharm"):
...
-            or self.charm.state.cluster.rebuild_cluster_in_progress
-            or self.charm.refresh_in_progress
-        ):
-            logger.warning(
-                "Cannot update certificates while cluster is in vulnerable state because of restore, refresh or cluster-rebuild"
+        invalid_requests = []
+
+        responses = []
+        for request in event.requests:
+            if request.mtls_cert is None or request.resource is None:
```
Mehdi-Bendriss: nit:
```suggestion
            if not request.mtls_cert or not request.resource:
```

src/events/external_clients.py
```diff
@@ -59,58 +64,92 @@ def __init__(self, charm: "EtcdOperatorCharm"):
...
-            return
+            relation_managed_user = self.charm.external_clients_manager.get_relation_managed_user(
+                event.relation.id, request.request_id
+            )
 
-        if self.charm.state.unit_server.tls_client_state == TLSState.TO_TLS:
-            logger.error("TLS is not ready")
-            event.defer()
-            return
+            if relation_managed_user is not None and relation_managed_user == common_name:
```
Mehdi-Bendriss: nit:
```suggestion
            if relation_managed_user and relation_managed_user == common_name:
```

src/events/external_clients.py
```diff
@@ -59,58 +64,92 @@ def __init__(self, charm: "EtcdOperatorCharm"):
...
 
-    def _on_mtls_cert_updated(self, event: MtlsCertUpdatedEvent[RequirerCommonModel]) -> None:  # noqa: C901
-        """Handle the ca chain updated event."""
-        if not self.charm.state.cluster.model or not self.charm.state.unit_server.model:
-            logger.error("peer data not available")
+    def _on_bulk_resources_requested(  # noqa: C901
+        self, event: BulkResourcesRequestedEvent[RequirerCommonModel]
+    ) -> None:
+        """Handle bulk resources requested event."""
+        prevent_reason = self._exists_preventing_reason()
```
Mehdi-Bendriss:
```suggestion
        preventing_reason = self._exists_preventing_reason()
```

(also in subsequent usages and definitions)

src/events/external_clients.py
```diff
@@ -59,58 +64,92 @@ def __init__(self, charm: "EtcdOperatorCharm"):
...
+    def _on_mtls_cert_updated(self, event: MtlsCertUpdatedEvent[RequirerCommonModel]) -> None:  # noqa: C901
+        """Handle the ca chain updated event."""
+        if not event.request.mtls_cert or not event.request.resource:
+            logger.error("CA chain, keys prefix, or common name not provided")
             return
 
-        if not self.charm.state.cluster.auth_enabled:
-            logger.error("Cluster authentication is not enabled")
+        prevent_reason = self._exists_preventing_reason()
+        if prevent_reason:
```
Mehdi-Bendriss: nit, walrus and `preventing_reason`

src/events/external_clients.py
```diff
@@ -156,56 +190,48 @@ def _on_mtls_cert_updated(self, event: MtlsCertUpdatedEvent[RequirerCommonModel]
...
+
         self.charm.state.statuses.delete(
             ExternalClientsStatuses.EC_USER_MANAGEMENT_ERROR.value,
             scope="app",
             component=self.charm.external_clients_manager.name,
         )
 
+    def _remove_user(
+        self, old_common_name: str, relation_id: int, request_id: str | None = None
+    ) -> None:
```
Mehdi-Bendriss, line 224: missing docstrings

src/managers/external_clients.py
```diff
@@ -42,47 +42,64 @@ def __init__(
         self.workload = workload
         self.substrate = substrate
 
-    def remove_managed_user(self, relation_id: int) -> None:
+    def _resolve_key(self, relation_id: int, request_id: str | None) -> str:
```
Mehdi-Bendriss: we are not exactly resolving a key here, we are constructing it, we also probably wanna add the `_managed_` word in the method name

Mehdi-Bendriss (commented):
Thanks Smail! I only have a few nits

reneradoi (approved):
All changes I'd like to see are there now, thank you!

Mehdi-Bendriss (approved):

Issue #189, opened 2025-11-27
Secret label causes ModelError when multiple charms share the same TLS certificates provider


mateoflorido:
## Steps to reproduce

1. Deploy `charmed-etcd` charm
2. Deploy another charm (e.g., `k8s` charm)
3. Deploy a single TLS provider (e.g, `self-signed-certificates`)
4. Relate both charms to the same provider
5. After the whole model is `active/idle`, perform a CA refresh (e.g., `juju config ssc ca-common-name="my-little-ca"`).

## Expected behavior

Both charms should successfully receive their TLS certificates after the CA refresh and reach `active/idle` status when sharing the same `tls-certificates` provider.

## Actual behavior
The `charmed-etcd` charm crashes with a `ModelError` when processing `secret-changed` events:
```
ops.model.ModelError: ERROR getting latest secret revision: secret label "etcd-client.10.2e72e7f4350be731.mtls.secret" for consumer "unit-charmed-etcd-0" already exists
```

## Versions

Operating system: Ubuntu 24.04

Juju CLI: 3.6.12-genericlinux-amd64

Juju agent: 3.6.11

Charm revision: 3.6/stable (120)

## Log output

Juju debug log: 

```
unit-charmed-etcd-0: 16:51:21 INFO unit.charmed-etcd/0.juju-log root:Received secret etcd-client.10.2e72e7f4350be731.mtls.secret but couldn't parse, seems i
unit-charmed-etcd-0: 16:51:21 ERROR unit.charmed-etcd/0.juju-log root:Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-charmed-etcd-0/charm/venv/lib/python3.12/site-packages/ops/model.py", line 3637, in _run
    result = subprocess.run(args, **kwargs)  # type: ignore
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '('/var/lib/juju/tools/unit-charmed-etcd-0/secret-get', 'secret:d4k86tpekrdpteeqk47g', '--label', 'etcd-client.10.2e7

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-charmed-etcd-0/charm/src/charm.py", line 415, in <module>
[...]

PR #188, opened 2025-11-26
Add pydantic serialization exception for update client relations

skourta:
## Description of issue or feature:
Fixes #187 
This pull request introduces improved error handling when updating client relations data in `src/events/etcd.py`. It adds support for catching serialization errors and logs them as errors rather than warnings.

## Solution:

* Added import of `PydanticSerializationError` from `pydantic_core` to handle serialization exceptions that may occur during client relations updates.
* Updated the `_update_client_relations` method to catch both `KeyError` and `PydanticSerializationError`, and changed the log level from warning to error for these exceptions.
* [Created an issue for the problem on data_interfaces](https://github.com/canonical/data-platform-libs/issues/251).

## How was this change tested?
- [ ] Manually
- [ ] Unit tests
- [ ] Integration tests

PR #190, opened 2025-12-01
[DPE-8989] Bugfix: allow same TLS-provider as client application

reneradoi:
This PR addresses #189 

If charmed-etcd and a client application share the same TLS provider and a CA-rotation happens, there are multiple issues because of the interfering events:

1) Interfering access from client and etcd on the `mtls-cert` secret if the client uses data-interfaces v0
This results in etcd running into `hook failed: "secret-changed"` with the following error:
```
ops.model.ModelError: ERROR getting latest secret revision: secret label "etcd-client.10.2e72e7f4350be731.mtls.secret" for consumer "unit-charmed-etcd-0" already exists
```
This is possibly a data-interfaces v1 issue (see https://github.com/canonical/data-platform-libs/issues/252), or even a Juju issue.

2) Client CA not recognized as a "new" CA
A unit of etcd might receive the renewed CA first from the client application before it receives it from the TLS provider. It will add the new CA to its client truststore already. When the "new" CA is received in the  `certificate_available` event, it will not be recognized as such, and the unit will therefore not participate in the CA rotation of the cluster. The CA rotation then gets stuck and never completes.

3) Restarts interfere with each other
If multiple rolling restarts are requested at the same time, rolling ops will only send one callback per unit. That means, if etcd requests a restart because of CA rotation and, at the same time, a new client application CA also requires/requests a restart, one of the restart requests gets lost. This is known behavior, see [rolling_ops](https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/lib/charms/rolling_ops/v0/rollingops.py#L146).

## Solution:
1) Temporary fix for the `secret_changed` event handler in data-interfaces v1 until issue https://github.com/canonical/data-platform-libs/issues/252 is resolved.

2) Adjustment to the `is_new_ca_saved_on_all_servers` check in the `TLSManager`: units with no CA rotation are ignored (in case of client TLS).

3) Consider the restart request issued when updating the client mtls-certs a "ca_rotation_restart", because it also loads the truststore and can continue the ca-rotation workflow if it happens at the same time.

## Drive-by changes:
In the fashion of #188: add temporary workaround for Pydantic serialization issues. These can also occur when updating client relation data because of changes on TLS. This workaround will be removed when https://github.com/canonical/data-platform-libs/issues/251 is fixed.

- [X] Manually
- [X] Unit tests
- [X] Integration tests

Integration test coverage has been adjusted: Client applications now use the same TLS provider as etcd to ensure a deployment scenario close to possible production setup.

### DIFF
--- a/lib/charms/data_platform_libs/v1/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v1/data_interfaces.py
@@ -2400,12 +2400,7 @@ class ResourceProviderEventHandler(EventHandlers, Generic[TRequirerCommonModel])
         repository = OpsRelationRepository(self.model, relation, component=relation.app)
         version = repository.get_field("version") or "v0"
 
-        try:
-            old_mtls_cert = event.secret.get_content().get("mtls-cert")
-        except ModelError as e:
-            logging.warning("Could not retrieve old mtls-cert: %s", e)
-            old_mtls_cert = None
-
+        old_mtls_cert = event.secret.get_content(refresh=True).get("mtls-cert")
         logger.info("mtls-cert-updated")
 
         # V0, just fire the event.
--- a/lib/charms/data_platform_libs/v1/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v1/data_interfaces.py
@@ -2400,7 +2400,12 @@ class ResourceProviderEventHandler(EventHandlers, Generic[TRequirerCommonModel])
         repository = OpsRelationRepository(self.model, relation, component=relation.app)
         version = repository.get_field("version") or "v0"
 
-        old_mtls_cert = event.secret.get_content().get("mtls-cert")
+        try:
+            old_mtls_cert = event.secret.get_content().get("mtls-cert")
+        except ModelError as e:
+            logging.warning("Could not retrieve old mtls-cert: %s", e)
+            old_mtls_cert = None
+
         logger.info("mtls-cert-updated")
 
         # V0, just fire the event.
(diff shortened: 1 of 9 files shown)


delgod (dismissed):

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -2400,7 +2400,12 @@ def _on_secret_changed_event(self, event: SecretChangedEvent) -> None:
         repository = OpsRelationRepository(self.model, relation, component=relation.app)
         version = repository.get_field("version") or "v0"
 
-        old_mtls_cert = event.secret.get_content().get("mtls-cert")
+        try:
```
skourta, line 2403: Should we deactivate the check_libraries workflow? it will fail if we have a locally modified version of the lib

tests/integration/client_relations/test_client_relations.py
```diff
@@ -122,7 +122,7 @@ async def test_build_and_deploy(
     logger.info("Integrating peer-certificates and client-certificates relations")
     await ops_test.model.integrate(f"{APP_NAME}:peer-certificates", TLS_NAME)
     await ops_test.model.integrate(f"{APP_NAME}:client-certificates", TLS_NAME)
-    await ops_test.model.integrate(REQUIRER_NAME, REQUIRER_TLS_NAME)
+    await ops_test.model.integrate(REQUIRER_NAME, TLS_NAME)
```
skourta, line 125: We shouldn't remove the existing test. We can add a second test because I believe the old version is a valid business case that should be tested.

skourta (commented):
Thank you @reneradoi. Nothing major just a couple of questions from my side.

lib/charms/data_platform_libs/v1/data_interfaces.py
reneradoi: It's already disabled (running but ignoring the result) because we use the `azure_storage` lib as a copy (not yet published).

tests/integration/client_relations/test_client_relations.py
reneradoi: Test has been re-added again.

delgod (approved):

skourta (approved):
Thank you @reneradoi

PR #195, opened 2026-01-07
[DPE-9136] test: rewrites backup integration tests with jubilant

Soundarya03:
This PR addresses [DPE-9136](https://warthogs.atlassian.net/browse/DPE-9136) and consists of changes to the backups and restore integration tests; it rewrites them to use jubilant.

## Solution:
- rewrites tests in `test_restore_verification_failed.py` and `test_s3_backup_restore.py` files to use Jubilant
- modifies a few helpers as required 

- [ ] Manually
- [ ] Unit tests
- [x] Integration tests

[DPE-9136]: https://warthogs.atlassian.net/browse/DPE-9136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/integration/helpers_deployment.py
+++ b/tests/integration/helpers_deployment.py
@@ -13,6 +13,7 @@ from uuid import uuid4
 import jubilant
 from data_platform_helpers.advanced_statuses.models import StatusObject
 from dateutil.parser import parse
+from jubilant.statustypes import StatusInfo
 from ops import StatusBase
 from pytest_operator.plugin import OpsTest
 from tenacity import RetryError, Retrying, stop_after_delay, wait_fixed
@@ -115,6 +116,9 @@ def _progress_line(units: List[Unit]) -> str:
         )
 
     return log
+
+
+ExpectedStatus = StatusObject | str
 
 
 async def get_unit_hostname(ops_test: OpsTest, unit_id: int, app: str) -> str:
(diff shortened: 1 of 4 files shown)


reneradoi (approved):
Thank you @Soundarya03! No concerns on the changes of this PR from my side.

A rather general remark, but can also be fixed in another PR: I've noticed that we are losing the output of the `juju status` and `juju debug-log` steps in the CI because we are now using a temporary model that is only yielding while the test is running. I'd recommend to change that into something like this:

```
@pytest.fixture(scope="module")
def juju(arch: str):
    # `testing` is the default model created by concierge
    juju = jubilant.Juju(model="testing")
    juju.wait_timeout = 1000
    juju.cli("set-model-constraints", f"arch={arch}")
    yield juju
```

This should resolve that issue. For the microk8s model that is added in some of the tests, we can still use a temporary model.

Soundarya03:
> Thank you @Soundarya03! No concerns on the changes of this PR from my side.
> 
> A rather general remark, but can also be fixed in another PR: I've noticed that we are losing the output of the `juju status` and `juju debug-log` steps in the CI because we are now using a temporary model that is only yielding while the test is running. I'd recommend to change that into something like this:
> 
> ```
> @pytest.fixture(scope="module")
> def juju(arch: str):
>     # `testing` is the default model created by concierge
>     juju = jubilant.Juju(model="testing")
>     juju.wait_timeout = 1000
>     juju.cli("set-model-constraints", f"arch={arch}")
>     yield juju
> ```
> 
> This should resolve that issue. For the microk8s model that is added in some of the tests, we can still use a temporary model.

Ah yes, makes sense. Thanks a lot, Rene! 
I've added this as a comment on the PR for HA tests, which is currently a WIP. Will make sure to include this change there.

tests/integration/helpers_deployment.py
```diff
@@ -506,16 +532,16 @@ async def wait_until(  # noqa: C901
 
 def does_status_match(
     model_status: jubilant.Status,
-    expected_unit_statuses: dict[str, List[StatusObject]] | None = None,
-    expected_app_statuses: dict[str, List[StatusObject]] | None = None,
+    expected_unit_statuses: dict[str, List[ExpectedStatus]] | None = None,
```
skourta, line 535: Can be in a separate PR but can we replace the `List` from typing with normal the `list`

skourta (approved):
Thank you for the great work @Soundarya03

tests/integration/helpers_deployment.py
Soundarya03: Oh, okay. Is it because of [this](https://stackoverflow.com/a/62033243)?
I'll add a comment to the HA PR

PR #200, opened 2026-01-15
[DPE-9097] Stable release upgrades test

reneradoi:
This PR adds an integration test to ensure we can upgrade the first `stable` version of `charmed-etcd` to the current version in a production-like setup with:
- TLS enabled
- persistent storage
- a client application (data-interfaces `v0`)

## Solution:
Ensure the upgrade runs without issues and doesn't break compatibility.

- [X] Manually
- [ ] Unit tests
- [X] Integration tests

### DIFF
--- a/tests/integration/ha/upgrades/test_stable_release_upgrade.py
+++ b/tests/integration/ha/upgrades/test_stable_release_upgrade.py
@@ -25,8 +25,10 @@ from tests.integration.helpers_deployment import wait_until
 logger = logging.getLogger(__name__)
 
 CHARM_CHANNEL = "3.6/stable"
+# these revisions are the ones from the first stable release, they are supposed to be kept
 CHARM_REVISIONS_TO_DEPLOY = {"x86_64": 119, "aarch64": 120}
 REQUIRER_NAME = "requirer-charm"
+REQUIRER_TLS_NAME = "requirer-tls-provider"
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


tests/integration/ha/upgrades/test_stable_release_upgrade.py
```diff
@@ -0,0 +1,126 @@
...
+    """Path to the requirer charm file to use for testing."""
+    return f"./tests/integration/client_relations/requirer-charm/requirer-charm_ubuntu@24.04-{platform}.charm"
+
+
+@pytest.mark.abort_on_fail
+async def test_deploy_stable_revision(ops_test: OpsTest, requirer_charm: str) -> None:
+    """Deploy the charm with the first stable release, in a production-like setup."""
+    logger.info("Create storage pool for persistent storage")
+    await ops_test.model.create_storage_pool("etcd-pool", "lxd")
+    storage = {
```
skourta, line 36: Just out of curiousity, why the storage?

skourta (approved):
Thank you @reneradoi

tests/integration/ha/upgrades/test_stable_release_upgrade.py
reneradoi: To have a "production-like setup" and make sure we don't break storage constraints that would lead to incompatibility on upgrading. And we don't have a test with upgrades and storage yet.

Mehdi-Bendriss (approved):

PR #208, opened 2026-02-26
[DPE-9407] feat: support for full keyspace access by client

Soundarya03:
Addresses [DPE-9407](https://warthogs.atlassian.net/browse/DPE-9407)

Client app integrating over `etcd_client` interface should be able to request access to entire keyspace, rather than "/" onwards, as is currently the case. This is necessitated by the benchmark tool, for instance, which doesn't support specifying a prefix; it simply writes all keys directly like so: "\0x1", etc. The existing version of charmed-etcd is unable to support this.

## Solution:
This PR introduces changes to:
- validations involving client request. Exception raised only when resource requested is None, rather than an empty string.
- `grant_permission` and `_run_etcdctl` methods in `EtcdClient`, to make use of the `--from-key ''` option to grant access to entire keyspace.
- unit and integration tests, to validate the changes above. 

- [x] Manually
- [x] Unit tests
- [x] Integration tests

This change has been tested manually, by deploying the modified charm alongside a client app requesting access to:
- a prefix : should work as expected
- full keyspace (resource="") : should work as expected
- No resource specified in the client request: should error out

[DPE-9407]: https://warthogs.atlassian.net/browse/DPE-9407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/events/external_clients.py
```diff
@@ -84,7 +84,7 @@ def _on_bulk_resources_requested(  # noqa: C901
 
         responses = []
         for request in event.requests:
-            if not request.mtls_cert or not request.resource:
+            if not request.mtls_cert:
                 logger.error("mTLS certificate or resource not provided")
```
reneradoi: nit: Please adjust the logging message.

src/events/external_clients.py
```diff
@@ -136,7 +136,7 @@ def _on_bulk_resources_requested(  # noqa: C901
 
     def _on_mtls_cert_updated(self, event: MtlsCertUpdatedEvent[RequirerCommonModel]) -> None:  # noqa: C901
         """Handle the ca chain updated event."""
-        if not event.request.mtls_cert or not event.request.resource:
+        if not event.request.mtls_cert:
             logger.error("CA chain, keys prefix, or common name not provided")
```
reneradoi: nit: adjust the logging message.

reneradoi (dismissed):
Great work, concise and straight to the point. Thank you very much Soundarya!

src/events/external_clients.py
Soundarya03: Done!

reneradoi (dismissed):

tests/integration/client_relations/requirer-charm/src/etcd_requires.py
```diff
@@ -216,10 +220,15 @@ def client_requests(self) -> list:
...
         return [
             RequirerCommonModel(
-                resource=f"/{common_name}/",
-                mtls_cert=self.charm.get_certificate_of_common_name(common_name) or "",
-            )
-            for common_name in self.charm.common_names
+                resource="",  # access to full keyspace
+                mtls_cert=self.charm.get_certificate_of_common_name(self.charm.common_names[0])
+                or "",
+            ),
```
skourta, line 226: Why did you break the list into the two instances?

src/events/external_clients.py
```diff
@@ -136,8 +136,8 @@ def _on_bulk_resources_requested(  # noqa: C901
 
     def _on_mtls_cert_updated(self, event: MtlsCertUpdatedEvent[RequirerCommonModel]) -> None:  # noqa: C901
         """Handle the ca chain updated event."""
-        if not event.request.mtls_cert or not event.request.resource:
-            logger.error("CA chain, keys prefix, or common name not provided")
+        if not event.request.mtls_cert:
```
skourta, line 139: Can you please explain why you removed the check on resource?

skourta (commented):
Just minor questions. Thank you @Soundarya03

tests/integration/client_relations/requirer-charm/src/etcd_requires.py
Soundarya03: I wanted to test two scenarios: one where we set the resource to a non-empty string as we were doing earlier, and another where we test the newly supported empty resource (that denotes access to full keyspace, with no particular prefix)

src/events/external_clients.py
Soundarya03: The resource can now be an empty, which is what denotes a request to access the full keyspace. Hence, I removed this check.

src/common/client.py
```diff
@@ -571,6 +578,9 @@ def grant_permission(self, rolename: str, key_prefix: str) -> None:
             rolename (str): The role name to grant permission to.
             key_prefix (str): The key prefix to grant permission for.
         """
+        if "" == key_prefix:
```
Mehdi-Bendriss: (Applies to all occurrences)

```suggestion
        if key_prefix == "":
```

Having the constant first is a Java pattern to avoid NPEs (`"".equals(variable)`), but feels unnatural in Python

Other question, is the `key_prefix` ever trimmed? (`.strip()`)

src/common/client.py
```diff
@@ -579,6 +589,7 @@ def grant_permission(self, rolename: str, key_prefix: str) -> None:
             auth_password=self.password,
             user=rolename,
             prefix=key_prefix,
+            from_key=key_prefix if ("" == key_prefix) else None,
```
Mehdi-Bendriss: `from_key` as a string feels a bit awkward since it only concerns "empty string" or None, how about as a bool?

src/events/external_clients.py
Mehdi-Bendriss: (applies to all occurrences) We should then have an explicit check on the empty string, `not event.request.resource` includes both empty strings and `None` ?

Mehdi-Bendriss (commented):
I haven't reviewed the tests, i'll leave those to René and Smail, but i have a couple of questions in the code

src/common/client.py
Soundarya03: Makes sense, done! 

No, it is not stripped anywhere. But I'm not sure if I should? Etcd prefixes are byte literal right, so technically it could be a valid prefix..

src/common/client.py
Soundarya03: I kept it this way to align better with the `etcdctl role grant-permission` options. In future, if we need to invoke etcdctl with a non-empty `from_key` option, this would be helpful. 
Besides, since the `_run_etcdctl` method matches its argument names more or less against the etcdctl commands' options, I felt it would be right to follow the pattern.

src/events/external_clients.py
Soundarya03: Yes, this check has been removed entirely because it was decided that `None` is a valid case too, and is taken to mean "access to full keyspace" as well. In fact, the distinction between None and empty string cannot be made in our etcd code, because data platform libs sets `resource` to a default value of '' (empty string). 
For more info, [this thread](https://chat.canonical.com/canonical/pl/a5e7ykwebfdoiyi55um9qesabw) on MM can be viewed.

reneradoi: If I understand it correctly, `data-interfaces` sets the default for `resource` to an empty string, see [here](https://github.com/canonical/data-platform-libs/blob/main/lib/charms/data_platform_libs/v1/data_interfaces.py#L970). So it doesn't make sense to check for `None` on the provider side.

src/common/client.py
reneradoi: I have double-checked: It is possible to grant a role the permissions to a prefix of an empty string. Example: 

```
ubuntu@dev-vm:~$ charmed-etcd.etcdctl put mykey 42 --user="myuser" --password="secret"
{"level":"warn","ts":"2026-04-07T10:45:06.604998+0200","logger":"etcd-client","caller":"v3@v3.6.7/retry_interceptor.go:65","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc0003ba5a0/127.0.0.1:2379","method":"/etcdserverpb.KV/Put","attempt":0,"error":"rpc error: code = PermissionDenied desc = etcdserver: permission denied"}
Error: etcdserver: permission denied

ubuntu@dev-vm:~$ charmed-etcd.etcdctl role grant-permission --prefix myrole readwrite "" --user="root" --password="root"
Role myrole updated

ubuntu@dev-vm:~$ charmed-etcd.etcdctl put mykey 42 --user="myuser" --password="secret"
OK

ubuntu@dev-vm:~$ charmed-etcd.etcdctl get mykey --user="myuser" --password="secret"
mykey
42
```

So we don't need the extra parameter `from-key` for the `grant-permission` command.

Soundarya03: Makes sense, thanks for pointing out!

Mehdi-Bendriss (approved):
Approved - I haven't reviewed the tests @reneradoi / @skourta

reneradoi (approved):
Re-approval, changes are fine from my side, tests have been adjusted to account for the feature.

PR #217, opened 2026-03-31
[DPE-9709] Migrate to charmlibs pathops

Soundarya03:
Addresses [DPE-9709](https://warthogs.atlassian.net/browse/DPE-9709).

This PR introduces changes, primarily to `src/workload.py`, to use charmlibs.pathops in place of pathlibs, in order to generalize the code and make it reusable for kubernetes workloads in future.

- [ ] Manually
- [x] Unit tests
- [x] Integration tests

[DPE-9709]: https://warthogs.atlassian.net/browse/DPE-9709?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/workload.py
```diff
@@ -219,7 +234,7 @@ def data_storage_size(self) -> int:
         Returns:
             int: The size of the data storage in Bytes.
         """
-        return shutil.disk_usage(SNAP_DATA_PATH).total
+        return shutil.disk_usage(str(self.root_path / SNAP_DATA_PATH)).total
```
Soundarya03: This method is the only one that will still need an effort to be rewritten for k8s workloads. Is this acceptable for now? Is there a different approach I can take?

src/managers/tls.py
```diff
@@ -124,9 +131,9 @@ def load_trusted_ca(self, tls_type: TLSType) -> set[str]:
             tls_type (TLSType): The TLS type. Defaults to TLSType.PEER.
         """
         if tls_type == TLSType.CLIENT:
-            ca_certs_path = Path(self.workload.paths.tls.client_ca)
+            ca_certs_path = self.root / self.workload.paths.tls.client_ca
```
Mehdi-Bendriss: there is no need to externalise `self.root` here, better include it directly in all exposed paths of the workload

src/charm.py
```diff
@@ -48,20 +49,21 @@ class EtcdOperatorCharm(ops.CharmBase):
     def __init__(self, *args):
         super().__init__(*args)
         # Show logger name (module name) in logs
+        self.root: pathops.PathProtocol = pathops.LocalPath("/")
```
Mehdi-Bendriss: similar comment

src/workload.py
```diff
@@ -132,25 +132,40 @@ def restart(self) -> None:
 
     @override
     def copy_file(self, src_file: str, dst_file: str) -> None:
-        copyfile(src_file, dst_file)
+        src = self.root_path / src_file
+        dst = self.root_path / dst_file
+        dst.parent.mkdir(exist_ok=True, parents=True)
```
Mehdi-Bendriss: this seems to be diverging from the previous behavior, we should not create parent dirs when the goal is to simply copy. A missing parent dir may be an issue worth raising on.

Mehdi-Bendriss (commented):
Thanks Soundarya, I have a couple of comments

src/managers/tls.py
Soundarya03: Alright!

src/managers/tls.py
```diff
@@ -44,10 +44,16 @@ class TLSManager(ManagerStatusProtocol):
...
+    def __init__(
+        self,
+        state: ClusterState,
+        workload: WorkloadBase,
+        substrate: SUBSTRATES,
+    ):
         self.state = state
         self.workload = workload
         self.substrate = substrate
+        self.root = pathops.LocalPath("/")
```
reneradoi: I'm afraid this will not work, because it is not substrate agnostic. I would recommend to make `root` part of the workload base class instead, and then override it in the per-substrate class.

An example for this can be found in the Valkey code base:
- Base workload class: https://github.com/canonical/valkey-operator/blob/9/edge/src/core/base_workload.py#L47
- VM workload class: https://github.com/canonical/valkey-operator/blob/9/edge/src/workload_vm.py#L53

All other paths should be build on top of that, then. An example of how the TLS paths could be structured can be found here: https://github.com/canonical/valkey-operator/blob/9/edge/src/core/base_workload.py#L17

src/workload.py
```diff
@@ -102,25 +102,25 @@ def is_reachable(self, host: str, port: int) -> bool:
 
     @override
     def write_file(self, content: str, file: str) -> None:
-        path = Path(file)
+        path = self.root / file
```
reneradoi: I think it makes sense to implement these methods in the `WorkloadBase` class instead, as the purpose of the `pathops` lib is to abstract away differences between substrates. So while we previously had just an abstract method in the base class, now the implementation should be moved there and the substrate-specific classes should inherit the implementation.

This applies to all file- or directory-based operations.

reneradoi (changes requested):
Thank you for working on this, Soundarya! I have a few general comments, assuming the goal of the PR is to make the code substrate-agnostic with regards to file / directory handling? If so, I think the implementation of these methods should be moved to the base class (see detailed comment). And I would also like to see the TLS paths refactored.

src/workload.py
Soundarya03: Ah, yes. This makes sense. Thanks for the guidance, René!

src/managers/tls.py
Soundarya03: Got it! Done.

reneradoi:
I want to point you to another thing that needs to be adapted, and it is hidden deep in the `EtcdClient` class implementation. It is about adding TLS certificates to the operator's `etcdctl` commands if they exist. Please take a look here: https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/src/common/client.py#L427

Soundarya03:
> I want to point you to another thing that needs to be adapted, and it is hidden deep in the `EtcdClient` class implementation. It is about adding TLS certificates to the operator's `etcdctl` commands if they exist. Please take a look here: https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/src/common/client.py#L427

Thank you! Done. I've checked for file ops involving backups and database files too, and made the necessary changes. Can you please take a look?

src/core/workload.py
```diff
@@ -22,56 +26,82 @@
 class TLSPaths:
     """Paths for TLS."""
 
-    root_dir: str = TLS_ROOT_DIR
+    def __init__(self, root_dir: pathops.LocalPath | pathops.ContainerPath):
+        self.tls_root = root_dir / TLS_ROOT_DIR
 
     @property
-    def peer_ca(self) -> str:
+    def peer_ca(self):
```
reneradoi: Please add return types for all added properties and functions.

src/managers/config.py
Soundarya03, line 57: qq @reneradoi: considering the comment below, I believe I should keep this as is?

src/charm.py
```diff
@@ -48,6 +49,7 @@ class EtcdOperatorCharm(ops.CharmBase):
...
```
reneradoi: Two comments here:
1) This can be removed as it is not used.
2) For understanding: `LocalPath` is not substrate agnostic, meaning it would fail in K8s environments. The `LocalPath` or `ContainerPath` classes should only be instantiated in substrate-specific classes, otherwise we should ensure to only deal with the `PathProtocol`, which is for code that should work on both substrates.

src/workload.py
```diff
@@ -219,7 +171,7 @@ def data_storage_size(self) -> int:
         Returns:
             int: The size of the data storage in Bytes.
         """
-        return shutil.disk_usage(SNAP_DATA_PATH).total
+        return shutil.disk_usage(str(self.root_dir / SNAP_DATA_PATH)).total
```
reneradoi: Instead of casting to `str`, we can use the `.as_posix()` method of a `pathops` object.

src/workload.py
```diff
@@ -64,7 +62,8 @@ def install(self, revision: str | None = None, retry_and_raise: bool = True) ->
             True if successfully installed, False if errors occur and `retry_and_raise` is False.
         """
         if not revision:
-            versions = self.load_toml_file(f"{WORKING_DIR}/../{VERSIONS_FILE}")
+            versions_path = (WORKING_DIR / ".." / VERSIONS_FILE).resolve()
+            versions = tomllib.loads(versions_path.read_text())
```
reneradoi: I think we should revert this to use the method `load_toml_file` again as before, otherwise we would have to add error handling for possible errors on `read_text()` to this method here.

src/managers/config.py
reneradoi: Yes correct, this runs on the unit where the charm code is installed, not in the workload container. Therefor it should stay as is.

reneradoi (commented):
Just a few minor comments, then we should be good.

src/charm.py
Soundarya03: Oops, I seemed to have cleaned up other usages, but missed this primary one, sorry! Done.

src/workload.py
Soundarya03: Got it, will cast this to LocalPath (since it handles substrate specific installation for VM), and use the workload method.

reneradoi (dismissed):
No other concerns from my side, thank you Soundarya!

src/core/workload.py
```diff
@@ -22,56 +26,82 @@
 class TLSPaths:
     """Paths for TLS."""
 
-    root_dir: str = TLS_ROOT_DIR
+    def __init__(self, root_dir: pathops.PathProtocol):
+        self.tls_root = root_dir / TLS_ROOT_DIR
 
     @property
-    def peer_ca(self) -> str:
+    def peer_ca(self) -> pathops.PathProtocol:
```
Mehdi-Bendriss: across the board in this PR (all files), please only import the specific class, without the reference to the full module

Mehdi-Bendriss (commented):
Thanks Soundarya, that's a great addition, I only have a minor comment on the import

src/core/workload.py
Soundarya03: Got it!

reneradoi (approved):
Re-approval

Mehdi-Bendriss (approved):

PR #224, opened 2026-04-23
[DPE-9515] chore: update etcd to 3.6.10

Soundarya03:
Update snap revisions. Also update workload versions for tests.

- [ ] Manually
- [x] Unit tests
- [x] Integration tests


tests/integration/ha/upgrades/literals.py
```diff
@@ -6,6 +6,6 @@
 
 NUM_UNITS = 3
 CHARM_CHANNEL = "3.6/edge"
-CHARM_REVISIONS_TO_DEPLOY = {"x86_64": 135, "aarch64": 136}
-WORKLOAD_VERSION = {"previous": "3.6.5", "target": "3.6.7"}
+CHARM_REVISIONS_TO_DEPLOY = {"x86_64": 159, "aarch64": 160}
+WORKLOAD_VERSION = {"previous": "3.6.7", "target": "3.6.10"}
```
Mehdi-Bendriss: i would rather we keep the source `3.6.5` and change the target instead

refresh_versions.toml
```diff
@@ -8,6 +8,6 @@ name = "charmed-etcd"
 
 [snap.revisions]
 # amd64
-x86_64 = "27"
+x86_64 = "32"
```
Mehdi-Bendriss, line 11: are these the new snap revisions?

Mehdi-Bendriss (commented):
Thanks Soundarya, left a couple of comments

refresh_versions.toml
Soundarya03: Yes

tests/integration/ha/upgrades/literals.py
Soundarya03: Done!

reneradoi: @Mehdi-Bendriss Why that? This test should ensure that users can upgrade from the previous etcd version to newer one. We have a separate test to ensure upgrading from the first stable release is possible, to.

In addition: If the test should update from 3.6.5, the charm revisions to be deployed in the line above also need to be adjusted again.

Soundarya03:
Updated the `get_raft_leader` helper and `get_cluster_members` invocations to always include username and password, as the [latest version of etcd](https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.6.md) includes changes that tighten auth requirements.

reneradoi (approved):
No concerns from my side, thank you Soundarya! Please ensure integration tests are passing before merging.

Mehdi-Bendriss (approved):

PR #225, opened 2026-04-30
Move to GH-hosted runners

Soundarya03:
The PR migrates integration tests to runner on Github instead of self-hosted runners.

Changes:
- use `22.04` runners because of the kernel issue in `6.17`
- add a Python `3.12` installation to the integration tests, because this is the default Python version required by the project
- update a couple of workflows to newer Github actions due to `Node.js 20 actions are deprecated`
- adjust the timing of integration tests for upgrades because it takes up to 75 seconds (as per observation) to run `juju refresh` on arm runners
- adjust the timing of the backup integration test to ensure etcd has processed the relation with s3 integrator before running a backup action


reneradoi (dismissed):

reneradoi (approved):

skourta (approved):
Thank you @reneradoi

PR #228, opened 2026-05-13
chore: fix the workload version in `refresh_versions.toml`

reneradoi:
addresses #231 

fix the workload version in `refresh_versions.toml` after recent updates


Soundarya03 (approved):
Ah, this was a miss when I updated the snap, I see.. Thanks, René!

skourta (approved):

